PR #388, opened 2024-02-20
Fix nflatlev in test for fused_vel_adv_8_to_13. Also remove stencil

muellch:
Should fix the bug in fused_velocity_advection_stencil_8_to_13 in the unit test.


model/atmosphere/dycore/tests/dycore_stencil_tests/test_fused_velocity_advection_stencil_8_to_13.py
```diff
@@ -117,7 +117,7 @@ def input_data(self, grid):
         k = indices_field(KDim, grid, is_halfdim=True, dtype=int32)
 
         nlev = grid.num_levels
-        nflatlev = 13
+        nflatlev = 4
 
         istep = 1
```
nfarabullini, line 122: can you edit import and type annotations like we do in other stencil tests? e.g.
```
def input_data(self, grid: base.BaseGrid) -> dict[str, gtx.Field | state_utils.ScalarType]:
```
```
from icon4py.model.common import dimension as dims
```

PR #462, opened 2024-05-15
Tracer advection utest

huppd:
adds more advection stencil tests


msimberg:
@huppd I'll close this as a thing or two has changed in icon4py in two years... please feel free to reopen if you think it's worth reviving.

PR #514, opened 2024-07-30
DaCe Orchestration for the Diffusion Granule

kotsaloscv:
This PR concerns only the DaCe backend.

Currently, the Diffusion granule calls the various stencils one after the other (in `_do_diffusion_step `). This means that for every stencil, there is a separate SDFG, and consequently one stencil is not aware of the others (this behavior limits the analyzability of the full granule). This PR introduces a decorator that fuses all these SDFGs under one compilation unit, allowing DaCe for further analysis and optimizations. Placing a GT4Py program inside a DaCe program region, and extracting the underlying SDFG, is possible due to [this GT4Py PR](https://github.com/GridTools/gt4py/pull/1527).

The halo exchanges are also taken care from the DaCe orchestrator. A follow-up PR will introduce automated halo exchanges. Currently, the halo exchange class implements the SDFGConvertible interface like GT4Py Program.

The orchestrator is activated either through an env var `ICON4PY_DACE_ORCHESTRATION=AnyValue` or through this pytest option `--dace-orchestration=AnyValue`.

The orchestrator suppports ahead of time compilation, however given that DaCe does not support nested Structures -like self in the Diffusion class-, some of the arguments need to be provided at compile time, through `dace.compiletime`. This annotation means that the correspoding argument will be considered from the closure of the function.

The orchestrator provides full caching support and takes into consideration when a dace.compiletime argument changes, with a subsequent re-compilation of the fused SDFG.


model/common/src/icon4py/model/common/orchestration/dtypes.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+if dace:
+    # Define DaCe Symbols
+    CellDim_sym = dace.symbol("CellDim_sym")
+    EdgeDim_sym = dace.symbol("EdgeDim_sym")
+    VertexDim_sym = dace.symbol("VertexDim_sym")
+    KDim_sym = dace.symbol("KDim_sym")
+
+    DiffusionDiagnosticState_symbols = {
```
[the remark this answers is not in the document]
kotsaloscv: I have modified the code as you asked. Now everything is dynamic according to the data class members.

model/common/src/icon4py/model/common/orchestration/__init__.py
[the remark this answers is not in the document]
kotsaloscv: the orchestration module tries to import dace and if it does not find it, then it complaints accordingly.
Do you think that I should add it as optional dependency as well?

kotsaloscv:
cscs-ci run dace

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+                for attr_name, attr_value in self.__dict__.items():
+                    if isinstance(attr_value, decomposition.ExchangeRuntime) and not has_exchange:
+                        has_exchange = True
+                        exchange_obj = getattr(self, attr_name)
+                    if isinstance(attr_value, icon_grid.IconGrid) and not has_grid:
+                        has_grid = True
+                        grid = getattr(self, attr_name)
+
+                if not has_exchange or not has_grid:
+                    raise ValueError("No exchnage/grid object found.")
```
edopao, line 99: Typo: exchange

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+
+                compile_time_args_kwargs = {}
+                all_args_kwargs = list(args) + list(kwargs.values())
+                for i, (k, v) in enumerate(fuse_func.__annotations__.items()):
+                    if v is dace.compiletime:
+                        compile_time_args_kwargs[k] = all_args_kwargs[i]
+                unique_id = hash(tuple([id(e) for e in compile_time_args_kwargs.values()]))
+
+                default_build_folder = os.path.join(
+                    ".dacecache", f"MPI_rank_{exchange_obj.my_rank()}"
```
edopao, line 109: Is there a way to get ".dacecahe" through the dace API? like `dace.config.Config().get("default_build_folder")`

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+            )
+        )
+
+        def _concretize_symbols_for_dace_structure(dace_cls, orig_cls):
+            concretized_symbols = {}
+            for k_v in flattened_xargs_type_value:
+                if k_v[0] is not dace_cls:
+                    continue
+                for member in orig_cls.__dataclass_fields__.keys():
+                    for stride in [0, 1]:
```
edopao, line 500: Can we generalize `for stride in [0, 1]` by looping on the number of dimensions?

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
```diff
@@ -0,0 +1,325 @@
...
+    def __sdfg__(self, *args, **kwargs) -> dace.SDFG:
+        sdfg = dace.SDFG("DummyNestedSDFG")
+        state = sdfg.add_state()
+
+        sdfg.add_scalar(name="__return", dtype=dace.int32)
+
+        tasklet = dace.sdfg.nodes.Tasklet(
+            "DummyNestedSDFG",
+            inputs=None,
+            outputs=None,
```
edopao, line 303: My experience tells that `outputs` should be the set of output connectors: `{"__out"}`

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
```diff
@@ -0,0 +1,325 @@
...
+        state.add_node(tasklet)
+
+        state.add_edge(
+            tasklet,
+            "__out",
+            state.add_write("__return"),
+            None,
+            dace.Memlet(data="__return", subset="0"),
+        )
+        tasklet.out_connectors = {"__out": dace.int32}
```
edopao, line 317: So you can avoid this line:
`tasklet.out_connectors = {"__out": dace.int32}`

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -204,6 +223,174 @@ def exchange_and_wait(self, dim: Dimension, *fields: tuple):
...
+        def dace__sdfg__(self, *args, **kwargs):
+            sdfg = dace.SDFG("_halo_exchange_wait_")
+            state = sdfg.add_state()
+
+            # The communication handle used in the halo_exchange tasklet is a global variable
+            # ghex::communication_handle<communication_handle_type> h_{id(self.exchange_object)};
+            # Therefore, this tasklet calls the wait() method on the communication handle -disregards any input-
+            tasklet = dace.sdfg.nodes.Tasklet(
+                "_halo_exchange_wait_",
+                inputs=None,
```
edopao, line 323: Same comment applies to multiple places: `inputs` and `outputs` should be set of strings representing the names of input and output connectors, respectively:
```suggestion
                inputs={f"IN_{buffer_name}"},
                outputs={"__out"},
```

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -204,6 +223,174 @@ def exchange_and_wait(self, dim: Dimension, *fields: tuple):
...
+                language=dace.dtypes.Language.CPP,
+                side_effects=False,
+            )
+            state.add_node(tasklet)
+
+            # Dummy input to maintain same interface with non-DaCe branch
+            buffer_name = HaloExchangeWait.buffer_name
+            sdfg.add_scalar(name=buffer_name, dtype=dace.int32)
+            buffer = state.add_read(buffer_name)
+            tasklet.in_connectors["IN_" + buffer_name] = dace.int32.dtype
```
edopao, line 335: So you can avoid these lines:
`tasklet.in_connectors["IN_" + buffer_name] = dace.int32.dtype`

edopao (commented):
LGTM

edopao:
Please also merge again latest main to get some fixes in stencil tests.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -691,42 +713,47 @@ def _do_diffusion_step(
...
+        cached.apply_diffusion_to_w_and_compute_horizontal_gradients_for_turbulence.with_connectivities(
+            self.compile_time_connectivities
+        )(
             area=self.cell_params.area,
             geofac_n2s=self.interpolation_state.geofac_n2s,
             geofac_grg_x=self.interpolation_state.geofac_grg_x,
             geofac_grg_y=self.interpolation_state.geofac_grg_y,
             w_old=self.w_tmp,
             w=prognostic_state.w,
-            type_shear=gtx.int32(self.config.shear_type.value),
```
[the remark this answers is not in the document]
kotsaloscv: that's right. it messes up with the DaCe parser.

model/atmosphere/diffusion/tests/diffusion_tests/mpi_tests/test_parallel_diffusion.py
```diff
@@ -117,3 +118,167 @@ def test_parallel_diffusion(
...
+)
+def test_parallel_diffusion_multiple_steps(
+    experiment,
+    step_date_init,
+    linit,
+    ndyn_substeps,
+    processor_props,  # fixture
+    decomposition_info,
+    icon_grid,
+    diffusion_savepoint_init,
```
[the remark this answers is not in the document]
kotsaloscv: In this case, the save points are not needed. I just wanted the `test_parallel_diffusion_multiple_steps` to have the same signature as the `test_parallel_diffusion` (maybe in the future, we could use them).

model/common/src/icon4py/model/common/orchestration/decorator.py
kotsaloscv: nice catch ;)

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
kotsaloscv: you are right. This is just an initialisation of the tasklet. If you check a bit below, I set it properly.
```
out_connectors["__out"] = ...
tasklet.out_connectors = out_connectors
```

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
kotsaloscv: you are right that I could write it like this, but I prefer to do a simple initialisation there and afterwards I set the inputs/outputs. I just maintain the same approach as the `GHexMultiNodeExchange`.

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

ci/dace.yml
```diff
@@ -18,7 +18,23 @@ test_model_stencils_x86_64:
...
+    - COMPONENT: [atmosphere/diffusion/tests/diffusion_tests]
+      BACKEND: [dace_cpu]
+      DACE_ORCHESTRATION: ['--dace-orchestration=True', '']
+test_model_datatests_x86_64:
+  extends: [.test_model_datatests, .test_template_x86_64]
+test_model_datatests_aarch64:
+  extends: [.test_model_datatests, .test_template_aarch64]
+
+benchmark_model_stencils:
+  extends: .test_template
```
edopao, line 37: Remove this line (not in baseline):
` extends: .test_template`

kotsaloscv:
cscs-ci run dace

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
```diff
@@ -0,0 +1,325 @@
...
+from dace.memlet import Memlet
+from dace.properties import CodeBlock
+from dace.sdfg.state import SDFGState
+from gt4py.next.ffront.fbuiltins import Dimension
+
+import icon4py.model.common as common
+from icon4py.model.common import dimension as dims
+
+
+ghex_ptr_names = (
```
egparedes, line 45: Follow standard PEP8 conventions (https://peps.python.org/pep-0008/#constants)
```suggestion
GHEX_PTR_NAMES: Final = (
```

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
```diff
@@ -0,0 +1,325 @@
...
+    # Tasklet code generation part
+    fields_desc_glob_vars = "\n"
+    fields_desc = """
+    bool levels_first;
+    std::size_t outer_strides;
+    std::size_t levels;
+
+    """
+    descr_unique_names = []
+    for i, arg in enumerate(copy.deepcopy(list(global_buffers.values()))):
```
egparedes, line 156: Why is a deepcopy needed here? Is the loop body modifying the nested contents of the sequence?

model/common/src/icon4py/model/common/orchestration/halo_exchange.py
```diff
@@ -0,0 +1,325 @@
...
+                ghex::communication_handle<{communication_handle_type}> h_{unique_id};
+                """
+    else:
+        code = fields_desc_glob_vars
+
+    tasklet.code_global = CodeBlock(code=code, language=dace.dtypes.Language.CPP)
+
+
+@dace.library.environment
+class DaceGHEX:
```
egparedes, line 257: I not sure, but this whole class looks to me like it should be filled with information provided by the GHEX bindings instead of hardcoded here. Maybe a TODO here if it makes sense, to move this functionality to GHEX in the future

egparedes (changes requested):
It's a large and non-trivial PR adding complicated functionality, kudos for the hard work. It looks good but I have some comments and suggestion to improve it.

model/common/src/icon4py/model/common/config.py
```diff
@@ -29,13 +44,28 @@ class GT4PyBackend(Enum):
...
+            return backend
+        else:
+            raise ValueError(
+                f"Invalid ICON4Py backend: {backend}. \n"
+                f"Available backends: {', '.join([f'{k}' for k in GT4PyBackend.__members__.keys()])}"
+            )
+
+    @cached_property
+    def icon4py_dace_orchestration(self):
+        return os.environ.get("ICON4PY_DACE_ORCHESTRATION", None)
```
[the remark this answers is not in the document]
kotsaloscv: any value activates DaCe Orchestration, that's why it is not critical to do value checking.
We are going to restructure the config file and for this reason I followed this simplified approach.

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -156,6 +161,122 @@ def my_rank(self):
...
+            )
+
+        def dace__sdfg_signature__(self) -> tuple[Sequence[str], Sequence[str]]:
+            raise NotImplementedError(
+                "__sdfg_signature__ is only supported when the 'dace' module is available."
+            )
+
+    __sdfg__ = dace__sdfg__
+    __sdfg_closure__ = dace__sdfg_closure__
+    __sdfg_signature__ = dace__sdfg_signature__
```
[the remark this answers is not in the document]
kotsaloscv: https://github.com/GridTools/gt4py/pull/1527#discussion_r1687591550

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+                        has_grid = True
+                        grid = getattr(self, attr_name)
+
+                if not has_exchange or not has_grid:
+                    raise ValueError("No exchange/grid object found.")
+
+                compile_time_args_kwargs = {}
+                all_args_kwargs = list(args) + list(kwargs.values())
+                for i, (k, v) in enumerate(fuse_func.__annotations__.items()):
+                    if v is dace.compiletime:
```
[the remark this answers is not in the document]
kotsaloscv: you are right. the function called above (`to_dace_annotations()`) makes sure that all args are annotated ;)

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+                        )
+                        (
+                            compiled_sdfgs[unique_id]["compiled_sdfg"],
+                            _,
+                        ) = dace_program.load_precompiled_sdfg(
+                            dace_program_location, **compile_time_args_kwargs
+                        )
+                except:  # noqa: E722
+                    raise ValueError(
+                        "Corrupted cache. Remove `.dacecache` folder and re-run the program."
```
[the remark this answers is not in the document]
kotsaloscv: by construction yes. Although I further modified the code to make it more robust.

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+        fuse_func: Callable,
+        fuse_func_orig_annotations: dict[str, Any],
+        args: Any,
+        kwargs: Any,
+    ) -> dict[str, Any]:
+        flattened_xargs_type_value = list(
+            zip(
+                list(fuse_func.__annotations__.values()),
+                list(args) + list(kwargs.values()),
+                strict=False,
```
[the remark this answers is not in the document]
kotsaloscv: this is imposed by the auto-formatter.

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -599,12 +612,13 @@ def _do_diffusion_step(
             smag_offset:
 
         """
-        num_levels = self.grid.num_levels
         # dtime dependent: enh_smag_factor,
-        scale_k(self.enh_smag_fac, dtime, self.diff_multfac_smag, offset_provider={})
+        scale_k.with_connectivities(self.compile_time_connectivities)(
```
halungge, line 616: That is a gt4py question really: 
By using `with_connectivities()` is it then still necessary to add `offset_provider` kwarg, or can that be removed? To me it looks a bit redundant, but it is probably needed to pass the actual runtime connectivity tables?

Also what is the advantage or disadvantage of using it? Using the `with_backend` call it used to be significantly slower than setting a `backend` `@program` decorator, one of the reasons of the backend mess that we have in icon4py...
Maybe that changed as well. 

Maybe we can have a discussion on this with somebody from the gt4py team?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -351,6 +358,7 @@ def __init__(
         self, exchange: decomposition.ExchangeRuntime = decomposition.SingleNodeExchange()
     ):
         self._exchange = exchange
+        self.halo_exchange_wait = decomposition.create_halo_exchange_wait(self._exchange)
```
halungge, line 361: Can you make this private? 
```suggestion
        self._halo_exchange_wait = decomposition.create_halo_exchange_wait(exchange)
```
and is the `self._exchange`still needed? Otherwise please delete it?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -62,6 +63,12 @@
 from icon4py.model.common.settings import xp
 from icon4py.model.common.utils import gt4py_field_allocation as field_alloc
 
+from icon4py.model.common.orchestration.decorator import (
+    orchestrate,
+    build_compile_time_connectivities,
```
halungge, line 68: As I wrote elsewhere, I think the `build_compile_time_connectivities` should probably be somewhere in `common/grid/utils.py` or so. It is not anything dace specific.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
halungge: Or put differently, I just realised that I don't understand how those 2 the `ExchangeRuntime` and the `ExchangeAndWait` should be used in the `Diffusion` most of the halo exchange calls are as before via the runtime, you only use the ExchangeAndWait at the end, for the single `exchange.wait`. Shouldn't one replace the other or only use it internally?

halungge (commented):
I still have a few questions, some of them also gt4py related... Maybe we can discuss this during the next days?

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,568 @@
...
+
+@dace_inhibitor
+def wait(comm_handle: Union[int, decomposition.ExchangeResult]):
+    if isinstance(comm_handle, int):
+        pass
+    else:
+        comm_handle.wait()
+
+
+def build_compile_time_connectivities(
```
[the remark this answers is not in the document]
kotsaloscv: Currently, this is used only in DaCe backends. 
See: https://github.com/GridTools/gt4py/blob/f7e970782db20b66dd768fc347dfd9936c87af7d/src/gt4py/next/ffront/decorator.py#L77

This is the reason I have it there.

requirements-dev-opt.txt
```diff
@@ -1,4 +1,4 @@
-git+https://github.com/ghex-org/GHEX.git@61ca5f0e3b6b473f9ca397052b2b0c1d8a4d3c55#subdirectory=bindings/python
+git+https://github.com/ghex-org/GHEX.git@master#subdirectory=bindings/python
```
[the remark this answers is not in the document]
kotsaloscv: By running the tests in the CI, everything works fine. Should I do it locally also?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
halungge: same as above.

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

requirements-dev-opt.txt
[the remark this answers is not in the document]
halungge: CI does not run yet the parallel tests. But we tried it on Friday and that looked good.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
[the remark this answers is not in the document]
halungge: In case that cannot be fixed easily in dace you could add a reminder in comment that it needs to be like this due to dace. Otherwise somebody might cleanup/change it, breaking dace orchestration...

halungge (approved):

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

egparedes (commented):
Final round of reviews. There are only small comments and it should be ready to go.

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -156,6 +161,135 @@ def my_rank(self):
...
+        res = self.exchange(dim, *args)
+        if wait:
+            res.wait()
+        else:
+            return res
+
+    if dace:
+        # Implementation of DaCe SDFGConvertible interface
+        # For more see [dace repo]/dace/frontend/python/common.py#[class SDFGConvertible]
+        def dace__sdfg__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
kotsaloscv: Hi @egparedes 
the `dace__sdfg__` returns an SDFG. The problem is that in the else branch, there is no dace such that I can specify the return type. What should I do in these cases?

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,584 @@
...
+
+                if not has_exchange or not has_grid:
+                    raise ValueError("No exchange/grid object found.")
+
+                compile_time_args_kwargs = {}
+                all_args_kwargs = list(args) + list(kwargs.values())
+                for i, (k, v) in enumerate(fuse_func.__annotations__.items()):
+                    if v is dace.compiletime:
+                        compile_time_args_kwargs[k] = all_args_kwargs[i]
+                unique_id = hash(tuple([id(e) for e in compile_time_args_kwargs.values()]))
```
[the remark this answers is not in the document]
kotsaloscv: As discussed, the new approach deals with the issue.

model/common/src/icon4py/model/common/orchestration/decorator.py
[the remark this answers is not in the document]
kotsaloscv: in my initial version, I was using zip but without specifying the strict kwarg. Therefore, the autformatter was introducing it as `strict=False`. I want through all the zips of the decorator and changed the strict to True because I want an error to be raised if the iterables are not of equal size.

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,673 @@
...
+    uid = hash_object.hexdigest()
+    return uid
+
+
+def order_kwargs_by_annotations(f: Callable, kwargs: dict[str, Any]) -> None:
+    """Order in-place the kwargs based on the annotations of the function."""
+    annotations = f.__annotations__
+    ordered_items = [(key, kwargs[key]) for key in annotations if key in kwargs]
+    kwargs.clear()
+    kwargs.update(ordered_items)
```
egparedes, line 279: Is this still used? Otherwise, it should be removed
```suggestion
```

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,673 @@
...
+            obj_to_traverse = obj.__dict__ if hasattr(obj, "__dict__") else obj
+            for key, value in obj_to_traverse.items():
+                _populate_entry(key, value, parent_key)
+
+    if hasattr(obj, "__dict__") or isinstance(obj, dict):
+        _populate_unique_dict(obj)
+    else:
+        _populate_entry(obj_name, obj)
+
+    return uid_from_hashlib(str(unique_dict))
```
egparedes, line 264: The names of this function and the previous one do not follow a similar naming scheme despite they are closely related, I would try to rename them. Additionally, both functions are lacking the arg description in their docstrings and this one seems more convoluted than needed. I would suggest to refactor it to something with a single internal helper function like:

```suggestion
def generate_orchestration_uid(
    obj: Any, obj_name: str = "", members_to_disregard: tuple[str] = ()
) -> str:
    """Generate a unique id for a runtime object.

    The unique id is generated by creating a dictionary that describes the runtime state of the object.
    For primitive types, the dictionary contains the type and the value.
    For arrays, the dictionary contains the shape and the dtype -not their content-.

    Keep in mind that this function is not supposed to be generic, and should only be used for
    DaCe orchestration purposes.

    Args:
        obj:
        obj_name:
        members_to_disregard:
    """
    primitive_dtypes = (*orchestration_dtypes.ICON4PY_PRIMITIVE_DTYPES, str, uuid.UUID, np.dtype)

    static_data = {}

    def _populate_entry(key: str, value: Any) -> None:
        if key in members_to_disregard:
            return

        if isinstance(value, primitive_dtypes):
            static_data[key] = {"type": "primitive_dtypes", "value": str(value)}

        elif isinstance(value, (np.ndarray, gtx.Field)):
            static_data[key] = {
                "type": "array/field",
                "shape": str(value.shape),
                "dtype": str(value.dtype),
            }

        elif isinstance(value, (list, tuple)):
            if all(isinstance(i, primitive_dtypes) for i in value):
                static_data[key] = {
                    "type": f"array-like[{'empty' if len(value) == 0 else type(value[0])}]",
                    "length": str(len(value)),
                }
            else:
                for child_key, child_value in enumerate(value):
                    _populate_entry(f"{key}.{child_key!s}", child_value)

        elif isinstance(value, decomposition.ExchangeRuntime):
            pass

        elif value is None:
            static_data[key] = {"type": "None", "value": "None"}

        elif isinstance(value, dict) or getattr(obj, "__dict__", None):
            for child_key, child_value in getattr(obj, "__dict__", obj).items():
                _populate_entry(f"{key}.{child_key!s}", child_value)

        else:
            raise ValueError(f"Type {type(value)} is not supported.")

    _populate_entry(obj_name, obj)

    return uid_from_hashlib(str(static_data))
```

egparedes (commented):
I only looked at the `decorator.py` module and tried to provide as much information and suggestions as possible to make your life easier and try to get this done and merged ASAP.

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,673 @@
...
+            if "dace" not in settings.backend.name.lower():
+                raise ValueError(
+                    "DaCe Orchestration works only with DaCe backends. Change the backend to a DaCe supported one."
+                )
+
+            self_name = next(iter(inspect.signature(fuse_func).parameters))
+
+            # If not explicitly set by the user, assume the provided callable is a method
+            # when its first argument is called 'self'
+            func_is_method = method or (self_name == "self")
```
DropD, line 76: There is a simpler and more robust way of checking for "method-ness":
```suggestion
            # If not explicitly set by the user, check the argument is a method
            func_is_method = method or isinstance(fuse_func, types.MethodType)
```

Note: this of course requires `types` (a builtin) to be imported.

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,673 @@
...
+
+            if not func_is_method:
+                raise NotImplementedError(
+                    "The orchestration decorator is only for methods -at least for now-."
+                )
+
+            def wrapper(*args, **kwargs):
+                self = args[0]
```
DropD, line 84: Why not simply
```suggestion
            def wrapper(self, *args, **kwargs):
```

model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -0,0 +1,673 @@
...
+                    )
+                else:
+                    cache["sdfg"] = dace_program.to_sdfg(
+                        **compile_time_args_kwargs, simplify=False, validate=True
+                    )
+                sdfg = cache["sdfg"]
+
+                if sdfg.name != f"{fuse_func.__module__}.{fuse_func.__name__}".replace(".", "_"):
+                    raise ValueError(
+                        "fused_SDFG.name != {fuse_func.__module__}.{fuse_func.__name__}"
```
DropD, line 442: I assume the intention was to use a format-string:
```suggestion
                        f"fused_SDFG.name != {fuse_func.__module__}.{fuse_func.__name__}"
```

model/common/src/icon4py/model/common/orchestration/decorator.py
kotsaloscv: The orchestrator could be used for non member functions as well -not yet implemented, but possibly in the future-. So, this is the reason for not having self in the arg list.

model/common/src/icon4py/model/common/orchestration/decorator.py
kotsaloscv: Good catch ;)

kotsaloscv:
cscs-ci run dace

model/common/src/icon4py/model/common/orchestration/decorator.py
kotsaloscv: I have updated the docstring and the sorting of the static_data dictionary but for now I would prefer to keep the two helper functions, if that is ok with you.

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

kotsaloscv:
cscs-ci run dace

egparedes (approved):
LGTM

PR #528, opened 2024-08-16
disable flaky timeloop test

halungge:
This test still crashes from time to time. I think it is something else than the domain indices as the error message hints to some 
problem with reference counting/ pointer allocation.
I propose to disable it until we have time to investigate it further: 
- temporarily disabling test that sometimes fails with `double free / corruption error`


jcanton:
is this not in main yet and tests are now passing despite the random memory issues?

halungge:
Problem should be fixed by [PR 531](https://github.com/C2SM/icon4py/pull/531)

PR #531, opened 2024-08-23
Refactor grid domain zones

halungge:
## Refactoring of the `IconGrid.get_start_index` and `IconGrid.get_end_index`:

The way to retrieve start and end indices on the grid is now based on `Domain` object which contains the `Dimension` and the Grid `Zone` that the user is interested in. This allows to control the access to the `_start_idx` and `end_idx` arrays which do contain entries that are not safe and must not be accessed. 

Previously we had no control over the access because the `HorizontalMarkerIndex` was essentially an `int` and it was possible (and heavily used to compute on it for example `HorizontalMarkerIndex(CellDim) + 1`)

New way of retrieving start and end indices:
```python
lb_start = grid.start_index(horizontal.domain(CellDim)(horizontal.Zone.LATERAL_BOUNDARY))
```
where 
```python
def horizontal.domain(CellDim):
      pass 

```
is a factory function for `CellDomain` objects such that

```python
domain: CellDomain = horizontal.domain(CellDim)(horizontal.Zone.LATERAL_BOUNDARY)
```
returns a fully defined   `Domain` object to be passed to `IconGrid.start_index` and `IconGrid.end_index`

Mapping tables for translation of the ICON (Fortran) constants used in `get_indices_e, v, c` were added to as doc_string to`horizontal.py`

### Further changes:

- split `horizontal.py` into `horizontal.py` and `geometry.py`
- fix the GHEX version used in `requirements-dev-opt.txt`

PR #539, opened 2024-09-04
Push index resolution to init

halungge:
Spill over from this [PR](https://github.com/C2SM/icon4py/pull/531): pushed all grid start and end index resolution to `init` in `SolveNonHydro` and `Diffusion`.

### DIFF
--- a/model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
+++ b/model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
@@ -407,7 +407,7 @@ class Diffusion:
             nrdmax=self.vertical_grid.end_index_of_damping_layer,
         )
         self._determine_horizontal_domains()
-        
+
         self._initialized = True
 
     @property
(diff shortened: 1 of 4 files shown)


halungge:
cscs-ci run default

halungge:
launch jenkins spack

OngChia (approved):

OngChia (approved):
I now think these grid indices are something similar to global physical constants and there are only finite number of them. Just my thought: Is it better to define them in common and import them from granules and restrict `horizontal_start` and `horizontal_end` to be equal to one of them? If no, why?

jcanton:
> I now think these grid indices are something similar to global physical constants and there are only finite number of them. Just my thought: Is it better to define them in common and import them from granules and restrict `horizontal_start` and `horizontal_end` to be equal to one of them? If no, why?

in `IconGrid` directly then?
then it's self-contained and is already there for future domain nesting

jcanton (approved):

halungge:
They are ob

> I now think these grid indices are something similar to global physical constants and there are only finite number of them. Just my thought: Is it better to define them in common and import them from granules and restrict `horizontal_start` and `horizontal_end` to be equal to one of them? If no, why?

they are obviousely dependent on the grid, so you can only know what they are once you have a grid file loaded. That is way they way to look them up  (in our case the domain boundary that we want to use is separate from the real index.

PR #547, opened 2024-09-13
Add tach dependency enforcer

egparedes:
This PR only adds the configuration for the [tach](https://docs.gauge.sh/getting-started/introduction) tool and its pre-commit hook in non-strict mode. In further PRs stricter checks can be added.


egparedes:
cscs-ci run default

egparedes:
launch jenkins spack

tach.toml
```diff
@@ -0,0 +1,83 @@
+exclude = []
```
halungge, line 1: I guess it cannot read the dependencies from the `pyproject.toml`?

tach.toml
```diff
@@ -0,0 +1,83 @@
...
+[[modules]]
+path = "icon4py.model.driver"
+depends_on = [
+    { path = "icon4py.model.atmosphere.diffusion" },
+    { path = "icon4py.model.atmosphere.dycore" },
+    { path = "icon4py.model.common" },
+]
+
+[[modules]]
+path = "icon4pytools"
```
halungge, line 49: For the tools, I don't know what kind of restriction we should have. Most of them look reasonable to me. In the end, they are all in the same package so at least you never end up installing something that cannot be built because it imports a missing dependency. This would currently be the case if we imported something from `model.atmosphere.diffusion` in `model.atmosophere.dycore`

tach.toml
```diff
@@ -0,0 +1,83 @@
...
+    { path = "icon4pytools.icon4pygen" },
+]
+
+[[modules]]
+path = "icon4pytools.py2fgen"
+depends_on = [
+    { path = "icon4py.model.atmosphere.diffusion" },
+    { path = "icon4py.model.common" },
+    { path = "icon4pytools.common" },
+    { path = "icon4pytools.icon4pygen" },
```
halungge, line 82: This one I am not sure it is by accident or not. Let'  s see what @samkellerhals  says.

tach.toml
```diff
@@ -0,0 +1,83 @@
...
+    { path = "icon4py.model.common" },
+]
+
+[[modules]]
+path = "icon4py.model.common"
+depends_on = []
+
+[[modules]]
+path = "icon4py.model.driver"
+depends_on = [
```
halungge, line 42: you could already add here 
```suggestion
depends_on = [
{path = "icon4py.model.atmosphere.advection`},
{path = "icon4py.model.atmosphere.subgrid_scale_physics.microphysics"}
```
Those are not there yet, but they are allowed.

tach.toml
```diff
@@ -0,0 +1,83 @@
...
+    { path = "icon4pytools" },
+    { path = "icon4pytools.common" },
+    { path = "icon4pytools.icon4pygen" },
+]
+
+[[modules]]
+path = "icon4pytools.py2fgen"
+depends_on = [
+    { path = "icon4py.model.atmosphere.diffusion" },
```
halungge, line 79: you might already add 
```suggestion
    { path = "icon4py.model.atmosphere.diffusion" },
    { path = "icon4py.model.atmosphere.dycore" },
```

for the dycore granule.

halungge (approved):

tach.toml
egparedes: Just FYI, the current dependencies are the ones that the tool automatically detects in the codebase.

tach.toml
egparedes: Sure, let's wait for @samkellerhals take on this.

tach.toml
egparedes: I don't understand the comment. `exclude` is a list of project directories to exclude from checks.

tach.toml
samkellerhals: I just had a look and took the opportunity to do a bit of refactoring around module dependencies in `icon4pytools`. I also updated the `tach.toml` file, hopefully it is correct!

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

egparedes:
I just ran the code formatters and tach to update the dependencies and the dependency graph now looks like this:
deps

@samkellerhals I would prefer to split the changes in 2 PRs: one with the code refactoring which should be merged first and another one with just the tach configuration. Can you extract your changes from this branch and open a new PR?

samkellerhals:
> I just ran the code formatters and tach to update the dependencies and the dependency graph now looks like this: deps
> 
> @samkellerhals I would prefer to split the changes in 2 PRs: one with the code refactoring which should be merged first and another one with just the tach configuration. Can you extract your changes from this branch and open a new PR?

I've moved the changes to https://github.com/C2SM/icon4py/pull/549

egparedes:
cscs-ci run default

egparedes:
launch jenkins spack

PR #508, opened 2024-07-16
[py2f]: Add dycore wrapper and datatest

huppd:
### Description

Adds dycore wrapper and datatest for single and multitimestep runs. The Fortran integration test is skipped for now as we need to pass the correct connectivities from Fortran in order to initialise the ICON grid on the Python side. In the future this may be achieved using serialised connectivities and using Serialbox to read them in on the Fortran side.

PR #548, opened 2024-09-16
[py2f]: Add diffusion grid initialisation and datatest

samkellerhals:
### Description

This update adds a new datatest for the diffusion wrapper, specifically testing it for a regional experiment using serialized data for the `Diffusion.init` and `Diffusion.run` functions. The wrapper has also been modified to initialize its own grid using connectivities provided by ICON.

Also added `test_diffusion_wrapper_granule_inputs` to verify that the diffusion wrapper correctly constructs and passes the necessary parameters to `Diffusion.init` and `Diffusion.run` when using granule inputs. This ensures proper integration between the wrapper and the diffusion module. The idea is to have tests like this for every granule and will also be implemented for the dycore granule in https://github.com/C2SM/icon4py/pull/508

The Fortran integration test is currently skipped, as we need to pass the correct connectivities from Fortran to initialize the ICON grid on the Python side. Future improvements may include using serialized connectivities and Serialbox to load them on the Fortran side.

**Note:** The introduction of `TEST_TEMP_DIR` as a file writing location for the code generation tests is to avoid `No space left on device` errors, which occur as files written to `/tmp` are stored in RAM (which fills up after a while...), and therefore we now store them in the current working directory instead.


samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

tools/tests/py2fgen/test_diffusion_wrapper.py
```diff
@@ -6,146 +6,174 @@
...
-    inv_dual_edge_length = as_field((dims.EdgeDim,), inv_dual_edge_length)
-    inv_vert_vert_length = as_field((dims.EdgeDim,), inv_vert_vert_length)
-    edge_areas = as_field((dims.EdgeDim,), edge_areas)
-    f_e = as_field((dims.EdgeDim,), f_e)
-    cell_areas = as_field((dims.CellDim,), cell_areas)
+    nudge_max_coeff = (
+        0.075 * DEFAULT_PHYSICS_DYNAMICS_TIMESTEP_RATIO
+    )  # this is done in ICON, so we replicate it here
+    itype_sher = TurbulenceShearForcingType.VERTICAL_HORIZONTAL_OF_HORIZONTAL_VERTICAL_WIND
+    nflat_gradp = 27
```
OngChia, line 60: You can set `nflat_gradp` from the `grid_savepoint`

OngChia (changes requested):

tools/tests/py2fgen/test_diffusion_wrapper.py
samkellerhals: Thanks did that now

tools/tox.ini
```diff
@@ -17,6 +17,7 @@ passenv =
     CUDAARCHS
     CUDACXX
     NVFORTRAN_COMPILER
+    TEST_TEMP_DIR
 deps =
```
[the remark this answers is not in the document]
samkellerhals: I think for now there's no need for an additional pytest flag, we can keep using the `--datatest` flag. If we start adding lots of additional tests using serialised data and want to for example run them apart from the rest (because they took long), then I would think about adding a new flag.

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

tools/src/icon4pytools/py2fgen/wrappers/diffusion.py
```diff
@@ -65,8 +67,64 @@ def profile_disable():
     stats.dump_stats(f"{__name__}.profile")
 
 
+def grid_init(
+    cell_starts: Field[[dims.CellIndexDim], int32],
```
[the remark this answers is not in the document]
samkellerhals: It's a good point. I moved it outside now, however this required building a bit more "infrastructure" around it, which makes sense since there are more things in this shared global state such as the profiler for example, or other granules.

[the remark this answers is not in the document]
samkellerhals: Also allows for cleaner separation of concerns.

tools/src/icon4pytools/py2fgen/wrappers/common.py
```diff
@@ -0,0 +1,135 @@
...
+import numpy as np
+from icon4py.model.common import dimension as dims
+from icon4py.model.common.grid import base, horizontal, icon
+from icon4py.model.common.settings import xp
+
+
+log = logging.getLogger(__name__)
+
+
+def offset_fortran_indices_return_numpy(inp) -> np.ndarray:
```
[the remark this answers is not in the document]
samkellerhals: No that was an oversight on my side.

model/common/src/icon4py/model/common/dimension.py
```diff
@@ -60,3 +60,6 @@
 
 
 VerticalDim: TypeAlias = Union[KDim, KHalfDim]
+CellIndexDim = Dimension("CellIndex")
```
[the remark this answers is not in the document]
samkellerhals: I ended up putting them into wrappers.common as I need them for the interface

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

model/common/src/icon4py/model/common/test_utils/diffusion_utils.py
OngChia, line 1: Just a question. I thought this was initially not done because this will make `common` depend on `diffusion`. Is it because the entire pytest infrastructure will be moved out from common in the future so that it is now okay to put these functions here in `common`?

tools/tests/py2fgen/test_diffusion_wrapper.py
```diff
@@ -6,146 +6,528 @@
...
+    construct_config,
+    construct_diagnostics,
+    construct_interpolation_state,
+    construct_metric_state,
+    vertical_grid,
+)
 
+import icon4pytools.py2fgen.wrappers.diffusion
+from icon4pytools.py2fgen.wrappers import wrapper_dimension
 from icon4pytools.py2fgen.wrappers.diffusion import diffusion_init, diffusion_run
```
OngChia, line 34: Is it possible to just use one import for ` icon4pytools.py2fgen.wrappers.diffusion`?

tools/tests/py2fgen/test_diffusion_wrapper.py
```diff
@@ -6,146 +6,528 @@
...
+        global_level=global_level,
+        num_vertices=num_vertices,
+        num_cells=num_cells,
+        num_edges=num_edges,
+        vertical_size=vertical_size,
+        limited_area=limited_area,
+    )
+
+    # --- Mock and Test diffusion_granule.init ---
+    with mock.patch("icon4py.model.atmosphere.diffusion.diffusion.Diffusion.init") as mock_init:
```
OngChia, line 224: I learn this mock class from your PR. Thanks.
The mock test here tests if the input arguments of the diffusion wrapper and the diffusion granule are the same. But it does not test whether the input argument remain the same when you pass arguments from the fortran side. Maybe you can extend the current test to do it in the future.

tools/tests/py2fgen/conftest.py
```diff
@@ -9,8 +9,67 @@
...
 @pytest.fixture
 def samples_path():
     return Path(__file__).parent / "fortran_samples"
+
+
+def compare_values_shallow(value1, value2, obj_name="value"):
+    # Compare if both are objects with attributes (__dict__)
+    if hasattr(value1, "__dict__") and hasattr(value2, "__dict__"):
+        if value1.__class__ != value2.__class__:
+            return False, f"Class mismatch for {obj_name}: {value1.__class__} != {value2.__class__}"
```
OngChia, line 24: This does not test whether the actual values of class members are the same or not. Do you think it is necessary to test the values? Or maybe I am thinking too much.

tools/tests/py2fgen/conftest.py
```diff
@@ -9,8 +9,67 @@
...
+    if isinstance(value1, xp.ndarray) and isinstance(value2, xp.ndarray):
+        try:
+            xp.testing.assert_equal(value1, value2)  # Compare arrays for equality
+            return True, None
+        except AssertionError:
+            return False, f"Array mismatch for {obj_name}"
+
+    if isinstance(value1, dict) and isinstance(value2, dict):
+        if value1.keys() != value2.keys():
+            return False, f"Dict keys mismatch for {obj_name}: {value1.keys()} != {value2.keys()}"
```
OngChia, line 36: Similarly here, is it better to also check the values?

model/common/src/icon4py/model/common/test_utils/diffusion_utils.py
samkellerhals: Hmm good question so I put it here because I need that functionality in my wrapper datatests. I suppose as long as there are no circular dependency issues I am personally fine with leaving it here as it is for now, and coming up with a potentially better solution to store pytest infrastructure in a separate module later.

tools/tests/py2fgen/conftest.py
samkellerhals: I had some issues at first doing this recursively, so I went with a simpler implementation here. However an expanded check definitely makes this test much more useful so I implemented this in the next PR to merge after this one (dycore granule PR) https://github.com/C2SM/icon4py/pull/508

See this file: https://github.com/C2SM/icon4py/blob/2bd09991871e55739f811d95e691ff5972b27a6d/tools/tests/py2fgen/conftest.py

tools/tests/py2fgen/test_diffusion_wrapper.py
samkellerhals: So what this actually tests is whether when we pass serialised data from Fortran to the wrapper, that the wrapper uses these fields and scalars to construct exactly the same objects that we would expect as inputs to `Diffusion.init` and `Diffusion.run`.

The `mock.patch` basically gives you access to the arguments of the `Diffusion.init` method inside `captured_args, captured_kwargs = mock_init.call_args` allowing us to compare the constructed objects (by the wrapper) to the expected objects (constructed using the fixtures used in the datatests).

tools/tests/py2fgen/conftest.py
samkellerhals: See my previous comment

tools/tests/py2fgen/conftest.py
samkellerhals: You are right, I've extended the checks now.

tools/tests/py2fgen/test_diffusion_wrapper.py
samkellerhals: Done

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

OngChia (approved):
Looks good to me.

tools/tests/py2fgen/test_diffusion_wrapper.py
OngChia: Okay. Understood now. Thanks!

halungge (changes requested):
Mostly the dependency issue: 
diffusion depends on common, but common must not depend on diffusion. You need to work your way around that.

All the different namespace packages are installable independently, the dependency you introduce breaks that.

samkellerhals:
cscs-ci run default

halungge (approved):

PR #550, opened 2024-09-17
Fix base image for docker build

edopao:
There is no reason for using 2 different base images between DaintXC and Todi image build. The issue that prevented gpu-tests on Todi to run with a plain ubuntu image (runtime error `cudaErrorInsufficientDriver`) was that Alps uses the Container Engine as container runtime, which requires some `NVIDIA_` environment variables to be set in the docker image. This was done by default in the cuda images provided by Nvidia docker hub. For plain ubuntu images, we have to explicitly set them, until the vCluster service manager configures the Container Engine on Todi to enable them transparently on all images.


edopao:
cscs-ci run default

edopao:
launch jenkins spack

samkellerhals (approved):

PR #551, opened 2024-09-18
sample model component for diagnostics (temperature)

halungge:
Implements the computation of temperature diagnostics as a model component: implementing `model.common.components.Component`


msimberg:
Closing as it's outdated, but the Component protocol needs to be made useful.

PR #559, opened 2024-09-25
Documentation dycore

jcanton:
Early stage dycore documentation.

config.py for automagically fetching docstrings in the middle of methods (not just the head one) by defining `autofull`. See the example in `dycore_numerics.rst`

```
.. autofull:: icon4py.model.atmosphere.dycore.nh_solve.solve_nonhydro.SolveNonhydro.run_predictor_step
   :no-index:
```

Instructions for compilation are in the main README.md

The "titles" of each section (names of the stencils) were supposed to be hyperlinks to the auto-documented stencils. However, these are gt4py-decorated programs (or field operators) that need a patch in gt4py in order to be picked up by sphinx's autodoc (see https://github.com/GridTools/gt4py/pull/1603  ).

The dycore docs (with the first two operations documented) should look like this:

Screenshot 2024-09-25 at 12 54 29


jcanton:
cscs-ci run default

jcanton:
launch jenkins spack

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity/velocity_advection.py
```diff
@@ -509,6 +710,12 @@ def run_corrector_step(
 
         self._update_levmask_from_cfl_clipping()
 
+        """
+        z_w_con_c_full (0:nlev-1):
+            Compute the vertical wind with contravariant correction at full levels (cell center) by
+            taking average of from it values at neighboring half levels.
+            z_w_con_c_full[k] = 0.5 (z_w_con_c[k] + z_w_con_c[k+1])
```
nfarabullini, line 717:
```suggestion
            z_w_con_c_full[k] = 0.5 (z_w_con_c[k] + z_w_con_c[k+1])
```
I would remove lines like this (e.g. above as well), usually docstrings are used to explain in words what fields are and two cents about what the stencils does

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity/velocity_advection.py
```diff
@@ -557,6 +784,17 @@ def run_corrector_step(
...
+        """
+        ddt_vn_apc_pc[ntnd] (0:nlev-1):
+            Compute the advection of normal wind at full levels (edge center).
+            ddt_vn_apc_pc[ntnd] = dKEH/dn + vt * (vorticity + coriolis) + wdvn/dz,
+            where vn is normal wind, vt is tangential wind, KEH is horizontal kinetic energy.
+            z_kin_hor_e is KEH at full levels (edge center).
+            zeta is vorticity at full levels (vertex). Its value at edge center is simply taken as average of vorticity at neighboring vertices.
+            w, which is vertical wind with contravariant correction, at edge center is obtained by linearly interpolating z_w_con_c_full at neighboring cells.
+            dvn/dz, the vertical derivative of normal wind, is computed by first order discretization from vn_ie.
+            TODO (Chia Rui): understand the coefficient coeff_gradekin
```
nfarabullini, line 796:
```suggestion
            ddt_vn_apc_pc[ntnd] = dKEH/dn + vt * (vorticity + coriolis) + wdvn/dz,
            where vn is normal wind, vt is tangential wind, KEH is horizontal kinetic energy.
            z_kin_hor_e = KEH at full levels (edge center).
            zeta = vorticity at full levels (vertex). Its value at edge center is an average of vorticity at neighboring vertices.
            w = vertical wind with contravariant correction. Its value at edge center is obtained by linearly interpolating z_w_con_c_full at neighboring cells.
            dvn/dz = vertical derivative of normal wind, computed by first order discretization from vn_ie.
            TODO (Chia Rui): understand the coefficient coeff_gradekin
```

nfarabullini: This is more or less out convention in other docstrings. Can you adapt documentation in other places in this PR plase?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity/velocity_advection.py
```diff
@@ -479,6 +657,14 @@ def run_corrector_step(
             offset_provider=self.grid.offset_providers,
         )
 
+        """
+        z_w_con_c (0:nlev):
```
nfarabullini, line 661:
```suggestion
        z_w_con_c:
```
You don't need to specify bounds unless there is a very particular corner case

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
```diff
@@ -140,6 +143,9 @@ def predictor_stencils_2_3(
     vertical_start: gtx.int32,
     vertical_end: gtx.int32,
 ):
+    """
+    Temporary docstring to see if this program is picked up
+    """
```
nfarabullini, line 148: is this still needed?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/sphinx.py
```diff
@@ -0,0 +1,99 @@
...
+
+    @classmethod
+    def can_document_member(cls, member, membername, isattr, parent):
+        return isinstance(member, (Program, FieldOperator, ScanOperator)) and member.__wrapped__
+
+    def format_args(self):
+        wrapped = getattr(self.object, "__wrapped__", None)
+        if wrapped is not None:
+            sig = inspect.signature(wrapped)
+            if "self" in sig.parameters or "cls" in sig.parameters:
```
nfarabullini, line 54: can you explain what you are trying to check here?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro.py
```diff
@@ -1842,6 +2461,27 @@ def run_corrector_step(
...
+            pi^{n+1} = pi_reference + z_exner_expl - dt * vwind_impl_wgt ( rd * pi^{n} ) / ( cvd * rho^{n} * theta_v^{n} ) ( rho^{n}_{k-1/2} w^{n+1}_{k-1/2} theta_v^{n}_{k-1/2} - rho^{n}_{k+1/2} w^{n+1}_{k+1} theta_v^{n}_{k+1/2} ) / dz_{k} )
+            eta_impl = 0.5 + vwind_offctr = vwind_impl_wgt
+            eta_expl = 1.0 - eta_impl = vwind_expl_wgt
+            Note that rho^{n} and theta_v^{n} are used for the implicit flux term.
+            rho^{n}_{k-1/2} theta_v^{n}_{k-1/2} is represented by z_alpha.
+            dt * vwind_impl_wgt (rd * pi^{n}) / (cvd * rho^{n} * theta_v^{n} ) / dz_{k} is represented by z_beta.
+        theta_v (0:nlev-1):
+            Update virtual potential temperature at full levels (cell center) from the equation of state (see eqs. 3.22 and 3.23 in ICON tutorial 2023).
+            rho^{n+1} theta_v^{n+1} = rho^{n} theta_v^{n} + ( cvd * rho^{n} * theta_v^{n} ) / ( rd * pi^{n} ) ( pi^{n+1} - pi^{n} )
+        """
```
nfarabullini, line 2484: This is a really long docstring and takes away focus from what computations are actually performed in this file. I would move it to a pure md file and place a reference here.

Same goes for other long docstrings above

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
jcanton: no because the patch doesn't work and this doesn't get picked up, I'll remove it

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/sphinx.py
jcanton: @havogt?

jcanton:
Would it be better if I removed all the non properly formatted docstrings and left only the two of the screenshot at the top of the page?

egparedes (commented):
Thanks for opening the PR and start a serious discussion about how to add scientific content to the documentation.

However, I think it's hard to contribute to the discussion just looking at the PR. It would really help to have a short meeting (ideally in person, but zoom also works) to discuss this, so you can present the idea, the rationale and show how the final documentation looks like. 

Many people is at OAT on Tuesdays, but it should be also possible to meet at other locations other days, just make sure that everyone who is interested on this project is aware of the meeting and can join.

What do you think? @jcanton @nfarabullini @OngChia @halungge

samkellerhals:
> Thanks for opening the PR and start a serious discussion about how to add scientific content to the documentation.
> 
> However, I think it's hard to contribute to the discussion just looking at the PR. It would really help to have a short meeting (ideally in person, but zoom also works) to discuss this, so you can present the idea, the rationale and show how the final documentation looks like.
> 
> Many people is at OAT on Tuesdays, but it should be also possible to meet at other locations other days, just make sure that everyone who is interested on this project is aware of the meeting and can join.
> 
> What do you think? @jcanton @nfarabullini @OngChia @halungge @havogt ...

I also agree with @egparedes, I think its great that you're thinking about documentation with Sphinx, but maybe we should discuss specifics offline.

jcanton:
cscs-ci run default

jcanton:
launch jenkins spack

jcanton:
launch jenkins spack

model/atmosphere/dycore/docs/icon4py_sphinx.py
```diff
@@ -0,0 +1,364 @@
...
+            elif in_block:
+                if stripped_line.startswith('#'):
+                    current_block.append(line)
+                else:
+                    in_block = False
+                    comment_blocks.append('\n'.join(current_block))
+                    current_block = []
+        if current_block:
+            comment_blocks.append('\n'.join(current_block))
+        return comment_blocks
```
egparedes, line 82: I think the code can be simplified by using the last element of the `comment_blocks` list as current block and avoid storing the non useful comment marks of the scidoc which need to be removed later. I haven't tested the code at all, but I think something along these lines should work:
```suggestion
        comment_blocks = []
        in_block = False
        for line in source.splitlines():
            stripped_line = line.strip()
            is_comment = stripped_line.startswith('#')
            if stripped_line.startswith(f'# {keyword}:'):
                comment_blocks.append([])
            elif in_block and is_comment:
                comment_blocks[-1].append(f"{stripped_line}\n")
            in_block = is_comment

        return comment_blocks
```

By the way, the suggested code will not support the case where a new scidoc is started in the middle of another one. I think that case should be considered as an error since two different scidocs cannot document the same stencil call.
```
# scidoc:
# Outputs:
#  - z_exner_ic :
# scidoc:
# Outputs:
#  - z_exner_ic :
foo_stencil(...)
```

model/atmosphere/dycore/docs/icon4py_sphinx.py
```diff
@@ -0,0 +1,364 @@
...
+
+    def format_docstring_block(self, docstr_lines, keyword):
+        # Clean up and format
+        if docstr_lines[0].strip().startswith(f'# {keyword}:'):
+            docstr_lines.pop(0) # remove the {keyword} prefix
+        while docstr_lines[0].strip() == '#':
+            # strip empty lines from the beginning
+            docstr_lines.pop(0)
```
egparedes, line 90: This could be skipped if the strings are already cleaned when adding them.

model/atmosphere/dycore/docs/icon4py_sphinx.py
```diff
@@ -0,0 +1,364 @@
...
+        # strip leading and trailing whitespace of every line (maintain indentation)
+        # as well as comment character and space (indent+2)
+        indent = len(docstr_lines[0]) - len(docstr_lines[0].lstrip(' '))
+        docstr_lines = [line[min(indent+2,len(line)):].rstrip() for line in docstr_lines]
+        # strip empty lines from the end
+        while docstr_lines[-1] == '':
+            docstr_lines.pop(-1)
```
egparedes, line 97: Use `textwrap.dedent` from the standard library (https://devdocs.io/python~3.13/library/textwrap#textwrap.dedent)

egparedes (changes requested):
This is a partial review because I could only check part of the autodoc extension. I will submit the second part tomorrow.

model/atmosphere/dycore/docs/icon4py_sphinx.py
```diff
@@ -0,0 +1,364 @@
...
+
+            # Identify LaTeX multiline blocks and align equations to the left
+            elif '$$' in line:
+                if not latex_multiline_block and '\\\\' in docstr_lines[iline+1]:
+                    latex_multiline_block=True
+                    continue
+                else:
+                    latex_multiline_block=False
+            if latex_multiline_block:
+                docstr_lines[iline] = insert_before_first_non_space(line, '&')
```
egparedes, line 148: If I properly understood the logic of this conditions I think it can be simplified to something like the following code, which adds the `&` char at the beginning of the lines inside a multiline math block

```suggestion
            elif '$$' in line:
                latex_math_block = not latex_math_block
            if latex_math_block and line.rstrip().endswith(r'\\'):  # multiline math block
                start_idx = len(line) - len(line.lstrip()) 
                docstr_lines[line_num] = f"{line[:start_idx]}&{line[start_idx:]}"
```

Also the `insert_before_first_non_space()` looks unnecessary to me.

egparedes (commented):
Some more comments but still not a full review. I think it'd be better to discuss and stabilize the code with the suggested changes (when possible) from these first two reviews, before continuing, since the code might change quite a bit.

model/atmosphere/dycore/docs/icon4py_sphinx.py
jcanton: I agree with most everything (and adapted), the only thing I kept is the entire block unformatted because I use it in its entirety when looking for the next method call:
`call_string = self.get_next_method_call(source, source.find(docstring) + len(docstring))`
I could re-write such that source.splitlines() is the "main processed object" and only `get_next_method_call` uses `''.join(source_splitlines)`, then base the start of the search on a line index instead of a char index as is now.
Let me know if you think it makes more sense.
(this option would also allow to clean up the lines directly when adding them to the block and avoid postprocessing them in a second step)

model/atmosphere/dycore/docs/icon4py_sphinx.py
jcanton: see comment above

model/atmosphere/dycore/docs/icon4py_sphinx.py
jcanton: it only works with one string, not a list of strings, which would mean joining them again just for this?

model/atmosphere/dycore/docs/icon4py_sphinx.py
```diff
@@ -0,0 +1,364 @@
...
+                    return s[:i] + char_to_insert + s[i:]
+            return s
+
+        for iline, line in enumerate(docstr_lines):
+
+            # Make collapsible Inputs section
+            if line.startswith('Inputs:'):
+                past_inputs = True
+                docstr_lines[iline] = '.. collapse:: Inputs'
+                docstr_lines.insert(iline+1, '')
```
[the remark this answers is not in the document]
jcanton: I had initially accepted this as is, but then realised it doesn't work:
running the debugger the line_num counter is not updated with +2 but only +1, which results in the empty newline being overwritten in the next loop iteration.
Is this because you're defining a "new" docstr_lines instead of inserting in the existing one and this makes a difference in the for loop?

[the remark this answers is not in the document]
jcanton: (this is one of the situations I mentioned where I feel I'm missing the correct CS dictionary or knowledge)

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,453 @@
...
+                                    prefix = '.'.join(var_longname.split('.')[:-1])
+                                    suffix = var_longname.split('.')[-1]
+                                    if prefix:
+                                        vname = '*' + prefix + '*. **' + suffix + '**'
+                                    else:
+                                        vname = '**' + suffix + '**'
+                                else:
+                                    # short name version
+                                    vname = variable
+                                split_line[j] = f"{vname} {self.var_type_formatting}{var_type}{self.var_type_formatting}"
```
egparedes, line 196: This can be simplified using `list.index()`:
```suggestion
                        if self.print_variable_longnames:
                            # long name version
                            var_longname = method_info['var_longnames_map'][variable]
                            prefix = '.'.join(var_longname.split('.')[:-1])
                            suffix = var_longname.split('.')[-1]
                            vname = (f"*{prefix}'*.' if prefix else "") + f"**{suffix}**"
                        else:
                            # short name version
                            vname = variable
                        j = split_line.index(variable)
                        split_line[j] = f"{vname} {self.var_type_formatting}{var_type}{self.var_type_formatting}"
```

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,453 @@
...
+            - 'method_name': The name of the method.
+            - 'module_local_name': The local name of the module where the method is defined.
+            - 'module_full_name': The full name of the module where the method is defined.
+            - 'annotations': The annotations of the method.
+            - 'var_names_map': A dictionary mapping variable names as argument to the called method and variable short name (or argument value) in the present scope.
+            - 'var_longnames_map': A dictionary mapping variable short and long names in the present scope.
+            - 'var_types': A dictionary mapping variable short names and their types.
+        """
+        method_info = {}
+        local_method_name = call_string[0].split('(')[0] # get the method name from the call string (remove open bracket)
```
egparedes, line 273: Optional suggestion:
```suggestion
        local_method_name = call_string[0][:call_string[0].index("(")] # get the method name from the call string (remove open bracket)
```

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,453 @@
...
+                                                    # the method is called from a module import, e.g.
+                                                    # import solve_nonhydro_program as nhsolve_prog
+                                                    # self._stencil = nhsolve_prog.stencil.with_backend(...)
+                                                    module_name = call_chain[0]
+                                                    original_method_name = call_chain[1]
+                                                    if module_name in self.module.__dict__.keys() and isinstance(getattr(self.module, module_name), types.ModuleType):
+                                                        module_obj = getattr(self.module, module_name)
+                                                        method_obj = getattr(module_obj, original_method_name)
+                                                        module_full_name = module_obj.__name__
+                                                        break
```
egparedes, line 356: Move all the statements related to the computation of `module_full_name` to a helper function.

egparedes, line 356: Move the part dealing with the AST to another function

egparedes (commented):
Another round of comments .

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,453 @@
...
+            The formatted documentation lines.
+        """
+        if docblock_lines[0].strip().startswith(f'# {keyword}:'):
+            docblock_lines.pop(0)  # remove the {keyword} prefix
+        while docblock_lines and docblock_lines[0].strip() == '#':
+            # Remove leading empty lines
+            docblock_lines.pop(0)
+        while docblock_lines and docblock_lines[-1].strip() == '#':
+            # Remove trailing empty lines
+            docblock_lines.pop(-1)
```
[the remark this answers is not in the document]
jcanton: white empty lines in the middle of the block are necessary (ReST) otherwise lists are not processed correctly.
`WARNING: Definition list ends without a blank line; unexpected unindent.`
I'll add the `assert` though

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
jcanton: much more elegant

jcanton:
cscs-ci run default

jcanton:
launch jenkins spack

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,532 @@
...
+                - The name of the parent module or class.
+        """
+        source = inspect.getsource(class_obj)
+        tree = ast.parse(source)
+        for node in ast.walk(tree):
+            if isinstance(node, ast.FunctionDef) and node.name == '__init__':
+                for stmt in node.body:
+                    if isinstance(stmt, ast.Assign):
+                        for target in stmt.targets:
+                            if isinstance(target, ast.Attribute) and target.attr == local_shortname:
```
egparedes, line 397: This code is quite hard to read. I would gather first all the relevant nodes from the AST and then process them in a different step. For example:

```python
# Capture relevant `ast.Assign` nodes
init_func = next(
    node for node in ast.walk(tree) if isinstance(node, ast.FunctionDef) and node.name == "__init__"
)
assigns = (
    node
    for node in init_func.body
    if isinstance(node, ast.Assign)
    and any(t for t in node.targets if isinstance(t, ast.Attribute) and t.attr == local_shortname)
)

# Process them according to the type of the assigned value
for node in assigns:
    match node.value:
        case ast.Attribute():
            ...
        case ast.Name():
            ...
        case ast.Call():
            ...

```

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
```diff
@@ -0,0 +1,532 @@
...
+
+        Returns:
+            The processed documentation string lines with applied formatting rules.
+        """
+        
+        latex_math = False
+        latex_math_multiline = False
+        past_inputs = False
+
+        for line_num, line in enumerate(docblock_lines):
```
egparedes, line 158: I know I've already reviewed this part but, as a final comment, I really think the lines shouldn't be modified in place and instead a new lines of processed lines should be generated.

egparedes (changes requested):
Hopefully final round of review.

model/atmosphere/dycore/docs/_ext/icon4py_sphinx.py
jcanton: sorry, forgot about this

jcanton:
cscs-ci run default

jcanton:
launch jenkins spack

jcanton:
cscs-ci run default

jcanton:
launch jenkins spack

jcanton:
launch jenkins spack

egparedes (approved):
LGTM. Thanks for all the hard work to get this done!

jcanton:
Thank you for all the time and the teaching ;-)

PR #568, opened 2024-10-15
saturation backend edits

nfarabullini:
introduced backend arguments in `SaturationAdjustment` class and following edits in stencil calls and field definitions

### DIFF
--- a/model/atmosphere/subgrid_scale_physics/microphysics/tests/test_saturation_adjusment.py
+++ b/model/atmosphere/subgrid_scale_physics/microphysics/tests/test_saturation_adjusment.py
@@ -39,6 +39,7 @@ def test_saturation_adjustment_in_gscp_call(
     metrics_savepoint,
     icon_grid,
     lowest_layer_thickness,
+    backend,
 ):
     entry_microphysics_savepoint = data_provider.from_savepoint_weisman_klemp_graupel_entry(
         date=date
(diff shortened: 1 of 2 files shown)


nfarabullini:
cscs-ci run default

nfarabullini:
launch jenkins spack

nfarabullini:
cscs-ci run default

nfarabullini:
launch jenkins spack

model/atmosphere/subgrid_scale_physics/microphysics/tests/test_saturation_adjusment.py
halungge, line 1: Can you check with @OngChia on what backend these tests run so far? There was an issue with some graupel related things that they need to run on `--backend=embedded`. In any case, it is not a problem with your changes, but we might have to change the CI configuration additionally. (That might only apply the graupel scheme not to saturation adjustement, I don't know.)

halungge (commented):
LGTM

model/atmosphere/subgrid_scale_physics/microphysics/tests/test_saturation_adjusment.py
nfarabullini: Yes, good point. I have assigned him as a PR reviewer as well 👍

OngChia (approved):
Looks good to me too. The pressure diagnostic computation currently cannot run on rountrip. But it is not recommended to use rountrip for real simulations, so I guess it is fine.

model/atmosphere/subgrid_scale_physics/microphysics/tests/test_saturation_adjusment.py
OngChia: The test will fail if you use roundtrip backend. Embedded or other compiled backends are fine. So, I think it is okay because datatest should not be run using roundtrip.

PR #570, opened 2024-10-16
orchestrator fix: keep exchange object optional

DropD:
The orchestrator decorator looks for an **optional** exchange object on the `self` argument of the decorated function. One helper function has code that doesn't check whether it's present before optionally using it.

This small PR puts that code inside an if condition, the same check used throughout the module to check for presence of the exchange object.

### DIFF
--- a/model/common/src/icon4py/model/common/orchestration/decorator.py
+++ b/model/common/src/icon4py/model/common/orchestration/decorator.py
@@ -65,15 +65,13 @@ R = TypeVar("R")
 
 
 @overload
-def orchestrate(func: Callable[P, R], *, method: bool | None = None) -> Callable[P, R]:
-    ...
+def orchestrate(func: Callable[P, R], *, method: bool | None = None) -> Callable[P, R]: ...
 
 
 @overload
 def orchestrate(
     func: None = None, *, method: bool | None = None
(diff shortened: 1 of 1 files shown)


model/common/src/icon4py/model/common/orchestration/decorator.py
```diff
@@ -539,35 +537,37 @@ def dace_specific_kwargs(
...
-            "__comm_ptr": expose_cpp_ptr(exchange_obj._comm)
-            if not isinstance(exchange_obj, decomposition.SingleNodeExchange)
-            else 0,
-            **{
-                f"__pattern_{dim.value}Dim_ptr": expose_cpp_ptr(exchange_obj._patterns[dim])
+            connectivity_identifier(k): v.table
+            for k, v in offset_providers.items()
+            if hasattr(v, "table")
+        }
+        if exchange_obj:
```
DropD, line 546: Would it be better to set all these keys to the default value (0)?

kotsaloscv: I would favour the `if exchange_obj is not None and not isinstance(exchange_obj, decomposition.SingleNodeExchange)` and set the pointers to 0.

msimberg:
The orchestration has been removed, so I'm closing this.

PR #573, opened 2024-10-24
Fix nlevp1 bug in dycore

OngChia:
A temporary fix of the illegal memory access caused by writing or reading full-level variables in a `field_operator` at `nlevp1` in `solve_nonhydro_program.py`. For example,
```
@field_operator
def foo(
    input_var,
    k_field,
    nlev,
):
    (half_level_var, full_level_var) =  where(
        (k_field >= 0) & (k_field < nlev),
        _compute_var(input_var),
        (half_level_var, full_level_var),
    )
    half_level_var = where(k_field == nlev, _init_cell_kdim_field_with_zero_vp(), half_level_var)
    return half_level_var, full_level_var

@program
def foo(
    input_var,
    k_field,
    nlev,
    full_level_var,
    half_level_var,
    horizontal_start,
    horizontal_end,
    vertical_start,
   vertical_end,
):
    _foo(
        input_var,
        k_field,
        nlev,
        out=(full_level_var, half_level_var),
        domain={
            dims.CellDim: (horizontal_start, horizontal_end),
            dims.KDim: (vertical_start, vertical_end),
        }
    )
)

foo(
    input_var=input_var,
    full_level_var=full_level_var,
    half_level_var=half_level_var,
    k_field=k_field,
    nlev=nlev,
    horizontal_start=start_cell,
    horizontal_end=end_cell,
    vertical_start=0,
    vertical_end=nlev + 1,
    offset_provider={},
)
```
The above example will lead to random behavior when the nvelp1 level is accessed for `full_level_var`.
There are five stencils that need to be fixed:
1. predictor_stencils_2_3
2. predictor_stencils_4_5_6
3. stencils_42_44_45_45b
4. stencils_43_44_45_45b
5. stencils_47_48_49
6. stencils_61_62

### DIFF
--- a/model/atmosphere/dycore/README.md
+++ b/model/atmosphere/dycore/README.md
@@ -4,16 +4,16 @@
 
 Contains code ported from ICON `src/atm_dyn_iconam`, which is the dynamical core of the ICON model.
 
-## Attention
+## Treatment of stencils operating on different vertical levels
 
 Variables in ICON can either sit on full levels or half levels. Let's denote the number of full levels is nlev.
 Full-level variables have nlev levels in vertical dimension, while half-level variables have nlev+1 in vertical dimension.
-Do not output full-level and half-level variables together when running a gt4py program if you need to access the nlev+1 level because it will cause random behavior (most probably illegal memory access).
+When running a GT4Py program over nlev+1 levels, do not output full-level and half-level variables together in a field_operator until the upper vertical bound of nlev+1 because the nlev+1 level will also be written for a full-level variable, which is out-of-bounds.
 The following code is an example:
 
 ```python
 @field_operator
-def foo(
+def _foo(
     input_var,
     k_field,
     nlev,
@@ -36,13 +36,13 @@ def foo(
     horizontal_start,
     horizontal_end,
     vertical_start,
-   vertical_end,
+    vertical_end,
 ):
     _foo(
         input_var,
         k_field,
         nlev,
-        out=(full_level_var, half_level_var),
+        out=(half_level_var, full_level_var),
         domain={
             CellDim: (horizontal_start, horizontal_end),
             KDim: (vertical_start, vertical_end),
@@ -63,6 +63,52 @@ foo(
     offset_provider={},
 )
 ```
+
+Here is a fix to the example above:
+
+```python
+@program
+def foo(
+    input_var,
+    full_level_var,
+    half_level_var,
+    horizontal_start,
+    horizontal_end,
+    vertical_start,
+    vertical_end,
+):
+    _compute_var(
+        input_var,
+        out=(half_level_var, full_level_var),
+        domain={
+            CellDim: (horizontal_start, horizontal_end),
+            KDim: (vertical_start, vertical_end - 1),
+        }
+    )
+    _init_cell_kdim_field_with_zero_vp(
+        out=half_level_var,
+        domain={
+            CellDim: (horizontal_start, horizontal_end),
+            KDim: (vertical_end - 1, vertical_end),
+        }
+    )
+)
+
+foo(
+    input_var=input_var,
+    full_level_var=full_level_var,
+    half_level_var=half_level_var,
+    horizontal_start=start_cell,
+    horizontal_end=end_cell,
+    vertical_start=0,
+    vertical_end=nlev + 1,
+    offset_provider={},
+)
+```
+
+In the fix above, the original field_operator `_foo` is removed and the GT4Py program directly calls `_compute_var` and `_init_cell_kdim_field_with_zero_vp` separately with different vertical bounds.
+The illegal access to nlev+1 level of the full-level variable can thus be avoided by calling `_compute_var` with vertical bounds of (0, nlev).
+The computation on nlev+1 level `half_level_var = where(k_field == nlev, _init_cell_kdim_field_with_zero_vp(), half_level_var)` in the removed field_operator `_foo` is instead performed explicitly by calling `_init_cell_kdim_field_with_zero_vp` with another vertical bounds of (nlev, nlev+1).
 
 ## Installation instructions
 
--- a/model/atmosphere/dycore/README.md
+++ b/model/atmosphere/dycore/README.md
@@ -4,6 +4,66 @@
 
 Contains code ported from ICON `src/atm_dyn_iconam`, which is the dynamical core of the ICON model.
 
+## Attention
+
+Variables in ICON can either sit on full levels or half levels. Let's denote the number of full levels is nlev.
+Full-level variables have nlev levels in vertical dimension, while half-level variables have nlev+1 in vertical dimension.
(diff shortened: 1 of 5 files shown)


OngChia:
cscs-ci run default

model/atmosphere/dycore/README.md
```diff
@@ -4,6 +4,66 @@
 
 Contains code ported from ICON `src/atm_dyn_iconam`, which is the dynamical core of the ICON model.
 
+## Attention
+
+Variables in ICON can either sit on full levels or half levels. Let's denote the number of full levels is nlev.
+Full-level variables have nlev levels in vertical dimension, while half-level variables have nlev+1 in vertical dimension.
+Do not output full-level and half-level variables together when running a gt4py program if you need to access the nlev+1 level because it will cause random behavior (most probably illegal memory access).
```
samkellerhals, line 11: Could you add a short explanation of *why* the random behaviour/out of bounds accesses are caused if we have a stencil where the output fields are half level and full level kdim variables, this isn't clear from the description and might be useful for the reader to understand.

model/atmosphere/dycore/README.md
```diff
@@ -4,6 +4,66 @@
 
 Contains code ported from ICON `src/atm_dyn_iconam`, which is the dynamical core of the ICON model.
 
+## Attention
```
samkellerhals, line 7: I would rename this to something like *Treatment of variables with different vertical levels*

model/atmosphere/dycore/README.md
```diff
@@ -4,6 +4,66 @@
...
+    k_field=k_field,
+    nlev=nlev,
+    horizontal_start=start_cell,
+    horizontal_end=end_cell,
+    vertical_start=0,
+    vertical_end=nlev + 1,
+    offset_provider={},
+)
+```
+
```
samkellerhals, line 66: It might also be useful to include an example *pattern* of how to fix the broken code above, explaining how it works.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
```diff
@@ -280,7 +236,8 @@ def _compute_perturbed_rho_and_potential_temperatures_at_half_and_full_levels(
 
 
 @gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
-def compute_perturbed_rho_and_potential_temperatures_at_half_and_full_levels(
+# def compute_perturbed_rho_and_potential_temperatures_at_half_and_full_levels(
```
samkellerhals, line 239: This comment can be removed?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
```diff
@@ -297,13 +254,12 @@ def compute_perturbed_rho_and_potential_temperatures_at_half_and_full_levels(
...
     z_th_ddz_exner_c: fa.CellKField[float],
     k_field: fa.KField[gtx.int32],
-    nlev: gtx.int32,
     horizontal_start: gtx.int32,
     horizontal_end: gtx.int32,
     vertical_start: gtx.int32,
     vertical_end: gtx.int32,
 ):
-    _compute_perturbed_rho_and_potential_temperatures_at_half_and_full_levels(
+    _compute_pressure_gradient_and_perturbed_rho_and_potential_temperatures(
```
samkellerhals, line 262: Is renaming and removing `nlev` enough here? Do you not need to now call two different field operators with different domains>

samkellerhals (changes requested):
It looks good to me apart from a few improvements in the README. Also for my own understanding, is the strategy to fix this mainly to instead of calling one `field_operator` with the whole domain, to have calls to separate field operators working on different subsets of the domain? Also I haven't checked if there are any other cases of stencils that are affected which you may have missed or not.

model/atmosphere/dycore/README.md
havogt, line 11:
```suggestion
Do not output full-level and half-level variables together when running a GT4Py program over nlev+1 levels because the nlev+1 level will also be written for a full-level variable, which is out-of-bounds.
```
The behavior is not random, only how it is observable. ;)

model/atmosphere/dycore/README.md
OngChia: Thanks for the good name. I will rename it.

model/atmosphere/dycore/README.md
OngChia: Okay. Thanks.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
OngChia: I actually do not know if I still need to split this stencil. I am now testing gpu run. I will respond to you again.

model/atmosphere/dycore/README.md
OngChia: True. I will add how it can be fixed in this example.

model/atmosphere/dycore/README.md
muellch: Yes, it is important here to differentiate between a program and field_operator. If a `field_operator has 2 variables with different output domains (nlev and nlevp1) there will be a problem. However, you can do this in a program, by calling 2 different field_operators for the two variables with different domains, that will work.

muellch (approved):
We should have another long term discussion on how to handle this so that it is as correct and as readable as possible. I would suggest that Chia Rui and me sit down next cycle and go through the dycore and make the changes. We should be able to pay attention to the readability and also the performance impact.

model/atmosphere/dycore/README.md
OngChia: Thanks for the comment. I have modified to emphasize the point where the user cannot output 2 variables with different vertical domains in a `field_operator`.

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro_program.py
OngChia: So, in the end, it is not necessary to split this stencil. I renamed it because the new name reflects correctly the computation done in this combined stencil. As suggested by Christoph, we need to refactor the dycore for better readability including renaming and correctness in maybe next cycle.

samkellerhals: Ok sounds good!

samkellerhals (approved):
LGTM

PR #579, opened 2024-10-30
Run icon4py on gpu

OngChia:
Enable GPU run for icon4py.
Changes:
1. Remove backend and `xp` imports from `common.setting.py`.  The backend is instead added into one of the click options. `xp` is decided explicitly by calling a global function that returns `np` if cpu backends are used or `cp` if gpu backends are used in `icon4py_configuration.py`.
2. Computation in vertical.py is changed to numpy and a backend argument is added in `get_vct_a_and_vct_b` to return the gt4py `vct_a` and `vct_b` fields with correct allocator.
3. A new field_operator `_interpolate_to_half_levels_wp` is used in `compute_virtual_potential_temperatures_and_pressure_gradient.py` to avoid CUDA illegal memory access for an unknown reason.
4. add backend when as_field is called to generate gt4py fields from serialized data. However, xp and backend are still read from `common.setting.py`, it is still not possible to run the driver with cpu backend if you set the environment variable ICON4PY_BACKEND to GPU. This remains to be done in a separate PR.


OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

model/driver/src/icon4py/model/driver/test_cases/gauss3d.py
```diff
@@ -90,13 +95,13 @@ def model_initialization_gauss3d(
...
     end_cell_end = grid.end_index(cell_domain(h_grid.Zone.END))
 
-    w_numpy = xp.zeros((num_cells, num_levels + 1), dtype=float)
-    exner_numpy = xp.zeros((num_cells, num_levels), dtype=float)
-    rho_numpy = xp.zeros((num_cells, num_levels), dtype=float)
-    temperature_numpy = xp.zeros((num_cells, num_levels), dtype=float)
-    pressure_numpy = xp.zeros((num_cells, num_levels), dtype=float)
-    theta_v_numpy = xp.zeros((num_cells, num_levels), dtype=float)
-    eta_v_numpy = xp.zeros((num_cells, num_levels), dtype=float)
+    w_ndarray = xp.zeros((num_cells, num_levels + 1), dtype=float)
```
nfarabullini, line 98:
```suggestion
    w_ndarray = np.zeros((num_cells, num_levels + 1), dtype=float)
```
?

model/driver/src/icon4py/model/driver/test_cases/utils.py
```diff
@@ -6,34 +6,37 @@
...
+    theta_ref_mc: np.ndarray,
+    theta_ref_ic: np.ndarray,
+    rho: np.ndarray,
+    exner: np.ndarray,
+    theta_v: np.ndarray,
     num_levels: int,
-):
+    backend: gt4py_backend.Backend,
+) -> tuple[np.ndarray, np.ndarray, np.ndarray]:
+    xp = driver_config.host_or_device_array(backend)
```
nfarabullini, line 38:
```suggestion
    xp = driver_config.host_or_device_array(gt4py_backend)
```
maybe? not sure

model/driver/src/icon4py/model/driver/test_cases/utils.py
```diff
@@ -130,6 +134,8 @@ def zonalwind_2_normalwind_numpy(
     """
     # TODO (Chia Rui) this function needs a test
 
+    xp = driver_config.host_or_device_array(backend)
```
nfarabullini, line 137: same here as above

nfarabullini (approved):
LGTM! I left a few comments but I'm not sure if they are needed

model/driver/src/icon4py/model/driver/test_cases/gauss3d.py
OngChia: Thanks for the comments. 
I declare the `xp` via `xp = driver_config.host_or_device_array(backend)` and try to scrap the dependency on `xp` in `settings.py`. This is something I am not sure whether it is a good idea to do that or not. Perhaps I should have a pytest fixture xp that switches between cp and np depending whether the backend is on CPU or GPU.

model/common/src/icon4py/model/common/test_utils/serialbox_utils.py
```diff
@@ -453,8 +454,8 @@ def construct_icon_grid(self, on_gpu: bool) -> icon.IconGrid:
         )
         c2e2c = self.c2e2c()
         e2c2e = self.e2c2e()
-        c2e2c0 = xp.column_stack(((range(c2e2c.shape[0])), c2e2c))
-        e2c2e0 = xp.column_stack(((range(e2c2e.shape[0])), e2c2e))
+        c2e2c0 = xp.column_stack((xp.asarray(range(c2e2c.shape[0])), c2e2c))
```
halungge, line 457: you have this `on_gpu` flag as an argument in this function. I don't actually now whether it is always set correctly and we might trade it for a backen argument as well, but you could do the conditional import of the `array_ns` based on that.

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
```diff
@@ -25,7 +25,7 @@ def allocate_zero_field(
     if is_halfdim:
         assert len(shapex) == 2
         shapex = (shapex[0], shapex[1] + 1)
-    return gtx.as_field(dims, xp.zeros(shapex, dtype=dtype), allocator=backend)
+    return gtx.as_field(dims, np.zeros(shapex, dtype=dtype), allocator=backend)
```
halungge, line 28: here you could directly use the `gt4py` function: 
```suggestion
def allocate_zero_field(
    *dims: gtx.Dimension,
    grid,
    is_halfdim=False,
    dtype=ta.wpfloat,
    backend: Optional[backend.Backend] = None,
):
    
    def size(dim:gtx.Dimension, is_half_dim: bool) -> int:
         if dim == dims.KDim and is_half_dim:
            return grid.size[dim] + 1
         else:
            return  grid.size[dim]
        
    dimensions = {d: range(size(d, is_halfdim)) for d in dims }
    return gtx.zeros(dimensions, dtype=dtype, allocator=backend)


```

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -21,6 +32,38 @@
...
+
+
+backend_map: Final[dict] = {
+    DriverBackends.GTFN_CPU: run_gtfn,
+    DriverBackends.GTFN_CPU_CACHED: run_gtfn_cached,
+    DriverBackends.GTFN_GPU: run_gtfn_gpu,
+    DriverBackends.GTFN_GPU_CACHED: run_gtfn_gpu_cached,
+}
+
+
```
halungge, line 49: Something like that I meant above. But I think it is better to have it in `common` and then use the properties of the `gtx.Backend`s `DeviceType` to determine which import to use (see above), because then we can reuse it in all model packages.

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -453,8 +456,28 @@ def initialize(
     default="af122aca-1dd2-11b2-a7f8-c7bf6bc21eba",
     help="uuid of the horizontal grid ('uuidOfHGrid' from gridfile)",
 )
+@click.option(
+    "--enable_output",
+    is_flag=True,
+    help="Enable all debugging messages. Otherwise, only critical error messages are printed.",
+)
+@click.option(
+    "--icon4py_driver_backend",
```
halungge, line 465: as this is a very long name, can you add a short option, like `-b` ?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -835,81 +835,82 @@ def _do_diffusion_step(
...
-            horizontal_start=self._edge_start_nudging,
-            horizontal_end=self._edge_end_halo,
-            vertical_start=(self._grid.num_levels - 2),
-            vertical_end=self._grid.num_levels,
-            offset_provider=self._grid.offset_providers,
-        )
-        log.debug(
-            "running stencils 11 12 (calculate_enhanced_diffusion_coefficients_for_grid_point_cold_pools): end"
-        )
+        if self.config.apply_to_temperature:
```
halungge, line 838: So that stencil should be inside the `if` and it was not before? We never catch these kind of bugs because we allways run the same configuration. So we only ever test on branch of ifs... :-( From this point of view it is very bad that the configuration of EXCLAIM.APE is similar to MCH_CH_R04B09_DSL. We had a discussion on this yesterday. We really should come up with a list of configurations that we want to support and then test all of them and delete what we don't want and test. @anuragdipankar @lxavier @muellch

halungge (changes requested):
Mostly, review the imports in `vertical.py` and I suggest to do what we do now in `host_or_device_array` in `icon4py.common` instead of `icon4py.driver`

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/compute_virtual_potential_temperatures_and_pressure_gradient.py
```diff
@@ -37,7 +40,7 @@ def _compute_virtual_potential_temperatures_and_pressure_gradient(
     wgtfac_c_wp, ddqz_z_half_wp = astype((wgtfac_c, ddqz_z_half), wpfloat)
 
     z_theta_v_pr_ic_vp = _interpolate_to_half_levels_vp(wgtfac_c=wgtfac_c, interpolant=z_rth_pr_2)
-    theta_v_ic_wp = wgtfac_c_wp * theta_v + (wpfloat("1.0") - wgtfac_c_wp) * theta_v(Koff[-1])
+    theta_v_ic_wp = _interpolate_to_half_levels_wp(wgtfac_c=wgtfac_c_wp, interpolant=theta_v)
```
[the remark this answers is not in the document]
OngChia: I always faced cuda illegal memory access error with the inline version when running with resolutions greater than R2B6. I still do not understand why. I am not sure whether it is due to some deep reasons or bugs else where. At least so far for R2B7, everything is okay and results are verified with the new functional call. (As I also indeed like the functional call for interpolation) I tend to keep it this way.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/interpolate_to_half_levels_wp.py
```diff
@@ -0,0 +1,48 @@
...
+from icon4py.model.common.settings import backend
+from icon4py.model.common.type_alias import wpfloat
+
+
+@field_operator
+def _interpolate_to_half_levels_wp(
+    wgtfac_c: fa.CellKField[wpfloat],
+    interpolant: fa.CellKField[wpfloat],
+) -> fa.CellKField[wpfloat]:
+    """Formerly known mo_velocity_advection_stencil_10 and as _mo_solve_nonhydro_stencil_05."""
```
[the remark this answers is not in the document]
OngChia: Agree.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
...
-            else gtx.int32(xp.max(xp.where(vct_a >= flat_height)[0]).item())
+            else gtx.int32(np.max(np.where(vct_a >= flat_height)[0]).item())
         )
 
 
 def _read_vct_a_and_vct_b_from_file(
-    file_path: pathlib.Path, num_levels: int
-) -> tuple[fa.KField, fa.KField]:
+    file_path: pathlib.Path, num_levels: int, backend: gt4py_backend.Backend
+) -> tuple[np.ndarray, np.ndarray]:
```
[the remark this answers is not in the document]
OngChia: Oh no. I forgot to change them back. I was hesitating about whether I should return numpy arrays and create the corresponding gt4py  fields in VerticalGrid. And then I decided to just return the gt4py fields here with an extra backend input argument.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -111,7 +112,7 @@ class VerticalGrid:
     _end_index_of_flat_layer: Final[gtx.int32] = dataclasses.field(init=False)
     _min_index_flat_horizontal_grad_pressure: Final[gtx.int32] = None
 
-    def __post_init__(self, vct_a, vct_b):
+    def __post_init__(self, vct_a: np.ndarray, vct_b: np.ndarray):
```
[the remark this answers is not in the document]
OngChia: I created a new TypeAlias to replace the xp.ndarray as I think that the original xp in setings.py will be moved somewhere  in tools. 
https://github.com/C2SM/icon4py/pull/579/files#diff-6e95c68fb62ed2bc2f706890e79669fc53b7a380a9f4e830cf42c5f9f33e277cR30

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
 
     @classmethod
     def _determine_start_level_of_moist_physics(
-        cls, vct_a: xp.ndarray, top_moist_threshold: float, nshift_total: int = 0
+        cls, vct_a: np.ndarray, top_moist_threshold: float, nshift_total: int = 0
     ) -> gtx.int32:
         n_levels = vct_a.shape[0]
         interface_height = 0.5 * (vct_a[: n_levels - 1 - nshift_total] + vct_a[1 + nshift_total :])
-        return gtx.int32(xp.min(xp.where(interface_height < top_moist_threshold)[0]).item())
+        return gtx.int32(np.min(np.where(interface_height < top_moist_threshold)[0]).item())
```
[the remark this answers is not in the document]
OngChia: In the xp here. I created a new function that takes a numpy or cupy array and deduce the correct xp (np or cp) to be used.
https://github.com/C2SM/icon4py/pull/579/files#diff-504650524ba0d60c795e988ecd8898bee6069cf42bbf9d6dbebeb1e0aa31279eR40

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
 
     @classmethod
     def _determine_start_level_of_moist_physics(
-        cls, vct_a: xp.ndarray, top_moist_threshold: float, nshift_total: int = 0
+        cls, vct_a: np.ndarray, top_moist_threshold: float, nshift_total: int = 0
```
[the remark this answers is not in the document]
OngChia: Sam as above

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
...
     ) -> gtx.int32:
         n_levels = vct_a.shape[0]
         interface_height = 0.5 * (vct_a[: n_levels - 1 - nshift_total] + vct_a[1 + nshift_total :])
-        return gtx.int32(xp.min(xp.where(interface_height < top_moist_threshold)[0]).item())
+        return gtx.int32(np.min(np.where(interface_height < top_moist_threshold)[0]).item())
 
     @classmethod
-    def _determine_damping_height_index(cls, vct_a: xp.ndarray, damping_height: float) -> gtx.int32:
+    def _determine_damping_height_index(cls, vct_a: np.ndarray, damping_height: float) -> gtx.int32:
```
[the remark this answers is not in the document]
OngChia: See my comment above about the new TypeAlias. 
How do you think about that? Maybe there is a better way to do it.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
...
         assert damping_height >= 0.0, "Damping height must be positive."
         return (
             0
             if damping_height > vct_a[0]
-            else gtx.int32(xp.argmax(xp.where(vct_a >= damping_height)[0]).item())
+            else gtx.int32(np.argmax(np.where(vct_a >= damping_height)[0]).item())
```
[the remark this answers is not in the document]
OngChia: Same as above.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
...
     def _determine_end_index_of_flat_layers(
-        cls, vct_a: xp.ndarray, flat_height: float
+        cls, vct_a: np.ndarray, flat_height: float
     ) -> gtx.int32:
         assert flat_height >= 0.0, "Flat surface height must be positive."
         return (
             0
             if flat_height > vct_a[0]
-            else gtx.int32(xp.max(xp.where(vct_a >= flat_height)[0]).item())
+            else gtx.int32(np.max(np.where(vct_a >= flat_height)[0]).item())
```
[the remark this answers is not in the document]
OngChia: Same as above.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -231,36 +232,36 @@ def size(self, dim: gtx.Dimension) -> int:
...
             0
             if damping_height > vct_a[0]
-            else gtx.int32(xp.argmax(xp.where(vct_a >= damping_height)[0]).item())
+            else gtx.int32(np.argmax(np.where(vct_a >= damping_height)[0]).item())
         )
 
     @classmethod
     def _determine_end_index_of_flat_layers(
-        cls, vct_a: xp.ndarray, flat_height: float
+        cls, vct_a: np.ndarray, flat_height: float
```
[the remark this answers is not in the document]
OngChia: See the comment above

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
OngChia: Super thanks!
I did not know that how I could properly create a new dimension until now...

model/driver/src/icon4py/model/driver/icon4py_configuration.py
OngChia: Thanks. I have removed this function and used the one as suggested in your comment above.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
OngChia: Yes, those stencils are inside the `if` statement. The temperature diffusion should be turned off in the standard Jablonowski Williamson test.

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

model/common/src/icon4py/model/common/test_utils/serialbox_utils.py
OngChia: I have decided to use numpy for reading these connectivities from serialized data and used `array_ns` and `on_gpu` to decide between `cp` or `np` in `_get_offset_provider` of `BaseGrid` class. It is enough to make the driver run on gpu or
cpu depending on the backend. (I think all those remaining fields can also be read in as numpy arrays. I am not sure about the strategy, so I did not change the remaining part in `serialbox_utils.py`).

model/driver/src/icon4py/model/driver/test_cases/utils.py
OngChia: I rename the backend module imported from gt4py.next as gt4py_backend to differentiate it from the input argument `backend`. So, `backend` should be correct.

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

PR #580, opened 2024-10-31
Add convergence study in tracer advection

dastrm:
The linear-advection tests in `driver/tests/scientific_validation/linear_tracer_advection_tests.py` read YAML experiment configurations from `experiment_configs` folder. These configurations specify the initial condition, tracer-advection scheme, limiter, driver, vertical grid, and other relevant settings, with unspecified options using their default values.

Each convergence test is performed at three different grid spacings. 

Two advection schemes are tested:
1. `SECOND_ORDER_LINEAR_MIURA`
2. `THIRD_ORDER_PPM`

The convergence tests demonstrate second-order spatial convergence for `SECOND_ORDER_LINEAR_MIURA` and third-order spatial convergence for `THIRD_ORDER_PPM`.

- New `FirstOrderUpwind` horizontal advection class with a new stencil compute_horizontal_tracer_flux_upwind (plus stencil test).
- Enum renames: `LINEAR_2ND_ORDER` → `SECOND_ORDER_LINEAR_MIURA`, `UPWIND_1ST_ORDER` → `FIRST_ORDER_UPWIND`, `PPM_3RD_ORDER` → `THIRD_ORDER_PPM`.
- `AdvectionPrepAdvState` moved from `tracer_advection_states.py` to `model/common/states/tracer_prep_adv_states.py` (as `TracerPrepAdvState` + an `initialize_tracer_prep_adv_state` factory), since it's now allocated by the driver and filled by the initial condition in `common`.
- Docstrings in tracer advection cleaned up.

The multi rank convergence test of the tracer advection will be tackled in another PR.


msimberg:
@dastrm is this something worth keeping open?

dastrm:
Yes, in fact this should have been merged a long time ago. Convergence studies represent an elegant form of verification.

OngChia:
@msimberg @dastrm 
Hi, I am updating this branch. I will move the initial condition to the place where we hold other initial conditions such as JW and warm bubble, and then run the convergence test from the driver as the first part of our effort to implement the scientific validation.

OngChia:
@dastrm 
What resolution did you use in your convergence test? Do you still keep the grids that you generated? If so, where can I get your grids? Is the reason that you first generated the tracer IC on vertices and derive the cell center value using second-order quadrature over a symmetric triangle to get a second-order accurate IC?
I am sorry in advance that I may have more questions coming up when I work through this PR. I sincerely appreciate your help, and I hope you won't mind answering them whenever you have time.

dastrm:
> @dastrm What resolution did you use in your convergence test? Do you still keep the grids that you generated? If so, where can I get your grids? Is the reason that you first generated the tracer IC on vertices and derive the cell center value using second-order quadrature over a symmetric triangle to get a second-order accurate IC? I am sorry in advance that I may have more questions coming up when I work through this PR. I sincerely appreciate your help, and I hope you won't mind answering them whenever you have time.

Hey Chia Rui, I trust you're doing well.

Unfortunately I don't have the five torus grids anymore. These test ran on my laptop and completed in a few minutes maximum. Therefore my best guess is somewhere between 1e4 and 1e5 cells, although you can probably adjust the resolution and tweak the convergence tolerances a bit.

Regarding the initial conditions, even with pure advection there are cases where the convergence is limited by the accuracy of the ICs. Remember that finite volume schemes model cell averages (one constant function per cell) instead of cell center values. To be on the safe side it is good practice to initialize the cell averages with a sufficiently high-order quadrature rule such that the test actually measures the accuracy of the scheme and not of the initial distribution. If I remember correctly this is exactly what is happening here. I used a higher-order quadrature rule for arbitrary triangles which makes use of the values on the vertices.

OngChia:
I am using torus grids with width and height 1km by 1km. The mean edge length is 100, 50, 25, 12.5, 6.25, 3.125, and 1.5625 m. The convergence (L1 norm) rate seems to be only first order. The initial condition is GAUSSIAN_2D. CFL number is 0.2. And the velocity field is constant with magnitude `(Lx, Ly)`, where Lx and Ly are domain length and height, respectively. The integration time is 0.5 s. I am going to investigate whether this is because I removed the part where the domain length and height are fixed when calculating the initial condition. But it is hard for me to believe that a slight difference in the domain size will have any significant impact on the convergence rate. Otherwise, if we need higher-order quadrature. @dastrm do you have any ideas why?
image

OngChia:
When measuring the error with L_inf norm, the convergence rate is slightly worse.
image

dastrm:
> I am using torus grids with width and height 1km by 1km. The mean edge length is 100, 50, 25, 12.5, 6.25, 3.125, and 1.5625 m. The convergence (L1 norm) rate seems to be only first order. The initial condition is GAUSSIAN_2D. CFL number is 0.2. And the velocity field is constant with magnitude `(Lx, Ly)`, where Lx and Ly are domain length and height, respectively. The integration time is 0.5 s. I am going to investigate whether this is because I removed the part where the domain length and height are fixed when calculating the initial condition. But it is hard for me to believe that a slight difference in the domain size will have any significant impact on the convergence rate. Otherwise, if we need higher-order quadrature. @dastrm do you have any ideas why? image

First of all, the resolution looks good to me. You should be able to see the second-order accuracy with these grids, and moreover you would probably not even be able to tell the results apart from the `p = 2.0` line below an error of ~1e-2.

Now that we know what to expect, obviously something is fishy here. As always it could be anything, namely the scheme itself or the test setup. For now I'd assume the latter and start with plotting the initial conditions, the solution and maybe also the error to double-check that you're actually calculating what you think you are. A visual inspection could help you rule out trivial mistakes such as too large ICs that would introduce a significant kink in the solution due to the periodicity of the grid. Given that you have some kind of convergence in the infinity norm already, it may be only a small thing that needs adjustment.

OngChia:
@dastrm I still have no concrete evidence on what went wrong. Did you have a report on your horizontal convergence test? Could you send me the report of if there is one?

OngChia:
@dastrm Thanks for your hints!! It was really helpful to look at the plots. Good news is that the problems have been resolved. It was my fault after all. 1) Indeed the decay factor for the Gaussian shape was too large and a kink appears in the domain. 2.) I fixed the CFL number to compute dtime. However, the driver rounds off the number of time steps because the integration time is not an integer multiple of dtime. And I used the integration time to compute the reference solution. It then introduced a first-order error because the actual integration time is different from what I specified at the beginning. After fixing that, the convergence rate is correct.

OngChia:
cscs-ci run default;LEVELS=validation;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBPACKAGES=driver;SESSIONS=model

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
```diff
@@ -480,6 +486,117 @@ def _update_unknowns(
     ) -> None: ...
 
 
+class FirstOrderUpwind(FiniteVolume):
```
nfarabullini, line 494: earlier you have `HorizontalAdvectionType.UPWIND_1ST_ORDER`, now `FirstOrderUpwind`. Can you choose one order of words?

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
```diff
@@ -480,6 +486,117 @@ def _update_unknowns(
...
+    """Class that does one horizontal first-order accurate upwind finite volume advection step."""
```
nfarabullini:
```suggestion
    """Perform one horizontal first-order accurate upwind finite volume advection step."""
```

model/atmosphere/tracer_advection/tests/tracer_advection/stencil_tests/test_compute_horizontal_tracer_flux_upwind.py
```diff
@@ -0,0 +1,53 @@
...
+
+from icon4py.model.atmosphere.tracer_advection.stencils.compute_horizontal_tracer_flux_upwind import (
+    compute_horizontal_tracer_flux_upwind,
+)
+from icon4py.model.common import dimension as dims
+from icon4py.model.common.utils import data_allocation as data_alloc
+from icon4py.model.testing import stencil_tests
+
+
+class TestComputeHorizontalTracerFluxUpwind(stencil_tests.StencilTest):
```
nfarabullini, line 21: a new PR should be merged soon with stencil tests conventions: https://github.com/C2SM/icon4py/pull/1116/changes#diff-77c24472f3c343e888a4d4b3f782bcf75d4bf6b4caee6ac34e4ef622478bddea

Make sure this test is consistent

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -65,12 +65,12 @@ def unstructure_abstime(abstime: time.AbsoluteTime) -> str:
 def structure_reltime(reltime_val: str, _: typing.Any) -> time.RelativeTime:
     if isinstance(reltime_val, time.RelativeTime):
         return reltime_val
-    return time.RelativeTime(seconds=int(reltime_val))
+    return time.RelativeTime(seconds=float(reltime_val))
```
nfarabullini, line 68: why are you changing this type from int to float?

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_advection.py
```diff
@@ -0,0 +1,447 @@
...
+
+
+@config_io.register_enum
+class VelocityField(int, enum.Enum):
+    """
+    Velocity field for idealized advection test cases.
+    """
+
+    #: constant velocity field in x and y directions
+    CONSTANT = 1
```
nfarabullini: maybe you a different name than just `CONSTANT`, it's a broadly used term

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_advection.py
```diff
@@ -0,0 +1,447 @@
...
+def _construct_idealized_tracer(
+    *,
+    config: LinearHorizontalAdvectionConfig,
+    tracer: data_alloc.NDArray,
+    tracer_center_x: float,
+    tracer_center_y: float,
+    domain_length: float,
+    domain_height: float,
+    weights: data_alloc.NDArray,
+    nodes: data_alloc.NDArray,
```
nfarabullini: This is an old thing, but we usually place all arrays first and then scalars

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_tracer_advection.py
```diff
@@ -0,0 +1,447 @@
...
+            vertex_tracer = array_ns.exp(-decay_factor * (dx**2))
+        case TracerProfile.GAUSSIAN_1D_Y:
+            decay_factor = -1.0 / (config.decay_radius * domain_height) ** (2) * math.log(1e-3)
+            vertex_tracer = array_ns.exp(-decay_factor * (dy**2))
+        case _:
+            raise NotImplementedError(
+                f"Initial tracer profile {config.tracer_profile} not implemented."
+            )
+    tracer[:, :] = array_ns.sum(weights * vertex_tracer, axis=0)[:, None]
+
```
nfarabullini, line 335: you don't return anything in this function?

model/common/src/icon4py/model/common/initial_condition/analytical/linear_vertical_tracer_advection.py
```diff
@@ -0,0 +1,241 @@
...
+from icon4py.model.common.grid import vertical as v_grid
+from icon4py.model.common.metrics import metrics_attributes as metrics_meta
+from icon4py.model.common.states import adv_states, prognostic_state as prognostics, tracer_states
+from icon4py.model.common.utils import data_allocation as data_alloc
+
+
+if TYPE_CHECKING:
+    from icon4py.model.common.metrics import metrics_factory
+
+
```
nfarabullini, line 31: VelocityField and TracerProfile are classes already defined in `model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_advection.py`
so redefining classes with the same name here will lead to confusion. How about, e.g. `VTracerProfile` and `HTracerProfile` for the other?

nfarabullini: same for VelocityField and _compute_idealized_velocity_field

model/common/src/icon4py/model/common/states/prep_adv_states.py
```diff
@@ -0,0 +1,62 @@
...
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""
+AdvectionPrepAdvState for tracer advection.
+
+It is allocated by the driver, filled by the initial condition and mutated by the
+tracer advection, so it lives here and not in the tracer advection package.
+"""
+
```
nfarabullini, line 15: use `tracer_advection`, not just `advection`. We had a PR a few weeks ago exactly tackling this different nomenclature as simply `advection` can be confused with velocity advection

model/driver/src/icon4py/model/driver/driver.py
```diff
@@ -279,7 +280,7 @@ def _integrate_one_time_step(
         prognostic_states: common_utils.TimeStepPair[prognostics.PrognosticState],
         tracers: common_utils.TimeStepPair[tracer_states.TracerState],
         prep_adv: dycore_states.PrepAdvection | None,
-        tracer_prep_adv: tracer_advection_states.AdvectionPrepAdvState | None,
+        tracer_prep_adv: adv_states.AdvectionPrepAdvState | None,
```
nfarabullini: same for all of these imports: `adv_states` --> `tracer_adv_states`
```suggestion
        tracer_prep_adv: tracer_adv_states.AdvectionPrepAdvState | None,
```

model/driver/tests/driver/integration_tests/test_initial_conditions.py
```diff
@@ -128,6 +128,7 @@ def test_initial_conditions(
         prognostic_state_now=prognostic_state_now,
         tracer_state_now=tracer_state_now,
         solve_nonhydro_diagnostic_state=None,
+        adv_prep_adv_state=None,
```
nfarabullini: adv prep adv?

model/driver/tests/driver/scientific_validation/linear_advection_tests.py
```diff
@@ -0,0 +1,350 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
+#
+# Copyright (c) 2022-2024, ETH Zurich and MeteoSwiss
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
```
nfarabullini, line 8: please rename to specify which advection this is

model/common/src/icon4py/model/common/states/prep_adv_states.py
nfarabullini: for reference: https://github.com/C2SM/icon4py/pull/1399

model/driver/tests/driver/unit_tests/test_driver_states.py
```diff
@@ -11,63 +11,76 @@
...
     assert (
-        prep_tracer_adv.mass_flx_ic is prep_adv.dynamical_vertical_mass_flux_at_cells_on_half_levels
+        dycore_prep_adv.dynamical_vertical_mass_flux_at_cells_on_half_levels
+        is adv_prep_adv_state.mass_flx_ic
     )
 
 
-def test_prep_tracer_advection_is_none_when_disabled(grid: base.Grid) -> None:
+@pytest.mark.parametrize(
+    "solve_nonhydro_enabled, with_adv_prep_adv_state",
```
nfarabullini: why do you parametrize `solve_nonhydro_enabled` is it's always False? You can also define it directly inside the test

nfarabullini (changes requested):

nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_advection.py
OngChia: Let me think a better name

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_tracer_advection.py
OngChia: Let me investigate

model/common/src/icon4py/model/common/initial_condition/analytical/linear_vertical_tracer_advection.py
OngChia: Okay!

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_tracer_advection.py
OngChia: I have renamed the function to `_fill_tracer_from_analytical_profile` and the argument tracer to `tracer_buffer` to clearly indicate we are filling some values into the array buffer.

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
OngChia: I found out that David already named other classes in this order while the `HorizontalAdvectionType` in the other way round (for example, `SecondOrderMiura`, `LINEAR_2ND_ORDER` ). I think usually we first say the order of accuracy first and then the scheme's name. So, I will prefer this ordering. But, I need to modify other others too for consistency. For example, `MIURA_2ND_ORDER` -> `SECOND_ORDER_LINEAR_MIURA`, `PPM_3RD_ORDER` -> `THIRD_ORDER_PPM`. How do you think about these names?
Maybe @jcanton should also give some opinions.
I do not have opinions about the naming here.

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_advection.py
OngChia: I rename it to UNIFORM_XY.  So, something like `if velocity_field == VelocityField.UNIFORM_XY` looks more understandable to me that we have a uniform wind field in both x and y directions. Feel free to suggest any other names if you find this is not okay.

model/common/src/icon4py/model/common/initial_condition/analytical/linear_vertical_tracer_advection.py
OngChia: I added horizontal or vertical in front.

model/driver/tests/driver/integration_tests/test_initial_conditions.py
OngChia: Changed it to tracer_prep_adv

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
nfarabullini: I'm fine with your preference

model/common/src/icon4py/model/common/states/prep_adv_states.py
nfarabullini: @OngChia can you also rename the file to include `tracer`?

nfarabullini (approved):

experiment_configs/linear_vertical_advection_box.yaml
```diff
@@ -0,0 +1,39 @@
+geometry: {}
```
msimberg, line 1: Just starting the review, but it feels like these should not be at the top level. Possibly somewhere closer to the testing module and closer to the experiment fixture definitions.

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_vertical.py
```diff
@@ -474,12 +475,17 @@ def run(
 
 
 class FiniteVolume(VerticalAdvection):
-    """Class that defines a finite volume vertical tracer_advection scheme."""
+    """
+    Abstract base class for finite-volume vertical tracer advection schemes.
```
msimberg, line 479: Not abstract base class? It has implementations?

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -65,12 +65,12 @@ def unstructure_abstime(abstime: time.AbsoluteTime) -> str:
...
 
 
 @CONV.register_unstructure_hook
-def unstructure_reltime(reltime: time.RelativeTime) -> int:
-    return int(reltime.total_seconds())
+def unstructure_reltime(reltime: time.RelativeTime) -> float:
```
msimberg, line 72: Just curious: what was the motivation for the changes in this file?

model/common/src/icon4py/model/common/initial_condition/analytical/linear_vertical_tracer_advection.py
```diff
@@ -0,0 +1,257 @@
...
+    model_top_height: float,
+) -> float:
+    # note: as we need vel_max at time n+1/2 and vel_max is needed for the time step, we have a chicken-and-egg problem
+    # instead of doing a fixed-point iteration, we simply estimate an upper bound for vel_max
+
+    w = _compute_idealized_vertical_velocity_field(
+        velocity_field=velocity_field,
+        model_top_height=model_top_height,
+    )
+    return data_alloc.array_namespace(w).max(data_alloc.array_namespace(w).abs(w))
```
msimberg: Maybe add array_ns as a local variable. But also, there may be methods that you can use instead and do w.abs().max(). Not 100% sure if they exist, in which case that's obviously not an option...

model/common/src/icon4py/model/common/states/prep_adv_states.py
```diff
@@ -0,0 +1,62 @@
...
+    vn_traj: fa.EdgeKField[ta.wpfloat]
+
+    #: mass flux at full level edges averaged over dynamics substeps [kg/m^2/s]
+    mass_flx_me: fa.EdgeKField[ta.wpfloat]
+
+    #: mass flux at half level centers averaged over dynamics substeps [kg/m^2/s]
+    mass_flx_ic: fa.CellKField[ta.wpfloat]
+
+
+def initialize_tracer_prep_adv_state(
```
msimberg: Nit: I think we can use prepare advection state generally instead of prep adv state (file names, classes, functions, etc.). But I know the short names already existed, so not a must to change in this PR.

model/common/tests/common/config/unit_tests/test_config_io.py
```diff
@@ -119,7 +119,7 @@ def test_write_yaml_str_read_yaml_str_roundtrip() -> None:
             time.AbsoluteTime,
             time.AbsoluteTime(year=2026, month=7, day=30, hour=14, minute=41, second=25),
         ),
-        ("300\n...\n", time.RelativeTime, time.RelativeTime(seconds=300)),
+        ("300.0\n...\n", time.RelativeTime, time.RelativeTime(seconds=300)),
```
msimberg, line 122: Related to other questions, why are these now floats?

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
```diff
@@ -0,0 +1,350 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
```
msimberg, line 1: I still need to get a slightly better overview of this PR, but we should at the very least be calling this directory `somnething_tests` to be consistent with others, and so that they're easy to find and glob for testing. Simplest would likely be `validation_tests` to mirror `unit_tests` and `integration_tests`.

model/testing/src/icon4py/model/testing/fixtures/torus_grids.py
```diff
@@ -0,0 +1,71 @@
...
+            edge_length=edge_length,
+            periodic_layout="rectangular",
+        )
+    )
+    grid.to_netcdf(out_file)
+    return out_file
+
+
+@pytest.fixture(scope="session")
+def generate_torus_grid(tmp_path_factory: pytest.TempPathFactory) -> Callable[..., pathlib.Path]:
```
msimberg, line 45: I have some concerns about this setup.

How flexible do you need to be with different torus grids? On one hand I wonder if we can avoid creating the torus grid file at all, and only keep it in memory. I don't know if the grid generator has a documented API that doesn't create nc files. That'd be the cleanest if you really want to be able to easily generate new test cases on the fly. On the other hand, if you don't need the flexibility, I think it'd probably be worth sticking with the existing path of GridDescription and uploading pre-generated grids uploaded to the test data storage since it avoids divergence. I'm not saying either of these is strictly better, but should be considered. We can discuss more in person as well.

model/testing/src/icon4py/model/testing/config.py
```diff
@@ -30,6 +30,9 @@ def _default_download_cache() -> pathlib.Path:
 ENABLE_GRID_DOWNLOAD: bool = env.flag_to_bool("ICON4PY_ENABLE_GRID_DOWNLOAD", True)
 ENABLE_TESTDATA_DOWNLOAD: bool = env.flag_to_bool("ICON4PY_ENABLE_TESTDATA_DOWNLOAD", True)
 TEST_DATA_PATH: pathlib.Path = env.path("ICON4PY_TEST_DATA_PATH", _project_root() / "testdata")
+EXPERIMENT_CONFIG_PATH: pathlib.Path = env.path(
+    "ICON4PY_EXPERIMENT_CONFIG_PATH", _project_root() / "experiment_configs"
+)
```
msimberg, line 35: It feels weird that this is overridable via an env var. Why is it?

model/testing/pyproject.toml
```diff
@@ -32,6 +32,7 @@ dependencies = [
   # safe or if this is a bug in filelock.
   "filelock>=3.18.0,<3.20",
   "gt4py==1.2.1",
+  "icon-grid-generator[netcdf]>=0.8.0",
```
msimberg, line 35: The big question is if we want to have this as a hard dependency now. Nothing per-se against it, but wondering if we can avoid it. See other comment about the `generate_torus_grid` fixture as well.

msimberg (commented):
Some preliminary comments. Some of the testing comments may need a bit more discussion in person.

model/common/src/icon4py/model/common/states/prep_adv_states.py
OngChia: And I forgot to remove this file... There is another file `tracer_prep_adv_states.py` in the same folder.
The `tracer_prep_adv_states.py` is the one that is actually in use.

model/common/src/icon4py/model/common/config/config_io.py
OngChia: In my test, the `dtime` is a float. I fixed the total integration time, and used maximum cfl number to compute the `dtime`. That is why I changed it to `float`.

model/common/src/icon4py/model/common/initial_condition/analytical/linear_vertical_tracer_advection.py
OngChia: There is, I only know that `array_ns.linalg.norm(w.ravel(), ord=array_ns.inf)` is equivalent to abs max of an array. Although it is clear from the function itself that we are taking the norm of a vector with `Linf`, it definitely looks more opaque than `array_ns.abs(w).max().item()`. I still prefer writing the abs max explicitly. 
I added `array_ns` as a local variable.

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
OngChia: Good point. I agree that we should be consistent and use the **_tests naming convention.
I remember that convergence tests are usually categorized as verification tests. However, since the convergence tests are more computationally demanding, we currently mark them as validation tests. Do you think there would be a better way to categorize them, or is this distinction not really a concern here?

jcanton:
@OngChia small note on the torus grids: the fixture currently writes a NetCDF file and `GridManager` reads it back.

Correcting what I wrote here earlier — I had assumed the ICON schema mapping only existed inside the generator's private NetCDF writer, so that in-memory ingestion would need a translation layer. That's wrong. `IconGrid` already exposes all nine geometry fields under their exact ICON names, and all eight connectivities in `neighbor_tables` in icon4py's own convention (0-based, `-1` for missing, `int32`, entity-major), plus the global attributes and refinement control fields. The only differences from the on-disk form are degrees vs radians and packed `(n, 3)` cartesians — both trivial on our side.

What's missing is that none of this is documented or declared stable, so https://github.com/ofuhrer/icon-grid-generator/issues/6 now just asks for that (plus a couple of small fixes). Nothing there blocks us: domain decomposition, halo construction and derived neighbour tables belong to `GridManager` anyway, so the icon4py-side change is to let it take arrays as an alternative to a file path.

Keeping the file round-trip on this PR regardless — it costs ~0.02 s per grid, so nothing here is urgent.

model/testing/src/icon4py/model/testing/experiment_configs/linear_vertical_advection_box.yaml
```diff
@@ -0,0 +1,39 @@
+geometry: {}
```
OngChia: Agreed. Moved to `testing`

model/testing/src/icon4py/model/testing/config.py
OngChia: Agreed! I removed the override support

model/testing/pyproject.toml
OngChia: As agreed in our discussion, I remove icon-grid-generator from this PR.

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_vertical.py
OngChia: Not 100% sure here, but my understanding is that it still remains abstract because it declares abstract methods
```
@abc.abstractmethod
    def _compute_numerical_flux(

@abc.abstractmethod
    def _update_unknowns(
```

msimberg: Ah yes, you're right. Sorry for the noise.

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
msimberg: In the end we might either just distinguish between verification and validation directly or have an independent axis which is "how heavy is this test to run" which is anyway the important part for deciding whether or not to run it in CI on every PR. Let's stick with validation for now?

However, can you check if validation_tests will now be included in test discovery in noxfile.py? I.e. do these tests actually end up running in the weekly CI?

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
```diff
@@ -0,0 +1,337 @@
...
+
+    error_l1: list[float] = []
+    error_linf: list[float] = []
+    mean_edge_length: list[float] = []
+
+    config_path = test_config.EXPERIMENT_CONFIG_PATH / f"{experiment_case}.yaml"
+
+    experiment_config = config_io.read_yaml_str(
+        config_path.read_text(), driver_config.ExperimentConfig
+    ).with_overrides(driver={"output_path": tmp_path / "ci_driver_output"})
```
msimberg: I think you could at least put this logic in a helper function, if not in a test fixture.

model/common/src/icon4py/model/common/config/config_io.py
msimberg: Hmm, would it not make more sense to cast to float where you use it if you need to have fractional times? If the times from the config are always integer seconds, then we should keep the type as int to keep that promise?

msimberg (commented):
Looks very nice. Still a few minor questions.

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBPACKAGES=driver;SESSIONS=model

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
OngChia: I created `conftest.py` in `validation_tests` and put the fixture there, because I think that other scientific validation tests can also use this fixture in the future, and it will only be used by those tests.

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
OngChia: Totally agree with the point that "how heavy a test is". Okay. Let's stick with validation_tests.

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

model/common/src/icon4py/model/common/config/config_io.py
OngChia: I think times in the config should be allowed to be fractional. It just happens that most of the test cases use an integer time step. This is probably users are running simulations with grid spacings ~100 m to 100 km. If we run icon4py with dx=1 ~ 10 m, we may need to set the time step less than 1 s. 
Or should we let users to choose the unit, let's say millisecond?

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

OngChia:
cscs-ci run default;BACKENDS=gtfn_gpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

model/driver/tests/driver/validation_tests/test_tracer_advection_convergence.py
OngChia: I ran cscs-ci default on back gpu and cpu backends. They both capture the tests and the tests pass. I assume the weekly CI will also capture the validation tests.

model/common/src/icon4py/model/common/config/config_io.py
OngChia: So, following youe suggestion, I added unit to dtime so that users can specify seconds, milliseconds, or microseconds. dtime can be kept as an int as before.

OngChia:
cscs-ci run default;BACKENDS=gtfn_cpu;LEVELS=validation;MODEL_SUBSETS=basic;MODEL_SUBPACKAGES=driver;SESSIONS=model

model/common/tests/common/config/unit_tests/test_config_io.py
```diff
@@ -144,7 +144,9 @@ def test_write_yaml_str_read_yaml_str_roundtrip() -> None:
             time.AbsoluteTime(year=2026, month=7, day=30, hour=14, minute=41, second=25),
             id="abstime",
         ),
-        pytest.param("300\n...\n", time.RelativeTime, time.RelativeTime(seconds=300), id="reltime"),
+        pytest.param(
+            "300 seconds\n...\n", time.RelativeTime, time.RelativeTime(seconds=300), id="reltime"
```
msimberg, line 148: Nice, having the units there is a welcome addition!

msimberg (approved):

model/common/src/icon4py/model/common/initial_condition/analytical/linear_horizontal_tracer_advection.py
```diff
@@ -0,0 +1,453 @@
...
+        primal_normal_x=geometry.get(geometry_meta.EDGE_NORMAL_U).ndarray,
+        primal_normal_y=geometry.get(geometry_meta.EDGE_NORMAL_V).ndarray,
+        domain_length=grid.grid_params.domain_length,
+        domain_height=grid.grid_params.domain_height,
+    )
+
+    center_x, center_y = _compute_tracer_center(
+        initial_center=config.initial_center,
+        origin_x=vertex_x.min(),
+        origin_y=vertex_y.min(),
```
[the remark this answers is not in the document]
OngChia: This is known. It will be tackled at next PR when I create parallel convergence tests.

model/driver/src/icon4py/model/driver/driver.py
```diff
@@ -827,6 +828,13 @@ def run_driver(
         if icon4py_driver.config.nonhydrostatic is not None
         else None
     )
+    tracer_prep_adv_state = (
+        prep_adv_states.initialize_tracer_prep_adv_state(
+            grid=icon4py_driver.grid, allocator=allocator
+        )
+        if icon4py_driver.config.tracer_advection is not None
+        else None
```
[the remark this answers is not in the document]
OngChia: No, it is not a problem. No initial conditions use prep_adv except the new convergence tests. If tracer is not turned on, the initial condition function for the convergence tests will throw an value error telling users that tracer advection should be turned on.

model/driver/tests/driver/mpi_tests/test_parallel_initial_conditions.py
```diff
@@ -180,6 +186,9 @@ def test_initial_conditions_compare_single_multi_rank(  # noqa: PLR0917 [too-man
     multi_rank_dycore_diagnostic = nonhydro_states.initialize_solve_nonhydro_diagnostic_state(
         grid=multi_rank_icon4py_driver.grid, allocator=allocator
     )
+    multi_rank_tracer_prep_adv_state = prep_adv_states.initialize_tracer_prep_adv_state(
+        grid=single_rank_icon4py_driver.grid, allocator=allocator
```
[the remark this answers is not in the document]
OngChia: This PR is for serial convergence tests. Parallel run will be tackled in the next PR.

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection.py
```diff
@@ -457,7 +460,14 @@ def convert_config_to_horizontal_vertical_advection(  # noqa: PLR0912 [too-many-
...
                 grid=grid, backend=backend
             )
-        case HorizontalAdvectionType.LINEAR_2ND_ORDER:
+        case HorizontalAdvectionType.FIRST_ORDER_UPWIND:
+            horizontal_advection = tracer_advection_horizontal.FirstOrderUpwind(
+                grid=grid,
+                interpolation_state=interpolation_state,
+                metric_state=metric_state,
+                backend=backend,
+            )
```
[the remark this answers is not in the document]
OngChia: Again. This is not possible in this PR. But definitely need to be well tested in a subsequent PR.

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -141,12 +141,19 @@ def unstructure_abstime(abstime: time.AbsoluteTime) -> str:
...
-def unstructure_reltime(reltime: time.RelativeTime) -> int:
-    return int(reltime.total_seconds())
+def unstructure_reltime(reltime: time.RelativeTime) -> str:
+    seconds = reltime.total_seconds()
+    if seconds == int(seconds):
+        return f"{int(seconds)} seconds"
+    elif (milliseconds := seconds * 1000) == int(milliseconds):
+        return f"{int(milliseconds)} milliseconds"
+    else:
+        return f"{int(seconds) * 1_000_000} microseconds"
```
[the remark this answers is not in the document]
OngChia: Good catch!!

model/testing/src/icon4py/model/testing/config.py
```diff
@@ -30,6 +30,9 @@ def _default_download_cache() -> pathlib.Path:
 ENABLE_GRID_DOWNLOAD: bool = env.flag_to_bool("ICON4PY_ENABLE_GRID_DOWNLOAD", True)
 ENABLE_TESTDATA_DOWNLOAD: bool = env.flag_to_bool("ICON4PY_ENABLE_TESTDATA_DOWNLOAD", True)
 TEST_DATA_PATH: pathlib.Path = env.path("ICON4PY_TEST_DATA_PATH", _project_root() / "testdata")
+EXPERIMENT_CONFIG_PATH: pathlib.Path = (
+    pathlib.Path(__file__).resolve().parent / "experiment_configs"
+)
```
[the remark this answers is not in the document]
OngChia: Added

PR #575, opened 2024-10-24
settings elements migration

nfarabullini:
- moved the following elements from `model/common/src/icon4py/model/common/settings.py`to `tools/src/icon4pytools/py2fgen/wrappers/settings.py`:
 ```
config = Icon4PyConfig()
backend = config.gt4py_runner
dace_orchestration = config.icon4py_dace_orchestration
xp = config.array_ns
device = config.device
limited_area = config.limited_area
```
- subsequent edit of imports and related in the code
- cleanup of left-over stencils backend specification on program definition

PR #582, opened 2024-11-01
[Py2F]: Parallel diffusion fixes

samkellerhals:
Fixes for parallel cpu/gpu runs in diffusion granule.


samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -227,9 +229,14 @@ def exchange(self, dim: definitions.Dimension, *fields: Sequence[Field]):
         # Slice the fields based on the dimension
         sliced_fields = [self._slice_field_based_on_dim(f, dim) for f in fields]
 
+        # todo: `arch` Required until GHEX make_field_descriptor function is fixed with arch not having any defaults.
+        #   Currently the default is Architecture.CPU and this breaks passing in cupy arrays without the arch
+        #   keyword as it will do checks as if was a numpy array.
+        arch = Architecture.GPU if device == Device.GPU else Architecture.CPU
```
halungge, line 235: @nfarabullini is currently working towards remove the settings.py from model in this (https://github.com/C2SM/icon4py/pull/575)[PR 575](it will be moved to `tools/py2fgen`). So please do not make her task even harder by using it once more... 

I don't understand why you need it at all: 
GHEX does this: 
```
    if not arch:
        if hasattr(field, "__cuda_array_interface__") or hasattr(field, "__hip_array_interface__"):
            arch = Architecture.GPU
        elif hasattr(field, "__array_interface__"):
            arch = Architecture.CPU
        else:
            raise ValueError()

    if arch == Architecture.CPU:
        assert hasattr(field, "__array_interface__")
    elif arch == Architecture.GPU:
        assert hasattr(field, "__cuda_array_interface__") or hasattr(
            field, "__hip_array_interface__"
        )

```
If I understand that code correctly you don't need to pass the `arch`. If you don't pass it, it will be determined based on the arrays you pass. If you pass it it query the very same attributes in the assert. So if that fails it would fail anyway regardless of wether you passed the argument. (Except for when you optimize away the `assert` .)

halungge: How do you build ghex btw? You need to enable the GPU build there: https://github.com/ghex-org/GHEX/blob/master/bindings/python/pyproject.toml

samkellerhals: Because as I mentioned in my comment the arch keyword argument in the `make_field_descriptor` function has a default of `Architecture.CPU`, so if I do not explicitly pass `arch` when running on GPU it will check for the `__array_interface__` attribute, which cupy arrays do not have, and thus the code crashes.

```python
def make_field_descriptor(
    domain_desc: DomainDescriptor,
    field: NDArray,
    *,
    arch: Optional[Architecture] = Architecture.CPU,
) -> Any:
```

So what I propose to do is instead of using `device`, is to just determine the architecture based on whether it is a cupy or numpy array that is being passed in.

samkellerhals: And yes we do enable the GPU build as follows which works nicely:

```bash
export GHEX_USE_GPU=ON
export GHEX_GPU_TYPE=NVIDIA
export GHEX_GPU_ARCH="70;80"
```

I've pushed a new commit which I think should work for everybody now.

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

samkellerhals:
launch jenkins spack

samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

samkellerhals:
launch jenkins spack

halungge (approved):
LGTM

PR #586, opened 2024-11-05
Grid geometry

halungge:
Computation and loading of geometry fields:
- adds computation of grid geometry fields needed in the `SolveNonHydro` and `Diffusion`
- geometry fields for the sphere are computed from the `lat`and `lon` coordinates for `edge`, `vertex` and `cell`s read from the grid file. Exceptions for this being: 
     -  the orientation of tangent vectors along `edge`s 
     - areas of cells - (there is a bug in the computation of this field in the grid generator, which has recently been fixed, but is still contained in our sample grid files)
that are read directly read from the grid file. 

As a first example `test_diffusion.py` uses those computed fields as in put instead of the previousely used serialized fields from the `grid_savepoint`.


halungge:
cscs-ci run default

model/common/tests/grid_tests/test_geometry.py
```diff
@@ -0,0 +1,402 @@
...
+        edge_indices = alloc.allocate_indices(dims.EdgeDim, grid)
+        owner_mask = np.ones((grid.num_edges,), dtype=bool)
+        decomposition_info = definitions.DecompositionInfo(klevels=grid.num_levels)
+        decomposition_info.with_dimension(dims.EdgeDim, edge_indices.ndarray, owner_mask)
+        return decomposition_info
+
+    def construct_grid_geometry(grid_file: str):
+        gm = utils.run_grid_manager(grid_file)
+        grid = gm.grid
+        decomposition_info = construct_decomposition_info(grid)
```
nfarabullini, line 42: since `construct_decomposition_info` is only used here, would it make sense to merge these two functions?

model/common/tests/grid_tests/test_grid_manager.py
```diff
@@ -21,233 +19,50 @@
...
     try:
         parser.open()
-        assert parser.dimension(gm.DimensionName.CELL_NAME) == grid.num_cells
-        assert parser.dimension(gm.DimensionName.VERTEX_NAME) == grid.num_vertices
-        assert parser.dimension(gm.DimensionName.EDGE_NAME) == grid.num_edges
+        assert parser.dimension(gm.DimensionName.CELL_NAME) == R02B04_GLOBAL_NUM_CELLS
+        assert parser.dimension(gm.DimensionName.VERTEX_NAME) == R02B04_GLOBAL_NUM_VERTEX
+        assert parser.dimension(gm.DimensionName.EDGE_NAME) == R02B04_GLOBAL_NUM_EDGES
     except Exception:
         pytest.fail()
```
nfarabullini, line 67: maybe add a message here

model/common/tests/grid_tests/utils.py
```diff
@@ -48,6 +51,9 @@ def resolve_file_from_gridfile_name(name: str) -> Path:
                 r02b04_global_data_file,
             )
         return gridfile
+    elif name == "0055_R02B05":
```
nfarabullini, line 54: this is a very specific name. I think we should define a string above with this name and then use it here since it may change in the future

nfarabullini: Not that it changes anything in terms of code functionality, but it is best practice

model/common/tests/grid_tests/utils.py
```diff
@@ -92,3 +98,19 @@ def valid_boundary_zones_for_dim(dim: dims.Dimension):
...
+        transformation = gm.ToZeroBasedIndexTransformation()
+    file_name = resolve_file_from_gridfile_name(experiment_name)
+    with gm.GridManager(
+        transformation, file_name, v_grid.VerticalGridConfig(num_levels)
+    ) as grid_manager:
+        grid_manager(limited_area=is_local(experiment_name))
+        return grid_manager
+
+
+def is_local(grid_file: str):
```
nfarabullini, line 115: should we call it `is_regional`?

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -0,0 +1,476 @@
...
+        }
+
+    def _validate_dependencies(self):
+        func_signature = inspect.signature(self._func)
+        parameters = func_signature.parameters
+        for dep_key in self._dependencies.keys():
+            parameter_definition = parameters.get(dep_key)
+            assert parameter_definition.annotation == xp.ndarray, (
+                f"Dependency {dep_key} in function {self._func.__name__}:  does not exist or has "
+                f"or has wrong type ('expected np.ndarray') in {func_signature}."
```
nfarabullini, line 440:
```suggestion
                f"Dependency {dep_key} in function {self._func.__name__}:  does not exist or has "
                f"wrong type ('expected np.ndarray') in {func_signature}."
```

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -0,0 +1,476 @@
...
+        grid_provider: GridProvider,
+    ) -> None:
+        self._validate_dependencies()
+        args = {k: factory.get(v).ndarray for k, v in self._dependencies.items()}
+        offsets = {k: grid_provider.grid.connectivities[v] for k, v in self._offsets.items()}
+        args.update(offsets)
+        args.update(self._params)
+        results = self._func(**args)
+        ## TODO: can the order of return values be checked?
+        results = (results,) if isinstance(results, xp.ndarray) else results
```
nfarabullini, line 427:
```suggestion
        ## TODO: can the order of return values be checked?
        results = (results,) if isinstance(self._func(**args), xp.ndarray) else results
```
Just as a personal preference

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -0,0 +1,476 @@
...
+
+
+    Args:
+        func: numpy function that computes the fields
+        domain: the compute domain used for the stencil computation
+        fields: Seq[str] names under which the results fo the function will be registered
+        deps: dict[str, str] input fields used for computing this stencil: the key is the variable name
+            used in the program and the value the name of the field it depends on.
+        params: scalar arguments for the function
+    """
```
nfarabullini, line 383: missing offsets description

model/common/src/icon4py/model/common/grid/geometry_program.py
```diff
@@ -0,0 +1,569 @@
...
+@gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
+def compute_edge_area(
+    owner_mask: fa.EdgeField[bool],
+    primal_edge_length: fa.EdgeField[ta.wpfloat],
+    dual_edge_length: fa.EdgeField[ta.wpfloat],
+    area: fa.EdgeField[ta.wpfloat],
+    horizontal_start: gtx.int32,
+    horizontal_end: gtx.int32,
+):
+    edge_area(
```
nfarabullini, line 530: in some field_operators call you have args and in some others keywargs. Can you put them all as keywargs?

nfarabullini (approved):
LGTM!

model/common/src/icon4py/model/common/grid/geometry_attributes.py
```diff
@@ -0,0 +1,302 @@
...
+        dims=(dims.EdgeDim,),
+        icon_var_name="",  # TODO
+        dtype=ta.wpfloat,
+    ),
+    EDGE_TANGENT_Y: dict(
+        standard_name=EDGE_TANGENT_Y,
+        long_name=EDGE_TANGENT_Y,
+        units="",  # TODO
+        dims=(dims.EdgeDim,),
+        icon_var_name="",  # TODO
```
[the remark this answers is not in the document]
halungge: I updated the data and removed most of them.

model/common/tests/grid_tests/utils.py
halungge: Good catch, I think I added it to try out something and it should be wanted to remove it again.

model/common/src/icon4py/model/common/grid/geometry_program.py
halungge: When calling field_ops in programs, I changed the input fields to positional args to make it constistent. The programs are only simple wrappers so the kwargs would not really help imho.

halungge:
cscs-ci run default

halungge:
launch jenkins spack

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
launch jenkins spack spackProject=C2SM/py_gt4py_add_diskcache_dependency

PR #590, opened 2024-11-07
Replace 2-elements sequences by `Pair` utility for double buffering

egparedes:
Update `icon4py_driver` to use the `Pair` utility classes and add a pair of `NamedTuple`s to document the multiple values returned by  the `initialize` function.


egparedes:
cscs-ci run default

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -260,13 +247,43 @@ def _do_dyn_substepping(
...
+    prognostic_swp: imc_utils.Swapping[prognostics.PrognosticState]
+    diagnostic: diagnostics.DiagnosticState
+
+
+class DriverParams(NamedTuple):
+    """
+    Parameters for the driver run.
+
+    Attributes:
+        divdamp_fac_o2: Second order divdamp factor.
```
halungge, line 281:
```suggestion
        divdamp_fac_o2: Second order divergence damping factor.
```
Variable name is taken directly from ICON... that is what it stands for.

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -10,20 +10,24 @@
...
 import pathlib
 import uuid
-from typing import Callable
+from typing import Callable, NamedTuple
 
 import click
 from devtools import Timer
 from gt4py.next import gtfn_cpu
 
+import icon4py.model.common.utils as imc_utils
```
halungge, line 19: why `imc_utils`? _i_con4py _m_odel  _c_ommon? Honestly, I don't like these kind of names. Why not `utils` or `common_utils`?

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -114,23 +105,17 @@ def n_time_steps(self):
...
     def _full_name(self, func: Callable):
         return ":".join((self.__class__.__name__, func.__name__))
 
     def time_integration(
         self,
         diffusion_diagnostic_state: diffusion_states.DiffusionDiagnosticState,
         solve_nonhydro_diagnostic_state: solve_nh_states.DiagnosticStateNonHydro,
-        # TODO (Chia Rui): expand the PrognosticState to include indices of now and next, now it is always assumed that now = 0, next = 1 at the beginning
-        prognostic_state_list: list[prognostics.PrognosticState],
+        prognostic_state_swp: imc_utils.Swapping[prognostics.PrognosticState],
```
halungge, line 115: can we simply call it `prognostic_states` in plural? `prognostic_states.swap()` is kind of nice....

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -260,13 +247,43 @@ def _do_dyn_substepping(
...
             if not self._is_last_substep(dyn_substep):
-                self._swap()
+                prognostic_state_swp.swap()
 
             self._is_first_step_in_simulation = False
 
         # TODO (Chia Rui): compute airmass for prognostic_state here
 
 
+class DriverStates(NamedTuple):
```
halungge, line 257: I don't know whether that makes sense, as I don't know wether all these state `structs`make sense. Potentially for every model component `X` you will have a different `XDiagnosticState` because there are so many diagnostic variables and each component uses a particular subset... 

But we might just introduce it for now and maybe later skip it. It makes the `initialize` function a bit easier to grasp. That said, I don't like the `initialize`at all. It does too many things at the time. I would split it apart.

model/driver/tests/driver_tests/test_timeloop.py
halungge, line 1: Does this run? wouldn't you have to propagate the usage of the `Swapping[PrognosticState]` down to SolveNonHydro?

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -198,36 +183,38 @@ def _integrate_one_time_step(
...
         if self.diffusion.config.apply_to_horizontal_wind:
             self.diffusion.run(
-                diffusion_diagnostic_state, prognostic_state_list[self._next], self.dtime_in_seconds
+                diffusion_diagnostic_state,
+                prognostic_state_swp.other,
+                self.dtime_in_seconds,
             )
 
-        self._swap()
+        prognostic_state_swp.swap()
```
havogt, line 208: Not 100% sure, but maybe better to introduce `Swappable` for the prognostic_state container, which has a `swap()` that calls `swap()` to all elements which have to be `Swapping`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro.py
```diff
@@ -895,12 +896,12 @@ def time_step(
...
             mask_prog_halo_c=self._metric_state_nonhydro.mask_prog_halo_c,
-            rho_now=prognostic_state_ls[nnow].rho,
-            theta_v_now=prognostic_state_ls[nnow].theta_v,
-            exner_new=prognostic_state_ls[nnew].exner,
-            exner_now=prognostic_state_ls[nnow].exner,
-            rho_new=prognostic_state_ls[nnew].rho,
-            theta_v_new=prognostic_state_ls[nnew].theta_v,
+            rho_now=prognostic_state_swp.current.rho,
+            theta_v_now=prognostic_state_swp.current.theta_v,
+            exner_new=prognostic_state_swp.other.exner,
```
havogt, line 901: Since `Swapping` is part of the application code, we could use `current` and `next`  or even `now` and `next` instead of `current` and `other`. 
`other` is perfect for a generic `Swapping` but less readable in the application code where this is used for timestepping.

model/driver/tests/driver_tests/test_timeloop.py
egparedes: Yes, that part was fully missing at that point but it is now included.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/nh_solve/solve_nonhydro.py
egparedes: Refactored `Swapping` to generic `Pair` which can be easily subclassed to set accessors with custom names.

egparedes:
cscs-ci run default

egparedes:
cscs-ci run default

halungge (approved):
LGTM, some very minor things. I still haven`t fully understood the swapping for the diagnostic tendencies. But I think we should try to fully understand that with @OngChia s help and fix or remove swapping/Pair where not needed.  That should become an extra PR.

halungge:
The CI failure is due to some corrupt data unrelated to your PR.

halungge:
cscs-ci run default

tools/src/icon4pytools/py2fgen/wrappers/dycore_wrapper.py
```diff
@@ -434,7 +432,7 @@ def solve_nh_run(
         rho=rho_new,
         exner=exner_new,
     )
-    prognostic_state_ls = [prognostic_state_nnow, prognostic_state_nnew]
+    prognostic_states = common_utils.TimeStepPair(prognostic_state_nnow, prognostic_state_nnew)
 
     # adjust for Fortran indexes
     nnow = nnow - 1
```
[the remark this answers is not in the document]
egparedes: Not, it doesn't, you're right.

egparedes:
cscs-ci run default

egparedes:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,59 @@ def _determine_local_domains(self):
...
-            self.ntl1 = nnow
-            self.ntl2 = nnew
-        else:
-            self.ntl1 = 0
-            self.ntl2 = 0
+    def update_time_levels_for_velocity_tendencies(
+        self,
+        diagnostic_state_nh: dycore_states.DiagnosticStateNonHydro,
+        at_first_substep: bool,
+        at_initial_substep: bool,
```
halungge, line 805: For me `at_first_substep` and `at_initial_substep` seem to be the same. Would it make sense to call it `at_initial_timestep` or even `at_initial_modelstep` and have it indicate only whether you are at the first model timestep? 

getting the first_supstep of the first timestep would then be `at_first_substep and at_initial_timestep`

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,59 @@ def _determine_local_domains(self):
...
+        """
+
+        assert (
+            self._config.itime_scheme == TimeSteppingScheme.MOST_EFFICIENT
+        ), f" only {TimeSteppingScheme.MOST_EFFICIENT} is supported. But {self._config.itime_scheme} is chosen."
+        if not at_initial_substep:
+            diagnostic_state_nh.ddt_w_adv_pc.swap()
+        if not at_first_substep:
+            diagnostic_state_nh.ddt_vn_apc_pc.swap()
+        return 1
```
halungge, line 843: If it always returns 1, do we then need the  return value at all Literal[0,1] or is it enough to swap or not to swap and always use `.next`?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,59 @@ def _determine_local_domains(self):
...
+        """
+        Set time levels of ddt_adv fields for call to velocity_tendencies.
+
+        When using `TimeSteppingScheme.MOST_EFFICIENT` (itime_scheme=4 in ICON Fortran),
+        `ddt_w_adv_pc.current` (advection term in vertical momentum equation in
+        predictor step) is not computed in the predictor step of each substep.
+        Instead, the advection term computed in the corrector step during the
+        previous substep is reused for efficiency (except, of course, in the
+        very first substep of the initial time step).
+        `ddt_v_apc.current` (advection term in horizontal momentum equation in
```
halungge, line 816:
```suggestion
        `ddt_vn_apc.current` (advection term in horizontal momentum equation in
```

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,59 @@ def _determine_local_domains(self):
...
+        predictor step) is only computed in the predictor step of the first
+        substep and the advection term in the corrector step during the previous
+        substep is reused for `ddt_v_apc.current` from the second substep onwards.
```
halungge, line 819:
```suggestion
        substep is reused for `ddt_vn_apc.current` from the second substep onwards.
```

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,59 @@ def _determine_local_domains(self):
...
+        Additionally, in this scheme the predictor and corrector outputs are kept
+        in separated elements of the pair (.current for the predictor and .next
```
halungge, line 821:
```suggestion
        in separate elements of the pair (.current for the predictor and .next
```

halungge: Just an idea: as the naming of the pair elements is flexible we could even have a 
```python
class PredictorCorrectorPair:(Pair[T]):
    predictor: T = Pair.first
    corrector: T = Pair.frozen_second
```
 and then use
`ddt_w_adv_pc.predictor` and `ddt_w_adv_pc.corrector`

tools/tests/py2fgen/test_dycore_wrapper.py
```diff
@@ -1602,16 +1604,16 @@ def test_granule_solve_nonhydro_multi_step_regional(
...
         )
         linit = False
         recompute = False
         clean_mflx = False
-        if not is_last_substep:
-            ntemp = nnow
-            nnow = nnew
-            nnew = ntemp
+
+        prognostic_states.swap()
```
halungge, line 1612: The wrapper function `dycore_wrapper.solve_nh_run` is the interface to Fortran. 
So the wrapper function needs all information such that it can determine internally wether the `Pair` has to be constructed  as `Pair(ddt_w_adv_ntl1, ddt_w_adv_ntl2)` or `Pair(ddt_w_adv_ntl2, ddt_w_adv_ntl1) `. I think that is possible with `linit` `ndyn_substep` and `idyn_substep` . 

Otherwise we would need to add the additional state that you are creating here outside of it (the Pairs for `ddt_vn_apc` and `ddt_w_adv` as a persistent state from one call to the next on the python side. (The wrapper does this with the grid, but I would try to avoid it for these advection tendencies.

Anyway it is good that is runs, that is good point to start with.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: It is a nice name! Correctly conveying why we need two buffers. I will use this name.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: Indeed. You are right. I will scrap the return value. And it is enough to just use `next` (now it becomes `corrector` following your suggestion on the name) in the corrector step in `velocity_advection`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: Thanks!

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: Thanks!

tools/tests/py2fgen/test_dycore_wrapper.py
OngChia: Thanks for the explanation. 
Because `solve_nonhydro.time_step` swaps the velocity tendencies internally, it needs to remember which `ddt_w_adv_ntl1` or `ddt_w_adv_ntl2` is the current or next. I agree that it is better not to add additional state just for advection tendencies.
Unless advection tendencies are also swapped outside of `time_step` like `PrognosticState`. But it will feel strange to me that we swap them outside since they are 'local' variables computed only by `solve_nonhydro`.
Possibly the best solution is keeping track of the 0, 1 indices for current and next.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: I like yoursuggestion to replace `at_initial_substep` or `l_init` in the original Fortran code by `at_initial_timestep`. I will do it.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -798,26 +798,36 @@ def _determine_local_domains(self):
         )
         self._end_vertex_halo = self._grid.end_index(vertex_domain(h_grid.Zone.HALO))
 
-    def set_timelevels(self, nnow, nnew):
-        #  Set time levels of ddt_adv fields for call to velocity_tendencies
+    def update_time_levels(
```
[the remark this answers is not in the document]
OngChia: It has been changed to `update_time_levels_for_velocity_tendencies`

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

OngChia:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
halungge: Yes, thank you. That was my suspicion from your flow write up on slack that the `l_recompute` is actually `at_first_substep`. But I did not want to open another bucket and thought we might fix it in another cleaning round...

OngChia:
cscs-ci run default

tools/tests/py2fgen/test_dycore_wrapper.py
halungge: I am still a bit worried about the eventual calls from FORTRAN. The wrapper interface takes a list of fields and internally constructs the pairs:
```
def solve_nh_run(...w_now, w_new, ddt_w_adv_ntl1, ddt_w_adv_ntl2)
       ...
       prog_pair = TimestepPar(w_now, w_new) 
       ddt_w_adv_pc = PredictorCorrectorPar(ddt_w_adv_ntl1, ddt_w_adv_ntl2)
       ...
```
where I use only `w` as replacement for the prognostic state.
Because the prognostic state the swap is done externally, every other call the array that is passed as `w_now` will be different, only every second call it will be the same. But for the advection tendencies? Aren't we missing some kind of logic in the _python wrapper_ not in the (real icon4py model code (greenline) and not in the python test, because that will not be executed in Fortran.

halungge (commented):
I think it starts looking good, some typos still to fix and I am still a bit unsure about the wrapper test, not for the test itself but for the eventual integration in Fortran of the wrapper and call from there.

tools/tests/py2fgen/test_dycore_wrapper.py
```diff
@@ -1596,22 +1577,25 @@ def test_granule_solve_nonhydro_multi_step_regional(
...
+
+        if itime_scheme == solve_nh.TimeSteppingScheme.MOST_EFFICIENT:
+            if not (at_initial_timestep and (i_substep - 1 == 0)):
+                ddt_w_adv.swap()
+                print(i_substep, "swap w")
+            if not (i_substep - 1 == 0):
+                ddt_vn_apc.swap()
+                print(i_substep, "swap vn")
+        else:
+            raise NotImplementedError(
```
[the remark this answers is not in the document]
OngChia: Thanks. I overlooked the beginning of the test that it sets `itime_scheme = solve_nh.TimeSteppingScheme.MOST_EFFICIENT`

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -919,37 +961,35 @@ def time_step(
...
-            f"running predictor step: dtime = {dtime}, init = {l_init}, recompute = {l_recompute} "
+            f"running predictor step: dtime = {dtime}, initial_timestep = {at_initial_timestep} at_first_substep = {at_first_substep}"
         )
-        if l_init or l_recompute:
-            if self._config.itime_scheme == TimeSteppingScheme.MOST_EFFICIENT and not l_init:
-                lvn_only = True  # Recompute only vn tendency
-            else:
-                lvn_only = False
+
+        if (at_initial_timestep and at_first_substep) or at_first_substep:
```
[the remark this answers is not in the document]
OngChia: Ah. You are right. `at_initial_substep` is redundant there. Thanks!

tools/tests/py2fgen/test_dycore_wrapper.py
OngChia: `solve_nh_run` will be pretty straightforward before this PR is merged because `ntl1` and `ntl2` in `ddt_w_adv_pc` are always set equal to `nnow` and `nnew`. 
To do it correctly when calling from Fortran, instead of doing what I saw in this PR https://github.com/C2SM/icon-exclaim/pull/282,
```
   ! Set time levels of ddt_adv fields for call to velocity_tendencies
      IF (itime_scheme >= 4) THEN ! Velocity advection averaging nnow and nnew tendencies
        ntl1 = nnow
        ntl2 = nnew
      ELSE                        ! Velocity advection is taken at nnew only
        ntl1 = 1
        ntl2 = 1
      ENDIF

      CALL solve_nh_run( 
                         w_now             = p_nh%prog(nnow)%w(:,:,1), &
                         w_new             = p_nh%prog(nnew)%w(:,:,1), &
                         ddt_vn_apc_ntl1   = p_nh%diag%ddt_vn_apc_pc(:,:,1,ntl1), &
                         ddt_vn_apc_ntl2   = p_nh%diag%ddt_vn_apc_pc(:,:,1,ntl2), &
                         ddt_w_adv_ntl1    = p_nh%diag%ddt_w_adv_pc(:,:,1,ntl1), &
                         ddt_w_adv_ntl2    = p_nh%diag%ddt_w_adv_pc(:,:,1,ntl2), &
```
we just need to pass in as below because `ntl1` and `ntl2` are set internally in `SolveNonhydro.time_step`.
```
   ! Set time levels of ddt_adv fields for call to velocity_tendencies
      IF (itime_scheme >= 4) THEN ! Velocity advection averaging nnow and nnew tendencies
        ntl1 = nnow
        ntl2 = nnew
      ELSE                        ! Velocity advection is taken at nnew only
        ntl1 = 1
        ntl2 = 1
      ENDIF

      CALL solve_nh_run( 
                         w_now             = p_nh%prog(nnow)%w(:,:,1), &
                         w_new             = p_nh%prog(nnew)%w(:,:,1), &
                         ddt_vn_apc_ntl1   = p_nh%diag%ddt_vn_apc_pc(:,:,1,1), &
                         ddt_vn_apc_ntl2   = p_nh%diag%ddt_vn_apc_pc(:,:,1,2), &
                         ddt_w_adv_ntl1    = p_nh%diag%ddt_w_adv_pc(:,:,1,1), &
                         ddt_w_adv_ntl2    = p_nh%diag%ddt_w_adv_pc(:,:,1,2), &
```

After we merge this branch using `PredictorCorrectorPair` class, we kind of decouple `ntl1` and `ntl2` from `nnow` and `nnew` and swap when we really need to (I like this because it reflects the time stepping scheme correctly). We now need to know when `ddt_vn_apc_pc` and `ddt_w_adv_pc` are swapped as done in this test
```
        if not (at_initial_timestep and (idyn_timestep == 1)) then
            ddt_w_adv.swap()
        if not (i_substep - 1 == 0):
            ddt_vn_apc.swap()
```
@samkellerhals should also notice this issue.

OngChia:
cscs-ci run default

tools/tests/py2fgen/test_dycore_wrapper.py
samkellerhals: @OngChia I think what you showed in your last example would make sense as the integration code in Fortran.

OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

PR #606, opened 2024-11-26
convencience functions for array ns import and field transfer

halungge:
adds convencience functions for 
- import of array_ns depending on backend
- transfer field to a given backend

### DIFF
--- a/model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
+++ b/model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
@@ -5,13 +5,49 @@
 #
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
+import logging as log
 from typing import Optional
 
+import gt4py._core.definitions as gt_core_defs
 import gt4py.next as gtx
 from gt4py.next import backend
 
-from icon4py.model.common import type_alias as ta
-from icon4py.model.common.settings import xp
+from icon4py.model.common import dimension, type_alias as ta
+
+
+def is_cupy_device(backend: backend.Backend) -> bool:
+    cuda_device_types = (
+        gt_core_defs.DeviceType.CUDA,
+        gt_core_defs.DeviceType.CUDA_MANAGED,
+        gt_core_defs.DeviceType.ROCM,
+    )
+    if backend is not None:
+        return backend.allocator.__gt_device_type__ in cuda_device_types
+    else:
+        return False
+
+
+def array_ns(try_cupy: bool):
+    if try_cupy:
+        try:
+            import cupy as cp
+
+            return cp
+        except ImportError:
+            log.warn("No cupy installed, falling back to numpy for array_ns")
+    import numpy as np
+
+    return np
+
+
+def import_array_ns(backend: backend.Backend):
+    """Import cupy or numpy depending on a chosen GT4Py backend DevicType."""
+    return array_ns(is_cupy_device(backend))
+
+
+def as_field(field: gtx.Field, backend: backend.Backend) -> gtx.Field:
+    """Convenience function to transfer an existing Field to a given backend."""
+    return gtx.as_field(field.domain, field.ndarray, allocator=backend)
 
 
 def allocate_zero_field(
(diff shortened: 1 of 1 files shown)


halungge:
cscs-ci run default

halungge:
launch jenkins spack

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
```diff
@@ -5,13 +5,49 @@
...
+def is_cupy_device(backend: backend.Backend) -> bool:
+    cuda_device_types = (
+        gt_core_defs.DeviceType.CUDA,
+        gt_core_defs.DeviceType.CUDA_MANAGED,
+        gt_core_defs.DeviceType.ROCM,
+    )
+    if backend is not None:
+        return backend.allocator.__gt_device_type__ in cuda_device_types
+    else:
+        return False
```
nfarabullini, line 27:
```suggestion
    return False
```

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
```diff
@@ -20,12 +56,15 @@ def allocate_zero_field(
...
-    if is_halfdim:
-        assert len(shapex) == 2
-        shapex = (shapex[0], shapex[1] + 1)
-    return gtx.as_field(dims, xp.zeros(shapex, dtype=dtype), allocator=backend)
+) -> gtx.Field:
+    def size(dim: gtx.Dimension, is_half_dim: bool) -> int:
+        if dim == dimension.KDim and is_half_dim:
+            return grid.size[dim] + 1
+        else:
+            return grid.size[dim]
```
nfarabullini, line 64:
```suggestion
        return grid.size[dim]
```

nfarabullini: do we need this function inside `allocate_zero_field` or can we move it out?

nfarabullini (approved):

OngChia (approved):

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
```diff
@@ -5,13 +5,49 @@
...
 import gt4py.next as gtx
 from gt4py.next import backend
 
-from icon4py.model.common import type_alias as ta
-from icon4py.model.common.settings import xp
+from icon4py.model.common import dimension, type_alias as ta
+
+
+def is_cupy_device(backend: backend.Backend) -> bool:
+    cuda_device_types = (
```
OngChia, line 19: Can we pull this tuple outside of this function and declare it as `_cuda_device_types`?

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
```diff
@@ -5,13 +5,49 @@
...
+        gt_core_defs.DeviceType.CUDA,
+        gt_core_defs.DeviceType.CUDA_MANAGED,
+        gt_core_defs.DeviceType.ROCM,
+    )
+    if backend is not None:
+        return backend.allocator.__gt_device_type__ in cuda_device_types
```
OngChia, line 25: I am simply asking a question here. 
There is this function `is_field_allocator_for` in gt4py.next. I wonder if it is better to use that function  to check if it is the gpu backend. If not, why. For example, 
`if allocators.is_field_allocator_for(backend.allocator, core_defs.DeviceType.CUDA)`

halungge: I remember that I had a look at this one, but then choose not to use it because we would need to call it for each GPU device type. 
```
if allocators.is_field_allocator_for(backend.allocator, core_defs.DeviceTyp.CUDA) or allocators.is_field_allocator_for(backend.allocator, core_defs.DeviceTyp.ROCM) or allocators.is_field_allocator_for(backend.allocator, core_defs.DeviceTyp.CUDA_MANAGED)
```
that looked very cumbersome to me.

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
halungge: yes.

model/common/src/icon4py/model/common/utils/gt4py_field_allocation.py
halungge: not necessarily. I thought there is no other usage for it elsewhere because the `is_half_dim` argument is pretty special to this case. But that might be wrong. It should be removed anyway once the `KDim` vs `KHalfDim` issue is resolved. 

Currently there are several solutions of the issue accross the code base.

halungge: I extracted it and used it in the other allocator as well.

PR #609, opened 2024-11-27
Grid manager with backend

halungge:
use GT4Py backend in `GridManager`:
- fields (coordinates geometry) are allocated using the backend
-  the existing `on_gpu` flag for the `IconGrid` construction is determined from the backend, for backwards compatibilty  
-  GridManager internally explicitly uses numpy for reading

**further changes and restrictions**
- test_io.py: uses a backend fixed to field view embedded as a sample CPU backend. 
- the `IconGrid` construction does not fully support the CUPY vs NUMPY import. It remains on CPU everywhere.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/grid_manager.py
+++ b/model/common/src/icon4py/model/common/grid/grid_manager.py
@@ -529,14 +529,14 @@ class GridManager:
         Refinement control contains the classification of each entry in a field to predefined horizontal grid zones as for example the distance to the boundaries,
         see [refinement.py](refinement.py)
         """
-        np = common_utils.gt4py_field_allocation.import_array_ns(backend)
+        xp = common_utils.gt4py_field_allocation.import_array_ns(backend)
         refinement_control_names = {
             dims.CellDim: GridRefinementName.CONTROL_CELLS,
             dims.EdgeDim: GridRefinementName.CONTROL_EDGES,
             dims.VertexDim: GridRefinementName.CONTROL_VERTICES,
         }
         refinement_control_fields = {
-            dim: np.asarray(self._reader.int_variable(name, decomposition_info, transpose=False))
+            dim: xp.asarray(self._reader.int_variable(name, decomposition_info, transpose=False))
             for dim, name in refinement_control_names.items()
         }
         return refinement_control_fields


halungge:
cscs-ci run default

halungge:
cscs-ci run default

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -11,13 +11,21 @@
...
     definitions as decomposition,
 )
 from icon4py.model.common.grid import base, icon, vertical as v_grid
-from icon4py.model.common.settings import xp
+
+
+try:
+    import cupy as xp
+except ImportError:
+    import numpy as xp
```
nfarabullini, line 28: I don't understand why we need this if we're using numpy afterwards

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -496,27 +519,30 @@ def _read_start_end_indices(
...
+        backend: gtx_backend.Backend,
+        decomposition_info: Optional[decomposition.DecompositionInfo] = None,
+    ) -> tuple[dict[dims.Dimension : NDArray]]:
         """
         Reads the refinement control fields from the grid file.
 
         Refinement control contains the classification of each entry in a field to predefined horizontal grid zones as for example the distance to the boundaries,
         see [refinement.py](refinement.py)
         """
+        np = common_utils.gt4py_field_allocation.import_array_ns(backend)
```
nfarabullini, line 532: since we use `np` for numpy direct import, would t be better to rename this to something else?

model/common/tests/grid_tests/utils.py
```diff
@@ -97,15 +99,19 @@ def valid_boundary_zones_for_dim(dim: dims.Dimension):
     yield from _domain(dim, zones)
 
 
-@functools.cache
-def run_grid_manager(experiment_name: str, num_levels=65, transformation=None) -> gm.GridManager:
+def run_grid_manager(
+    experiment_name: str,
+    backend: gtx_backend.Backend = gtx.gtfn_cpu,
+    num_levels=65,
```
nfarabullini, line 105: is it ok to hard code this?

model/common/tests/io_tests/test_io.py
```diff
@@ -32,15 +32,15 @@
 from icon4py.model.common.test_utils import datatest_utils, grid_utils, helpers
 
 
+# fixing backend to fieldview embedded here.
```
nfarabullini, line 35:
```suggestion
# setting backend to fieldview embedded here.
```

nfarabullini (approved):

model/common/tests/grid_tests/utils.py
halungge: I set it as default. It is the same default as before.

model/common/src/icon4py/model/common/grid/grid_manager.py
halungge: I added the `xp` only for the typing. (There is this `NDArray` TypeAlias)  Inside the grid construction I kept this flag `on_gpu` as it was before for compatibility, that seems to work, but I have not yet bent my head around how it works. Rather not touched the grid construction. It think the solution there will eventually be that we read (on CPU) and then copy to device probably right before we return the grid as for the other fields. 

For this we will need to do the import based on the backend I think.

model/common/src/icon4py/model/common/grid/grid_manager.py
halungge: Right, should be `xp` really.

halungge:
cscs-ci run  default

PR #616, opened 2024-12-02
[Py2f]: Cache Fortran pointers in dictionary

samkellerhals:
### Description

Unpacks pointers from Fortran only once and then caches them in a dictionary. This reduces overhead from repeatedly unpacking pointers at runtime, whilst the improvement is relatively small for diffusion ~5%, it may become more significant in cases where there are a lot of arrays passed to the embedded Python function.


samkellerhals:
cscs-ci run default

havogt:
outdated

PR #624, opened 2024-12-11
Freeze programs in diffusion

samkellerhals:


samkellerhals:
cscs-ci run default

samkellerhals:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -467,6 +463,55 @@ def __init__(
             self._grid.offset_providers
         )
 
+        self.scale_k = self.scale_k.with_connectivities(self.compile_time_connectivities).freeze()
```
halungge, line 466: I like that the all the massaging around with the stencil function is done in the init, and those `with_connectivities`are no longer all over the place. 

Can you check again with @tehrengruber whether that would still work if we were to call `field_operator`s here instead of `program`s. We shortly mentioned that in the standup yesterday but I dont' recall the answer.

PR #634, opened 2025-01-08
Enhance README file with more information about uv

egparedes:
Add further details about how to use `uv` properly in the development workflow.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ ICON4Py uses the `uv` tool to manage the development workflow. `uv` is a versati
 $ curl -LsSf https://astral.sh/uv/install.sh | sh 
 ```
 
-Finally, make sure **_boost >= 1.85.0_** is installed in your system, which is required by `gt4py` to compile generated high-performance code. 
+Finally, make sure **_boost >= 1.85.0_** is installed in your system, which is required by `gt4py` to compile generated C++ code. 
 
 ### ICON4Py Development Environment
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,31 +14,43 @@ The repository is organized as a _monorepo_, where various ICON model components
...
 
-### ICON4Py development environment
+Finally, make sure **_boost >= 1.85.0_** is installed in your system, which is required by `gt4py` to compile generated C++ code. 
+
+### ICON4Py Development Environment
+
+Once `uv` is installed in your system, it is enough to clone this repository and let `uv` handling the installation of the development environment. 
 
-Once `uv` is installed in your system, it is enough to clone this repository and let `uv` handling the installation of the development environment.
+**Important**: the `uv sync` command should always be executed from the **root folder** of the repository, to make sure it install all the workspace dependencies and not only the dependencies of a subproject.
```
nfarabullini, line 41:
```suggestion
**Important**: the `uv sync` command should always be executed from the **root folder** of the repository, to make sure it installs all the workspace dependencies and not only the dependencies of a subproject. 
```

nfarabullini (approved):

muellch (approved):
Looks good!

This is probably not the correct place to start the discussion, but we state explicitly that we need boost for icon4py to work. There is other implicit requirements such as a `gcc` new enough (at least if certain compiled packages need to work). Should we ever write these down? I do not know the conventions in the python world for this stuff.

PR #639, opened 2025-01-14
Fix interpolation field geofac grg

halungge:
(FIX) fix computation of geofac_grg: 

  -  fix central value for higher absolute tolerance
  - use tuples

- additional change: improve the w_1 property in prognostic state.


nfarabullini (approved):

halungge:
cscs-ci run default

PR #640, opened 2025-01-15
Run icon4py and datatests on gpu

OngChia:
Enable driver runs and datatests on gpu. Only single-node run is tested in this PR.

Major changes related to driver runs on gpu:
- A backend option is added into one of the click options (`icon4py_driver_backend` or shortform `-b`). `xp` is decided explicitly by calling the function in `data_allocation.py` that returns np if cpu backends are used or cp if gpu backends are used.
Computation in `vertical.py` is changed to numpy and a backend argument is added in `get_vct_a_and_vct_b` function to create `vct_a` and `vct_b` fields with the correct backend.

- A new field_operator `_interpolate_to_half_levels_wp` is used in `compute_virtual_potential_temperatures_and_pressure_gradient.py` to avoid CUDA illegal memory access for an unknown reason.

- Because backend is added as an instance variable of `IconSavepoint`, fields that are read from the serialized data are automatically allocated on the device if gpu backend is used.

Major changes related to datatests on gpu:
- Add backend as an instance variable of `IconSavepoint` to create gt4py fields on the device if running on GPU.

- Specify backend whenever allocation of gt4py fields is performed in all datatests.

- Connectivities  are allocated with `xp.asarray` when constructing IconGrid in `grid_manager.py` and `serialbox.py`.

Note:
- test_advection.py fails on GPU. Further debugging is required.


OngChia:
cscs-ci run default

OngChia:
launch jenkins spack

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

model/common/tests/states_test/test_factory.py
```diff
@@ -106,7 +108,9 @@ def height_coordinate_source(metrics_savepoint, grid_savepoint, backend):
 
 
 @pytest.mark.datatest
-def test_field_operator_provider(cell_coordinate_source):
+def test_field_operator_provider(cell_coordinate_source, backend):
```
halungge, line 111: I think you could even remove the `backend` fixture here in most of the tests the backend is hard coded any way.

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -40,6 +50,19 @@ class Icon4pyRunConfig:
 
     restart_mode: bool = False
 
+    backend_name: str = data_alloc.DEFAULT_BACKEND
+
+    def __post_init__(self):
+        if self.backend_name not in data_alloc.BACKENDS:
```
halungge, line 56: I would keep this backend list somewhere in the driver package instead of the common.

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -40,6 +50,19 @@ class Icon4pyRunConfig:
 
     restart_mode: bool = False
 
+    backend_name: str = data_alloc.DEFAULT_BACKEND
```
halungge, line 53: why not make the backend a mandatory parameter. I think we should force the user to set it.

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -435,13 +447,13 @@ def initialize(
 )
 @click.option(
     "--serialization_type",
-    default="serialbox",
+    default=driver_init.SerializationType.SB.value,
```
halungge, line 450: For a different PR, but this option does not make sense (I know I introduced it months ago)

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -464,8 +476,29 @@ def initialize(
...
 )
+@click.option(
+    "--enable_output",
+    is_flag=True,
+    help="Enable all debugging messages. Otherwise, only critical error messages are printed.",
+)
+@click.option(
+    "--icon4py_driver_backend",
+    "-b",
+    default=data_alloc.DEFAULT_BACKEND,
```
halungge, line 487: Does it really make sense to set a default value for the backend?

model/driver/src/icon4py/model/driver/test_cases/utils.py
```diff
@@ -6,29 +6,36 @@
...
-    wgtfac_c: np.ndarray,
-    ddqz_z_half: np.ndarray,
-    exner_ref_mc: np.ndarray,
-    d_exner_dz_ref_ic: np.ndarray,
-    theta_ref_mc: np.ndarray,
-    theta_ref_ic: np.ndarray,
-    rho: np.ndarray,
-    exner: np.ndarray,
-    theta_v: np.ndarray,
+def hydrostatic_adjustment_ndarray(
```
halungge, line 24: you might as well consider the pattern in the interpolation fields (`interpolation_fields.py`) for this:
```
def hydrostatic_adjustment(
     wgtfac_c: data_alloc.NDArray,
      dqz_z_half: data_alloc.NDArray,
     ... ,
array_ns: ModuleType = np):
    ...
      array_ns.sqrt(...)
     ...
    return rho, exner, theta_v
```
then then bind the array_ns via functools.partial where you call it. The advantage being that where you define the utility function (aka here) you don't need to splatter the rather heavy weight backend around.

model/driver/src/icon4py/model/driver/test_cases/utils.py
```diff
@@ -97,17 +104,18 @@ def hydrostatic_adjustment_constant_thetav_numpy(
...
-    edge_lat: np.ndarray,
-    edge_lon: np.ndarray,
-    primal_normal_x: np.ndarray,
-    eta_v_e: np.ndarray,
-):
+    edge_lat: data_alloc.NDArray,
+    edge_lon: data_alloc.NDArray,
+    primal_normal_x: data_alloc.NDArray,
+    eta_v_e: data_alloc.NDArray,
+    backend: gt4py_backend.Backend,
```
halungge, line 117: Same here I would rather use the array_ns: ModuleType argument pattern.

model/common/tests/interpolation_tests/test_interpolation_factory.py
halungge, line 1: The factory tests should run on any backend, even though some fields need to be computed on a specific one. But for me it is fine to keept it like this with the skip and have a look at it later and separately.

halungge (changes requested):
A couple of things: 
- I would move the backend list from common/utils/data_allocation.py to the driver.  As that is where we really use it.
- for simple functions using array API (hence any functions from numpy or cupy namespace) consider using a patther like in interpolation_fields.py where the array namespace is passed as an argument of type `ModuleType` and only bound where it is used instead of passing the backend and calling the import function inside the function.

model/common/src/icon4py/model/common/grid/utils.py
```diff
@@ -6,17 +6,20 @@
...
 
 def neighbortable_offset_provider_for_1d_sparse_fields(
+    on_gpu: bool,
     old_shape: tuple[int, int],
     origin_axis: Dimension,
     neighbor_axis: Dimension,
     has_skip_values: bool,
 ):
-    table = np.arange(old_shape[0] * old_shape[1], dtype=gtx.int32).reshape(old_shape)
+    xp = data_alloc.array_ns(on_gpu)
```
nfarabullini, line 21: we do this outside of the function: https://github.com/C2SM/icon4py/blob/64d5763c5763ed4b14147ebe7d3dbe4d3d925896/model/common/src/icon4py/model/common/interpolation/interpolation_fields.py#L155

nfarabullini (commented):
my 2 preliminary cents

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -112,41 +113,50 @@ class VerticalGrid:
...
         )
         object.__setattr__(
             self,
             "_vct_b",
-            vct_b,
+            gtx.as_field((dims.KDim,), vct_b, allocator=self.backend)
+            if vct_b is not None
+            else None,
         )
         vct_a_array = self._vct_a.ndarray
```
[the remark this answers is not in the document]
OngChia: I have reverted back and change  to ` vct_a_array = self._vct_a.asnumpy()` following your suggestion.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -247,38 +257,45 @@ def size(self, dim: gtx.Dimension) -> int:
 
     @classmethod
     def _determine_start_level_of_moist_physics(
-        cls, vct_a: data_alloc.NDArray, top_moist_threshold: float, nshift_total: int = 0
+        cls,
+        vct_a: data_alloc.NDArrayInterface,
```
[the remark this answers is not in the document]
OngChia: I use `vct_a: np.ndarray` instead because we have decided to compute on CPU.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -247,38 +257,45 @@ def size(self, dim: gtx.Dimension) -> int:
...
 
     @classmethod
     def _determine_damping_height_index(
-        cls, vct_a: data_alloc.NDArray, damping_height: float
+        cls, vct_a: data_alloc.NDArray, damping_height: float, backend: gt4py_backend.Backend
     ) -> gtx.int32:
         assert damping_height >= 0.0, "Damping height must be positive."
+        xp = data_alloc.import_array_ns(backend)
         return (
             0
```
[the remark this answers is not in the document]
OngChia: Thanks! I totally agree.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/interpolate_to_half_levels_wp.py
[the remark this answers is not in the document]
OngChia: Ah, True.

model/atmosphere/dycore/tests/dycore_tests/mpi_tests/test_parallel_solve_nonhydro.py
```diff
@@ -89,6 +89,7 @@ def test_run_solve_nonhydro_single_step(
         config=vertical_config,
         vct_a=grid_savepoint.vct_a(),
         vct_b=grid_savepoint.vct_b(),
```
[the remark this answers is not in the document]
OngChia: Yeah. I have reverted back.

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -112,41 +113,50 @@ class VerticalGrid:
...
-            vct_a,
+            gtx.as_field((dims.KDim,), vct_a, allocator=self.backend),
         )
         object.__setattr__(
             self,
             "_vct_b",
-            vct_b,
+            gtx.as_field((dims.KDim,), vct_b, allocator=self.backend)
+            if vct_b is not None
+            else None,
```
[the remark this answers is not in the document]
OngChia: Ah, I never thought about that. `vct_b` is used in initialization of vertical wind from real data and interpolated to icon grid as far as I know. The default value should be a decreasing function with height. It is up to users to decide. I agree that it should not be None, similar to `vct_a` which should also not be None. Maybe we need to put a check here if vct_a is an increasing function and vct_b is an decreasing function. But, our current experiments (warm bubble, mountain wave, JW) do not use vct_b. We can leave it later.

model/common/src/icon4py/model/common/field_type_aliases.py
```diff
@@ -24,3 +30,6 @@
 CellKField: TypeAlias = Field[Dims[dims.CellDim, dims.KDim], T]
 EdgeKField: TypeAlias = Field[Dims[dims.EdgeDim, dims.KDim], T]
 VertexKField: TypeAlias = Field[Dims[dims.VertexDim, dims.KDim], T]
+
+NDArrayInterface: TypeAlias = Union[np.ndarray, xp.ndarray, gtx.Field]
```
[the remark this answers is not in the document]
OngChia: I will delete it.

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -221,7 +221,7 @@ def start_index(self, domain: h_grid.Domain) -> gtx.int32:
             # special treatment because this value is not set properly in the underlying data.
             return gtx.int32(0)
         # ndarray.item() does not respect the dtype of the array, returns a copy of the value _as the default python type_
-        return gtx.int32(self._start_indices[domain.dim][domain()])
+        return gtx.int32(self._start_indices[domain.dim][domain()].item())
```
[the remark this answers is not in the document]
OngChia: I confirm that we do not need `.item()` even when we run on GPU. So, `start_index` and `end_index` can safely sit on CPU.

model/common/src/icon4py/model/common/grid/utils.py
OngChia: Thanks! 
Okay, I will pass in array_ns.

OngChia:
cscs-ci run default

model/common/src/icon4py/model/common/grid/icon.py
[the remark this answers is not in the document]
OngChia: So, I have removed item() in this PR.

OngChia:
cscs-ci run default

model/common/tests/interpolation_tests/test_interpolation_factory.py
OngChia: Okay. Thanks.

model/driver/src/icon4py/model/driver/test_cases/utils.py
OngChia: Thanks. 
I will follow the pattern there to pass in xp instead of backend.

model/driver/src/icon4py/model/driver/icon4py_driver.py
OngChia: I agree that user should be forced to choose a backend instead of giving a default value.
I deleted the default value.

model/driver/src/icon4py/model/driver/icon4py_driver.py
OngChia: I thought of "serialization_type"  changing to something else to include generating those interpolation and metric fields by the factory in the future. But I am not sure how it would go.
Okay, I will leave it as it is now in this PR.

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -576,7 +576,11 @@ def _compute(
     ) -> None:
         self._validate_dependencies()
         args = {k: factory.get(v).ndarray for k, v in self._dependencies.items()}
-        offsets = {k: grid_provider.grid.connectivities[v] for k, v in self._connectivities.items()}
+        xp = data_alloc.import_array_ns(backend)
```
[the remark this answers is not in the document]
OngChia: You are right. It is not necessary if connectivities are already on the correct device when running on GPU. I have removed it now.

model/common/tests/states_test/test_factory.py
OngChia: Ah, I see. I have removed backend here.

model/driver/src/icon4py/model/driver/icon4py_configuration.py
OngChia: Following our discussion, I have made a new python file `model_backends.py` and put the backend list there for pytest and driver to use.

model/driver/src/icon4py/model/driver/icon4py_configuration.py
OngChia: Great idea. Thanks. I add required=True to force user to pick a backend.

model/driver/src/icon4py/model/driver/test_cases/utils.py
OngChia: .

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

model/common/src/icon4py/model/common/utils/data_allocation.py
```diff
@@ -74,26 +79,32 @@ def import_array_ns(backend: backend.Backend):
     return array_ns(is_cupy_device(backend))
 
 
-def as_field(field: gtx.Field, backend: backend.Backend) -> gtx.Field:
+def as_field(field: gtx.Field, backend: backend.Backend = None) -> gtx.Field:
```
halungge, line 82:
```suggestion
def as_field(field: gtx.Field, backend: Optional[backend.Backend] = None) -> gtx.Field:
```
to allow `None`

model/common/src/icon4py/model/common/utils/data_allocation.py
```diff
@@ -74,26 +79,32 @@ def import_array_ns(backend: backend.Backend):
...
     """Convenience function to transfer an existing Field to a given backend."""
     return gtx.as_field(field.domain, field.ndarray, allocator=backend)
 
 
-def as_1D_sparse_field(field: gtx.Field, target_dim: gtx.Dimension) -> gtx.Field:
+def as_1D_sparse_field(
+    field: gtx.Field, target_dim: gtx.Dimension, backend: backend.Backend = None
```
halungge, line 88:
```suggestion
    field: gtx.Field, target_dim: gtx.Dimension, backend: Optional[backend.Backend] = None
```
to allow `None`

model/common/tests/metric_tests/test_compute_diffusion_metrics.py
```diff
@@ -36,24 +36,30 @@ def test_compute_diffusion_metrics(
     if experiment == dt_utils.GLOBAL_EXPERIMENT:
         pytest.skip(f"Fields not computed for {experiment}")
 
-    mask_hdiff = data_alloc.zero_field(icon_grid, dims.CellDim, dims.KDim, dtype=bool).asnumpy()
+    mask_hdiff = data_alloc.zero_field(
```
halungge, line 39: The combination of passing a parametrizable backend and then directly doing `.asnumpy()` does not make much sense to me. You could then also right away fix the `backend=None`

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -40,6 +41,19 @@ class Icon4pyRunConfig:
...
+    backend_name: str = model_backends.DEFAULT_BACKEND
+
+    def __post_init__(self):
+        if self.backend_name not in model_backends.BACKENDS:
+            raise ValueError(
+                f"Invalid driver backend: {self.backend_name}. \n"
+                f"Available backends are {', '.join([f'{k}' for k in model_backends.BACKENDS.keys()])}"
+            )
+
+    @property
```
halungge, line 53: maybe make this a `cached_property`?

halungge (changes requested):
LGTM,
two things: 
- Can you fix the typing for the backend to allow for `None` (embedded) type by making it an `Optional[Backend]` ?
- You should be able to run the datatest with gtfn_gpu now on the CI (once the CI runs again) Can you add this option in  `default.yml`: 
```
.test_model_datatests:
  stage: test
  script:
    - nox -s "test_model_datatest-3.10(subpackage='$COMPONENT')" -- --backend=$BACKEND 
  parallel:
    matrix:
    - COMPONENT: [atmosphere/advection, atmosphere/diffusion, atmosphere/dycore, atmosphere/subgrid_scale_physics/microphysics, common, driver]
      BACKEND: [gtfn_cpu, gtfn_gpu]
```

model/atmosphere/advection/tests/advection_tests/utils.py
```diff
@@ -37,10 +38,12 @@ def construct_config(
 
 
 def construct_interpolation_state(
-    savepoint: sb.InterpolationSavepoint,
+    savepoint: sb.InterpolationSavepoint, backend: gt4py_backend.Backend
```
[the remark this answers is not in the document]
OngChia: Okay. I added Optional to backend annotation except those in the `initialization_utils.py` because I think that embedded is not for users to choose to run the model. It is only for development or maybe some stencils in special cases.

model/common/src/icon4py/model/common/utils/data_allocation.py
OngChia: Okay. Done.

model/common/src/icon4py/model/common/utils/data_allocation.py
OngChia: Done.

model/driver/src/icon4py/model/driver/icon4py_configuration.py
OngChia: Ah. Yes.

model/common/tests/metric_tests/test_compute_diffusion_metrics.py
OngChia: If this is run on gtfn_cpu backend, I think it is still okay to call those gt4py stencils with the backend and only `compute_diffusion_metrics` be run with numpy. If running on GPu backend, then I agree that it does not make sense. I would suggest skipping this test for now (I have tried changing np to xp in `compute_diffusion_metrics` and it becomes super slow)

OngChia:
cscs-ci run default

halungge (approved):
LGTM

OngChia:
cscs-ci run default

havogt:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

PR #650, opened 2025-01-28
Add automatic grid download to grid_geometry

havogt:
Not sure if that is the correct fix, but I tried to run some tests locally and didn't get the grid data automatically.


halungge (approved):
makes sense to me. Thanks for catching..

havogt:
cscs-ci run default

PR #651, opened 2025-01-29
move scidoc to docstrings of individual stencils 

halungge:
Move the #scidoc strings from `SolveNonHydro` and  `VelocityAdvection` to the doc string of the individual stencil field operators.

- Adapted some of the outputs and input namings where they changed from the stencil call to the stencil implementation
- Split some of the #scidocs were they collected previousely fused stencils that most likely will not longer be used.


halungge:
cscs-ci run default

PR #636, opened 2025-01-09
Introduced customized markers for backend errors

nfarabullini:
setting `embedded` as default backend and related edits

PR #653, opened 2025-01-31
Re-enable diffusion tests when orchestration is False

edopao:
Follow-up on #636. Some tests are skipped my mistake.


edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run default

model/atmosphere/diffusion/tests/diffusion_tests/utils.py
```diff
@@ -47,7 +47,7 @@ def verify_diffusion_fields(
         assert helpers.dallclose(val_dwdx, ref_dwdx, atol=1e-18)
         assert helpers.dallclose(val_dwdy, ref_dwdy, atol=1e-18)
 
-    assert helpers.dallclose(val_vn, ref_vn, rtol=1e-5)
+    assert helpers.dallclose(val_vn, ref_vn, rtol=1e-4)
```
halungge, line 50: can you set an (additional) absolute tolerance here? It used to be one before [this PR](https://github.com/C2SM/icon4py/pull/586) where I changed it to this relative tolerance only. I checked the values on that field the range from something like 60  to -60 (velocities) so for the large values a `rtol` of 1e-4 does look suspicious to me, but once you get values close to 0 it is good to have one. I played around a bit, would this setting workfor all backends: 
`helpers.dallclose(val_vn, ref_vn, atol=1.0e-9, rtol=1.0e-10)`

edopao:
cscs-ci run default

model/atmosphere/diffusion/tests/diffusion_tests/utils.py
edopao: Thank you, I always forget the physical meaning of these numbers. I had to use `atol=1.0e-9, rtol=1.0e-9` because with `atol=1.0e-9` it still failed.

edopao:
cscs-ci run default

halungge (approved):
LGTM

PR #659, opened 2025-02-06
CI: cleanup the test runner config

edopao:
Small cleanup, since the test runner config `.container-runner-santis-gh200` is now available.


edopao:
cscs-ci run default

halungge (approved):

PR #663, opened 2025-02-10
Combined vel stencils verification 

yiluchen1066:


model/atmosphere/dycore/tests/dycore_tests/test_velocity_advection.py
```diff
@@ -450,3 +457,362 @@ def test_velocity_corrector_step(
...
+    z_kin_hor_e = savepoint_velocity_1_7_init.z_kin_hor_e()
+    z_v_grad_w = savepoint_velocity_1_7_init.z_v_grad_w()
+    k = data_alloc.allocate_indices(dim=dims.KDim, grid=icon_grid, is_halfdim=True, backend=backend)
+
+    lvn_only = savepoint_velocity_1_7_init.lvn_only()
+    edge = data_alloc.allocate_indices(dim=dims.EdgeDim, grid=icon_grid, backend=backend)
+    lateral_boundary_7 = icon_grid.start_index(edge_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_7))
+    halo_1 = icon_grid.end_index(edge_domain(h_grid.Zone.HALO))
+
+    horizontal_start = icon_grid.start_index(edge_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_5))
```
edopao, line 511: Just a question. I see that `horizontal_start` and `horizontal_end` differ from the stencil test. Should we back propagate these new values to the stencil test? Also the stencil tests use the ICON grids.

Stencil test:
https://github.com/C2SM/icon4py/blob/0b031505177d58a27032378e4c30b550318d465d/model/atmosphere/dycore/tests/dycore_stencil_tests/test_fused_velocity_advection_stencil_1_to_7.py#L247

nfarabullini: the stencil test is only meant to check against the numpy reference so as long as the bounds do not cause any errors in the simplest form, we can keep them the way they are

nfarabullini:
code moved to https://github.com/C2SM/icon4py/pull/674

PR #665, opened 2025-02-13
Cleanup py2fgen continued

havogt:
Changes in the model:
- Adds the `ltkeshs` options to diffusion

Changes in the diffusion wrapper:
- The size arrays in `grid_init_diffusion` are extracted from their fields and copied from cp to np. They should not live on the GPU in the first place, which will be cleaned up in a next PR.
- The wrapper has to deal with optional fields: before xp.zeros where passed from the bindings, however that was error-prone as we didn't have all sizes available. After the refactoring of passing all sizes, this is not even an option as the sizes are completely unknown.

Changes in py2fgen:
- Removes the `limited_area` command line flag, this is possible since...
- Adds support for optional fields: if a parameter is annotated with `Optional`, the Fortran parameter will be get the pointer attribute. If the Fortran pointer is `associated` we pass the pointer via `iso_c_bindings`, else we pass the `c_nullptr` and check for `ffi.NULL`
- For each Fortran array pass all its sizes. Before we passed a unique size per dimension which might lead to subtle issues in case of wrong typing. This is also a preparation to support plain arrays (not only `gtx.Field`s).
- Change the default for `ICON4PY_PARALLEL` to `True`. For debugging we keep the possibility to disable parallel mode.
- Introduced a new function `as_field()` to convert the ffi ptrs to `gtx.Field`s


tmp_py2f_gpu/diffusion_plugin.f90
```diff
@@ -0,0 +1,1209 @@
...
+                                  vct_b=c_loc(vct_b), &
+                                  theta_ref_mc=c_loc(theta_ref_mc), &
+                                  wgtfac_c=c_loc(wgtfac_c), &
+                                  e_bln_c_s=c_loc(e_bln_c_s), &
+                                  geofac_div=c_loc(geofac_div), &
+                                  geofac_grg_x=c_loc(geofac_grg_x), &
+                                  geofac_grg_y=c_loc(geofac_grg_y), &
+                                  geofac_n2s=c_loc(geofac_n2s), &
+                                  nudgecoeff_e=c_loc(nudgecoeff_e), &
+                                  rbf_coeff_1=c_loc(rbf_coeff_1), &
```
havogt, line 731: should we c_loc to the first entry

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -220,7 +220,10 @@ def start_index(self, domain: h_grid.Domain) -> gtx.int32:
         if domain.local:
             # special treatment because this value is not set properly in the underlying data.
             return gtx.int32(0)
-        return gtx.int32(self._start_indices[domain.dim][domain()])
+        res = self._start_indices[domain.dim][domain()]
```
havogt, line 223: don't merge this, should be fixed in the wrapper/bindings

tools/src/icon4py/tools/py2fgen/wrapper_utils.py
```diff
@@ -119,3 +128,26 @@ def int_array_to_bool_array(int_array: np.typing.NDArray) -> np.typing.NDArray:
     """
     bool_array = int_array != 0
     return bool_array
+
+
+def as_field(  # type: ignore[no-untyped-def] # CData type not public?
```
havogt, line 133: The previous implementation did more on the codegen side, avoiding a few Python indirections.
When we measure overhead we could go back to do more on the codegen.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -143,6 +143,7 @@ def __init__(
         max_nudging_coeff: float = 0.02,
         nudging_decay_rate: float = 2.0,
         shear_type: TurbulenceShearForcingType = TurbulenceShearForcingType.VERTICAL_OF_HORIZONTAL_WIND,
+        ltkeshs: bool = True,
```
havogt, line 146: Option was missing

tools/src/icon4py/tools/py2fgen/cli.py
```diff
@@ -59,15 +59,13 @@ def parse_comma_separated_list(ctx: click.Context, param: click.Parameter, value
     is_flag=True,
     help="Profile granule runtime and unpacking Fortran pointers into NumPy or CuPy arrays.",
 )
-@click.option("--limited-area", is_flag=True, help="Enable limited area mode.")
```
havogt, line 62: gone, because we support optionals

tools/src/icon4py/tools/py2fgen/settings.py
```diff
@@ -127,17 +127,12 @@ def device(self) -> Device:
         device = device_map[self.icon4py_backend]
         return device
 
-    @cached_property
-    def limited_area(self) -> bool:
-        return env_flag_to_bool("ICON4PY_LAM", False)
-
     @cached_property
     def parallel_run(self) -> bool:
-        return env_flag_to_bool("ICON4PY_PARALLEL", False)
```
havogt, line 136: The default should be to use the MPI exchange. I kept the flag for debugging for now.

tools/src/icon4py/tools/py2fgen/template.py
```diff
@@ -6,13 +6,12 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
-from types import ModuleType
```
havogt, line 9: Probably makes sense to discuss the changes by looking at the generated files.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -87,10 +90,10 @@ def diffusion_init(
     nudgecoeff_e: fa.EdgeField[wpfloat],
     rbf_coeff_1: gtx.Field[gtx.Dims[dims.VertexDim, dims.V2EDim], gtx.float64],
     rbf_coeff_2: gtx.Field[gtx.Dims[dims.VertexDim, dims.V2EDim], gtx.float64],
-    mask_hdiff: fa.CellKField[bool],
-    zd_diffcoef: fa.CellKField[wpfloat],
-    zd_vertoffset: gtx.Field[gtx.Dims[dims.CellDim, dims.E2CDim, dims.KDim], gtx.int32],
-    zd_intcoef: gtx.Field[gtx.Dims[dims.CellDim, dims.E2CDim, dims.KDim], gtx.float64],
+    mask_hdiff: Optional[fa.CellKField[bool]],
+    zd_diffcoef: Optional[fa.CellKField[wpfloat]],
+    zd_vertoffset: Optional[gtx.Field[gtx.Dims[dims.CellDim, dims.C2E2CDim, dims.KDim], gtx.int32]],
```
havogt, line 95: Was wrongly using the E2CDim which caused strange bugs because fields might get the wrong dimensions.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -221,6 +226,25 @@ def diffusion_init(
...
+    }
+    xp = wgtfac_c.array_ns
+    if mask_hdiff is None:
+        mask_hdiff = gtx.zeros(cell_k_domain, dtype=xp.bool_)
+    if zd_diffcoef is None:
+        zd_diffcoef = gtx.zeros(cell_k_domain, dtype=theta_ref_mc.dtype)
+    if zd_intcoef is None:
+        zd_intcoef = gtx.zeros(cell_c2e2c_k_domain, dtype=wgtfac_c.dtype)
+    if zd_vertoffset is None:
+        zd_vertoffset = gtx.zeros(cell_c2e2c_k_domain, dtype=xp.int32)
```
havogt, line 247: Previously the fields that were not present where constructed as zeros in the bindings, however if a have dimensions that are not available for any non-optional fields, we cannot properly construct the fields automagically. It seems cleaner to do this explicitly (and possibly cache them at module level.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -282,6 +306,15 @@ def diffusion_run(
...
 
+    backend = diffusion_wrapper_state["granule"]._backend
+    if hdef_ic is None:
+        hdef_ic = gtx.zeros(w.domain, dtype=w.dtype, allocator=backend)
+    if div_ic is None:
+        div_ic = gtx.zeros(w.domain, dtype=w.dtype, allocator=backend)
+    if dwdx is None:
+        dwdx = gtx.zeros(w.domain, dtype=w.dtype, allocator=backend)
+    if dwdy is None:
+        dwdy = gtx.zeros(w.domain, dtype=w.dtype, allocator=backend)
```
havogt, line 317: They are constructed in every diffusion step. We should cache them... They are only needed because GT4Py doesn't support optionals.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -361,7 +390,7 @@ def grid_init_diffusion(
         num_edges=num_edges,
         vertical_size=vertical_size,
         limited_area=limited_area,
-        on_gpu=True if config_settings.device == "GPU" else False,
+        on_gpu=config_settings.device == settings.Device.GPU,
```
havogt, line 393: string comparison resulted in always `False` which put some fields on host in gpu mode.

havogt: Is not in ci therefore was forgotten in #575 .

tmp_py2f_gpu/diffusion_plugin.h
```diff
@@ -0,0 +1,73 @@
+extern int diffusion_init_wrapper(
```
havogt, line 1: TODO add a test that makes sure the references are updated in case of a change

tools/src/icon4py/tools/py2fgen/template.py
```diff
@@ -50,40 +31,43 @@ class DimensionType(Node):
...
-        # side. So before generating the python wrapper code, we replace occurrences of KHalfDim with KDim
-        self.gtdims = [
-            dimension.value.replace("KHalf", "K") + "Dim" for dimension in self.dimensions
-        ]
-        self.gtdims = [
-            "dims." + gtdim if gtdim not in dir(wrapper_dimension) else gtdim
-            for gtdim in self.gtdims
-        ]
+        self.is_bool = self.d_type == ts.ScalarKind.BOOL
+        self.gtdims = [dim.value if not dim.value == "KHalf" else "K" for dim in self.dimensions]
```
havogt, line 50:
```suggestion
        self.gtdims = [dim.value for dim in self.dimensions]
```
should work now!

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

tools/src/icon4py/tools/py2fgen/template.py
```diff
@@ -220,22 +178,22 @@ def render_fortran_array_sizes(param: FuncParameter) -> str:
...
 {% if _this_node.profile %}import time{% endif %}
 from {{ plugin_name }} import ffi
 {% if _this_node.backend == 'GPU' %}import cupy as cp {% endif %}
-from gt4py.next.iterator.embedded import np_as_located_field
+import gt4py.next as gtx
+from gt4py.next.type_system import type_specifications as ts
 from icon4py.tools.py2fgen.settings import config
 from icon4py.tools.py2fgen import wrapper_utils
-from icon4py.model.common import dimension as dims
+xp = config.array_ns
```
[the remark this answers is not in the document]
havogt: I would like to make the codegen independent of GPU/CPU, therefore I would prefer this over the codegen conditional import. However, I also would like to get rid of that completely. Can I add it to my list for the next PR?

tools/src/icon4py/tools/py2fgen/wrappers/common.py
```diff
@@ -9,29 +9,34 @@
...
+
 from icon4py.model.common import dimension as dims
-from icon4py.model.common.decomposition import definitions
+from icon4py.model.common.decomposition import definitions, mpi_decomposition
 from icon4py.model.common.grid import base, horizontal, icon
 from icon4py.tools.py2fgen.settings import config
 
 
+# TODO(havogt) import needed to register MultNodeRun in get_processor_properties, does the pattern make sense?
+assert hasattr(mpi_decomposition, "get_multinode_properties")
```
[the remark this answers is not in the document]
havogt: `get_processor_properties` is `functools.singledispatch`, but the Multinode setup is only registered if `mpi_decomposition` is imported.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -87,10 +90,10 @@ def diffusion_init(
     nudgecoeff_e: fa.EdgeField[wpfloat],
     rbf_coeff_1: gtx.Field[gtx.Dims[dims.VertexDim, dims.V2EDim], gtx.float64],
     rbf_coeff_2: gtx.Field[gtx.Dims[dims.VertexDim, dims.V2EDim], gtx.float64],
-    mask_hdiff: fa.CellKField[bool],
-    zd_diffcoef: fa.CellKField[wpfloat],
-    zd_vertoffset: gtx.Field[gtx.Dims[dims.CellDim, dims.E2CDim, dims.KDim], gtx.int32],
-    zd_intcoef: gtx.Field[gtx.Dims[dims.CellDim, dims.E2CDim, dims.KDim], gtx.float64],
+    mask_hdiff: Optional[fa.CellKField[bool]],
+    zd_diffcoef: Optional[fa.CellKField[wpfloat]],
```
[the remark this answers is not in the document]
havogt: I agree, we should use the ICON4Py defines. Again I would like to postpone to a next PR.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -221,6 +226,23 @@ def diffusion_init(
         _min_index_flat_horizontal_grad_pressure=nflat_gradp,
     )
 
+    nlev = wgtfac_c.domain[dims.KDim].unit_range.stop - 1  # wgtfac_c has nlevp1 levels
```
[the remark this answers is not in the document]
havogt: Agree, completely random... Actually, the fields in this list are (as of today) not passed to a computation (because the computation is explicitly disabled). I could remove the allocation here, however with more combined stencils, this might break again.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -332,19 +362,47 @@ def grid_init_diffusion(
     vertical_size: gtx.int32,
     limited_area: bool,
 ):
-    # todo: write this logic into template.py
-    if isinstance(limited_area, int):
-        limited_area = bool(limited_area)
+    on_gpu = config_settings.device == settings.Device.GPU
+    xp = c2e.array_ns
```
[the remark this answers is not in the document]
havogt: totally random, any field would do, but this looks like it's actually not needed anyway...

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -332,19 +362,47 @@ def grid_init_diffusion(
...
+
+    # TODO(havogt): add direct support for ndarrays in py2fgen
+    cell_starts = cell_starts.ndarray
+    cell_ends = cell_ends.ndarray
+    vertex_starts = vertex_starts.ndarray
+    vertex_ends = vertex_ends.ndarray
+    edge_starts = edge_starts.ndarray
+    edge_ends = edge_ends.ndarray
+    c_owner_mask = c_owner_mask.ndarray if c_owner_mask is not None else None
```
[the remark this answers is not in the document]
havogt: The `None`s are coming from the test.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -332,19 +362,47 @@ def grid_init_diffusion(
...
+
+    if on_gpu:
+        cp = xp
+        cell_starts = cp.asnumpy(cell_starts)
+        cell_ends = cp.asnumpy(cell_ends)
+        vertex_starts = cp.asnumpy(vertex_starts)
+        vertex_ends = cp.asnumpy(vertex_ends)
+        edge_starts = cp.asnumpy(edge_starts)
+        edge_ends = cp.asnumpy(edge_ends)
+        c_owner_mask = cp.asnumpy(c_owner_mask) if c_owner_mask is not None else None
```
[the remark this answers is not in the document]
havogt: Yes all these lines should go away once we support numpy arrays.

tools/src/icon4py/tools/py2fgen/wrappers/diffusion_wrapper.py
```diff
@@ -282,6 +304,14 @@ def diffusion_run(
         rho=rho,
     )
 
+    if hdef_ic is None:
+        hdef_ic = gtx.zeros(w.domain, dtype=w.dtype, allocator=backend)
```
havogt, line 308: side-note: the `backend` here is taken from the global config, but for now I would keep it.

havogt (commented):
If you are ok, I would like to address the comments in a follow up PR. Mainly, because I have to run ICON manually on each small change that I do.

halungge (approved):
Good to go, to be continued in follow up PR.

muellch (approved):
Looks good!
A lot cleaner now.

PR #666, opened 2025-02-13
Bencher Baseline

kotsaloscv:
This pipeline runs every day at 21 (for the main branch), and pushes the benchmark results to Bencher cloud. In this way, we are creating a historical baseline such that we can compare every feature branch with this baseline.

All the parameters for bencher cloud are set in https://cicd-ext-mw.cscs.ch (currently I am pushing everything in a test project, while an icon4py-bencher project already exists and will be used once we merge this PR).


kotsaloscv:
cscs-ci run baseline

kotsaloscv:
cscs-ci run baseline

kotsaloscv:
@edopao is there any better way to get the name of `gh200` other than doing this -> https://github.com/C2SM/icon4py/pull/666/files#diff-e429d7e54b26aa2173a0cc252f21c81f5fd0ce98430d2d35d6a2433b6cddfccfR93 ?

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,38 @@
...
+      --testbed "ci-runner:$SYSTEM_NAME" \
+      --threshold-measure latency \
+      --threshold-test t_test \
+      --threshold-max-sample-size 64 \
+      --threshold-upper-boundary 0.99 \
+      --thresholds-reset \
+      --err \
+      --adapter python_pytest \
+      --file results.json \
+      "pytest --benchmark-json results.json -sv --benchmark-only --backend=$BACKEND --grid=$GRID ./model"
```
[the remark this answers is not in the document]
kotsaloscv: The above command is the same as the one called by nox -> https://github.com/C2SM/icon4py/blob/main/noxfile.py#L49

Although I am wondering if both should be removed.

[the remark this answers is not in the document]
edopao: I think `-v` is fine to keep, it's about verbosity of pytest command itself. `-s` instead is maybe better to skip, in benchmark mode.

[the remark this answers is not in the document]
kotsaloscv: agree

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,38 @@
+include:
+  - local: 'ci/base.yml'
+
+.benchmark_model_stencils:
+  stage: benchmark
+  script:
+    - curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+    - source "$HOME/.cargo/env"
+    - cargo install --git https://github.com/bencherdev/bencher --branch main --locked --force bencher_cli
+    - curl -LsSf https://astral.sh/uv/install.sh | sh
```
edopao, line 10: Isn't `uv` already installed in the base image? If so, is it possible to delete lines 10-11?

kotsaloscv: you are right. I have removed these lines.

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,38 @@
...
+    - source .venv/bin/activate
+    - |
+      bencher run \
+      --project "$BENCHER_PROJECT" \
+      --token "$BENCHER_API_TOKEN" \
+      --branch main \
+      --testbed "ci-runner:$SYSTEM_NAME" \
+      --threshold-measure latency \
+      --threshold-test t_test \
+      --threshold-max-sample-size 64 \
```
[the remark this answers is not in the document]
kotsaloscv: the metric is essentially one measurement per stencil. So, 64 means that it compares the latest metric with the last 64. I just used the default value, but we can extend as much as we want.

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

iomaganaris:
cscs-ci run benchmark-bencher-baseline

ci/benchmark_bencher_baseline.yml
edopao, line 1: When it is time to merge this PR, I suggest that this file should replace the existing `benchmark.yml`. There is no reason to run them both in daily test.

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

ci/benchmark_bencher_baseline.yml
kotsaloscv: This will indeed happen in https://github.com/C2SM/icon4py/pull/669

kotsaloscv:
@havogt @edopao @iomaganaris 
FYI,  the PR is ready.

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+      --testbed "ci-runner:${SYSTEM_NAME}:${BACKEND}:${GRID}" \
+      --threshold-measure latency \
+      --threshold-test percentage \
+      --threshold-max-sample-size 64 \
+      --threshold-upper-boundary 0.1 \
+      --thresholds-reset \
+      --err \
+      --adapter python_pytest \
+      --file $BENCHER_RESULTS_JSON \
+      "pytest --benchmark-json $BENCHER_RESULTS_JSON -v --benchmark-only --backend=$BACKEND --grid=$GRID ./model"
```
iomaganaris, line 27: I haven't measured the accuracy of the measurements over time without warm up iterations but we could also use:
```suggestion
      "pytest --benchmark-json $BENCHER_RESULTS_JSON -v --benchmark-only --backend=$BACKEND --grid=$GRID ./model --benchmark-warmup=on --benchmark-warmup-iterations=30"
```
For GPU runs specifically it might be helpful (although kernel time measurements are not precise atm)

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+      with open(args.output_file, "w") as file:
+          json.dump(flat_data, file, indent=2)
+      EOF
+    - uv pip install flatten_json
+    - python flatten.py $BENCHER_RESULTS_JSON $BENCHER_RESULTS_JSON
+    - | # Additional fields to BENCHER_RESULTS_JSON for Kibana
+      jq ". | {\"data_stream\": {\"type\": \"logs\", \"dataset\": \"service.icon4py.bencher.baseline.test\", \"namespace\": \"alps\"}} + ." $BENCHER_RESULTS_JSON > results_tmp.json && mv results_tmp.json $BENCHER_RESULTS_JSON
+    - |
+      jq ". + {\"bencher.testbed\": \"ci-runner:${SYSTEM_NAME}:${BACKEND}:${GRID}\"}" $BENCHER_RESULTS_JSON > results_tmp.json && mv results_tmp.json $BENCHER_RESULTS_JSON
+    - | # Upload BENCHER_RESULTS_JSON to Kibana
```
iomaganaris, line 54: Is it documented somewhere how to access the data in Kibana? If yes maybe we could add a pointer to it here

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+      curl -X POST \
+      -H "Content-Type: application/json" \
+      -d @"./${BENCHER_RESULTS_JSON}" \
+      "${CSCS_LOGSTASH_URL}"
+  parallel:
+    matrix:
+      - BACKEND: [gtfn_cpu, gtfn_gpu]
+        GRID: [icon_grid, icon_grid_global]
```
iomaganaris, line 62: For the moment these grids should be sufficient. As a next cycle TODO in my opinion we need to find a larger circuit to use for the GPU tests at least. IIRC the grids provided by `--dataset` are not large enough so we have to find some larger grid

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+    - source "$HOME/.cargo/env"
+    - cargo install --git https://github.com/bencherdev/bencher --branch main --locked --force bencher_cli # Install bencher CLI
+    - uv sync --no-dev --extra=dace --extra=io --extra=testing --extra=$ICON4PY_NOX_UV_CUSTOM_SESSION_EXTRAS --group=test
+    - source .venv/bin/activate
+    - BENCHER_RESULTS_JSON="results_${CI_COMMIT_SHA}_${BACKEND}_${GRID}.json"
+    - | # Run the benchmarks
+      bencher run \
+      --project "$BENCHER_PROJECT" \
+      --token "$BENCHER_API_TOKEN" \
+      --branch main \
```
edopao, line 17: For the daily CI we could try to use:
```suggestion
      --branch $CI_COMMIT_BRANCH \
```

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+
+      with open(args.input_file, "r") as file:
+          data = json.load(file)
+
+      flat_data = flatten(data, separator=".")
+
+      with open(args.output_file, "w") as file:
+          json.dump(flat_data, file, indent=2)
+      EOF
+    - uv pip install flatten_json
```
edopao, line 48: We could install `flatten_json` as part of the `test` dependencies.

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+  stage: benchmark
+  script:
+    - curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+    - source "$HOME/.cargo/env"
+    - cargo install --git https://github.com/bencherdev/bencher --branch main --locked --force bencher_cli # Install bencher CLI
+    - uv sync --no-dev --extra=dace --extra=io --extra=testing --extra=$ICON4PY_NOX_UV_CUSTOM_SESSION_EXTRAS --group=test
+    - source .venv/bin/activate
+    - BENCHER_RESULTS_JSON="results_${CI_COMMIT_SHA}_${BACKEND}_${GRID}.json"
+    - | # Run the benchmarks
+      bencher run \
```
havogt, line 14: I think we should investigate how we can integrate this with nox instead of adding a new way of constructing the environment.

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,68 @@
...
+      --threshold-measure latency \
+      --threshold-test percentage \
+      --threshold-max-sample-size 64 \
+      --threshold-upper-boundary 0.1 \
+      --thresholds-reset \
+      --err \
+      --adapter python_pytest \
+      --file $BENCHER_RESULTS_JSON \
+      "pytest --benchmark-json $BENCHER_RESULTS_JSON -v --benchmark-only --benchmark-warmup=on --benchmark-warmup-iterations=30 --backend=$BACKEND --grid=$GRID ./model"
+    - | # Script to flatten the BENCHER_RESULTS_JSON file. Needed for Kibana.
```
havogt, line 28: What about creating this script as a separate file instead of on-the-fly construction?

havogt (changes requested):

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

ci/benchmark_bencher_baseline.yml
kotsaloscv: I would prefer to create it on the fly such that we do not clutter the folder with scripts.

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -163,3 +163,12 @@ def pytest_runtest_setup(item):
         backend,
         is_datatest=item.config.getoption("--datatest"),
     )
+
+
+# Replace "fullname" of pytest benchmarks with a shorter name for better readability in bencher
+def pytest_benchmark_update_json(output_json):
```
havogt, line 169:
```suggestion
# pytest benchmark hook, see:
#     https://pytest-benchmark.readthedocs.io/en/latest/hooks.html#pytest_benchmark.hookspec.pytest_benchmark_update_json
def pytest_benchmark_update_json(output_json):
    "Replace 'fullname' of pytest benchmarks with a shorter name for better readability in bencher."
```

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -163,3 +163,12 @@ def pytest_runtest_setup(item):
...
+    for bench in output_json["benchmarks"]:
+        # Store the original fullname
+        bench["fullname_original"] = bench["fullname"]
```
havogt, line 172: where is this used?

ci/docker/base.Dockerfile
```diff
@@ -24,13 +24,22 @@ RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
...
 
+# Install Rust using rustup
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustc --version && which rustc && cargo --version && which cargo
+
+# Install Bencher for performance monitoring
+RUN cargo install --git https://github.com/bencherdev/bencher --branch main --locked --force bencher_cli
+RUN bencher --version
+
```
havogt, line 42: Why do we install from sources instead of the binaries? https://bencher.dev/docs/how-to/install-cli/#linux-mac--unix or https://bencher.dev/docs/how-to/install-cli/#packages

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,62 @@
+include:
+  - local: 'ci/base.yml'
+
+.benchmark_model_stencils:
+  stage: benchmark
+  script:
+    - | # Run the benchmarks
```
havogt, line 7: I was thinking of putting this inside a new nox session in the noxfile.py. This would make it easier for debugging in case something doesn't work in the future. In the nox tutorial there is an example with coverage, which is very similar to our use-case https://nox.thea.codes/en/stable/tutorial.html#requiring-sessions

ci/benchmark_bencher_baseline.yml
```diff
@@ -0,0 +1,62 @@
...
+      --threshold-measure latency \
+      --threshold-test percentage \
+      --threshold-max-sample-size 64 \
+      --threshold-upper-boundary 0.1 \
+      --thresholds-reset \
+      --err \
+      --adapter python_pytest \
+      --file results.json \
+      "nox -s benchmark_model-3.10 -- --backend=$BACKEND --grid=$GRID"
+    - | # Script to flatten the results.json file. Needed for Kibana.
```
havogt, line 22: I think it's essential to extract the kibana related setup into a script for maintainability. But, actually this could be even removed in this first setup.

noxfile.py
```diff
@@ -47,14 +47,13 @@
...
-@nox.parametrize("subpackage", MODEL_SUBPACKAGE_PATHS)
-def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> None:
+def benchmark_model(session: nox.Session) -> None:
     """Run pytest benchmarks for selected icon4py model subpackages."""
     _install_session_venv(session, extras=["dace", "io", "testing"], groups=["test"])
 
-    with session.chdir(f"model/{subpackage}"):
-        session.run(*"pytest -sv --benchmark-only".split(), *session.posargs)
-
+    results_json = os.path.abspath("results.json")  # Store results in the top directory
```
havogt, line 54:
```suggestion
    results_json_path = os.path.abspath("results.json")  # Store results in the top directory
```

noxfile.py
```diff
@@ -47,14 +47,13 @@
...
+    with session.chdir(f"model"):
+        session.run("pytest", "-v", "--benchmark-only", "--benchmark-warmup=on", "--benchmark-warmup-iterations=30", f"--benchmark-json={results_json}", *session.posargs)
```
havogt, line 56:
```suggestion
        session.run("pytest", "-v", "--benchmark-only", "--benchmark-warmup=on", "--benchmark-warmup-iterations=30", f"--benchmark-json={results_json_path}", *session.posargs)
```

havogt (changes requested):
I have a few more points to improve maintainability. Regarding nox setup, see my comment on how I meant it.

model/testing/src/icon4py/model/testing/pytest_config.py
iomaganaris: (Replying here since I've added this change)
`fullname_original` is not used anywhere at the moment. I just kept it in case this hook needs to be extended in the future or there is some need to read the original `fullname` by another tool that processes the generated `.json` file. It can be removed.

ci/docker/base.Dockerfile
kotsaloscv: I tried but the binaries do not support our architecture.

ci/benchmark_bencher_baseline.yml
kotsaloscv: I thought exactly the same. Possibly, we will skip the Kibana service and keep only the self-hosting stuff. So for now, I would leave it as is.

ci/docker/base.Dockerfile
havogt: For me the ARM64 binaries work on alps.

ci/benchmark_bencher_baseline.yml
havogt: No, I meant skip this setup. Before merge we should make it clean or remove it.

ci/docker/base.Dockerfile
kotsaloscv: In the CI it did not work though

ci/benchmark_bencher_baseline.yml
kotsaloscv: done

ci/benchmark_bencher_baseline.yml
kotsaloscv: I skip the kibana setup for now, and we can add it if needed.

ci/docker/base.Dockerfile
kotsaloscv: I will re-try it ;) (just to be sure)

kotsaloscv: Indeed, it works. I commit it now.

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
@havogt do you think it can be merged now?
The CI passes as expected and it uploads the data in https://bencher.dev/perf/c2sm-icon4py (project created yesterday by Matthieu from EXCLAIM)

egparedes (commented):
I have a couple of questions and comments, mostly for my own understanding...

ci/base.yml
```diff
@@ -90,6 +90,8 @@ build_image_aarch64:
   extends: [.container-runner-santis-gh200, .test_template]
   needs: [build_image_aarch64]
   variables:
+    RUNNER: "ci-runner"
+    SYSTEM_NAME: "gh200"
```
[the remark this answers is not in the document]
edopao: We could get the vCluster name from the variable `FIRECREST_SYSTEM`, which would give us "santis" in this case.

model/testing/src/icon4py/model/testing/pytest_config.py
iomaganaris: The original `fullname` also includes the file where the test was found but I agree that probably it won't ever be necessary

noxfile.py
```diff
@@ -47,14 +47,31 @@
 # TODO(egparedes): Add backend parameter
 # TODO(edopao,egparedes): Change 'extras' back to 'all' once mpi4py can be compiled with hpc_sdk
 @nox.session(python=["3.10", "3.11"])
-@nox.parametrize("subpackage", MODEL_SUBPACKAGE_PATHS)
-def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> None:
+def benchmark_model(session: nox.Session) -> None:
```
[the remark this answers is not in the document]
kotsaloscv: By iterating over the different folders, we generate multiple json files from the performance monitoring. I wanted to avoid this, and have a single report (json file), which will be uploaded by Bencher.

ci/base.yml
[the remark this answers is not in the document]
egparedes: I referred to `SYSTEM_NAME`, but I think I actually mixed two different questions:
- what should be used to create bencher's testbed: the actual system name (e.g. `saints.alps.cscs.ch` / `$FIRECREST_SYSTEM`) or some kind of tag identifying the software / hardware stack (e.g. `cscs-uenv-gh200`)?
- what name should the environment variable with the information have?

My comment was a bit confusing but I meant to ask about the first point, since the answer to the second is kind of clear once the first is decided.

noxfile.py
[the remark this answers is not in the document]
kotsaloscv: FYI, @egparedes I have updated the nox file and kept the original approach.

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

ci/docker/base.Dockerfile
```diff
@@ -24,13 +24,22 @@ RUN apt-get update -qq && apt-get install -qq -y --no-install-recommends \
...
     git \
-    rustc \
+    jq \
     htop && \
     rm -rf /var/lib/apt/lists/*
 
+# Install Rust using rustup
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
+ENV PATH="/root/.cargo/bin:${PATH}"
+RUN rustc --version && which rustc && cargo --version && which cargo
```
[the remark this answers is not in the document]
egparedes: I meant the whole rust compiler installation, since now bencher is installed directly from binaries.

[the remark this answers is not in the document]
kotsaloscv: it is needed in the other PR (about feature branches -> https://github.com/C2SM/icon4py/pull/669), where we need to compile Bencher's fork by Andreas Fink to allow for GitHub comments (from Bencher). Eventually. Andrea's will create a PR in Bencher but for now we need rust for this reason.

noxfile.py
```diff
@@ -52,9 +54,43 @@ def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> No
...
-
+        session.run(*f"pytest \
+                       -v \
+                       --benchmark-only \
+                       --benchmark-warmup=on \
+                       --benchmark-warmup-iterations=30 \
+                       --benchmark-json={results_json_path}".split(),
+                       *session.posargs)
+
+@nox.session(python=["3.10", "3.11"], requires=["benchmark_model-{python}" + f"({subpackage.id})" for subpackage in MODEL_SUBPACKAGE_PATHS])
```
egparedes, line 67: Have you tried to require the session without any argument? At least from the command line, when you omit the arguments it will run that session for all possible arguments so I guess it should work here also.
```suggestion
@nox.session(python=["3.10", "3.11"], requires=["benchmark_model-{python}"])
```

kotsaloscv: That was the first thing I tried, but it does not work :)
It runs only for one of the sub packages.

noxfile.py
```diff
@@ -52,9 +54,43 @@ def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> No
...
+        with open(file, "r") as f:
+            data = json.load(f)
+            merged_results["benchmarks"].extend(data["benchmarks"])
+    with open(bencher_json_file_name, "w") as f:
+        json.dump(merged_results, f, indent=4)
+
+    session.run(*f"bencher run \
+                   --project {os.environ['BENCHER_PROJECT']} \
+                   --token {os.environ['BENCHER_API_TOKEN']} \
+                   --branch main \
```
egparedes, line 83: Shouldn't the branch value be read directly from git? If we want to make sure that we only submit results to bencher from the `main` branch, maybe we should check at the beginning of the session, before running the tests, wether the currently checked out branch is `main`.

kotsaloscv: The problem is that in the CI the folder does not contain the .git folder. So we cannot get any info for the repo, except for accessing the CI env vars. In any case, this pipeline is supposed to be executed automatically only on the main branch. I would complicate it more for now.

ci/docker/base.Dockerfile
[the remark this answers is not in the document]
kotsaloscv: Do you mind if I keep it because Bencher is installed under /root/.cargo and so it demands an active rust installation? In any case, this is needed for Bencher.

ci/base.yml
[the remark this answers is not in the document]
kotsaloscv: How about this latest commit?

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

noxfile.py
```diff
@@ -52,9 +54,57 @@ def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> No
...
+                merged_results["benchmarks"].extend(data["benchmarks"])
+                # preserve the first file's metadata for a valid pytest-benchmark json file
+                for key in data.keys():
+                    if key != "benchmarks" and key not in merged_results:
+                        merged_results[key] = data[key]
+        except:
+            # Empty file, i.e. no benchmarks
+            continue
+    with open(bencher_json_file_name, "w") as f:
+        json.dump(merged_results, f, indent=4)
```
havogt, line 90: What's the reason for merging these files instead of uploading separately? Then we would only `requires` the specific subproject, which seems simpler and more in the style of the rest of the nox setup.

havogt (changes requested):
- [ ] The PR description mentions that this runs at a specific time, but as far as I can see it runs on merge. This is perfect for me, just update the PR description.
- [ ] I would remove the non-bencher benchmark pipeline.
- [ ] I think we should discuss which benchmarks to run before merging this, otherwise we are polluting the history with a lot of meaningless information. But maybe it's only me who is scared of too much data.

noxfile.py
kotsaloscv: By gathering all the reports in a unified one, I create less clutter in the Bencher Cloud Dashboard. Otherwise, it will be very messy for the user to navigate through all the reports. I prefer this way.

kotsaloscv:
> * [ ]  The PR description mentions that this runs at a specific time, but as far as I can see it runs on merge. This is perfect for me, just update the PR description.
> * [ ]  I would remove the non-bencher benchmark pipeline.
> * [ ]  I think we should discuss which benchmarks to run before merging this, otherwise we are polluting the history with a lot of meaningless information. But maybe it's only me who is scared of too much data.

FYI @havogt 
Regarding your points.

- I have created a pipeline in https://cicd-ext-mw.cscs.ch (benchmark-bencher-baseline), which has a chron schedule. The idea of this PR is to run every day and at every push to the main, such that we have an updated baseline.
- The non-bencher pipeline will be removed with this PR -> https://github.com/C2SM/icon4py/pull/669. This is because the non-bencher pipeline is executed for the feature branches, as is the case for my other PR.
- I would favor on merging it now for two reasons: (1) I want the changes made in nox/docker-base etc for my other branches (Bencher feature branches & self-hosting), (2) we can archive any reports that we do not want or even purge the project in Bencher and restart it (very easy)

kotsaloscv:
cscs-ci run benchmark-bencher-baseline

havogt (approved):
Looks good to me.

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

PR #671, opened 2025-02-17
CI: Fix for nox environment variables

edopao:
Read icon4py variable from OS environment, not nox environment. This became necessary after the nox version was upgraded. Refer to nox PR https://github.com/wntrblm/nox/pull/874


edopao:
cscs-ci run default

PR #675, opened 2025-02-19
Add memray statistics to bencher

iomaganaris:
- Add memory high watermark as reported by memray to bencher metrics
- Use [specific fork of pytest-memray](https://github.com/iomaganaris/pytest-memray) that dumps memray data to pytest json file
- Tests needed to be run twice to avoid memray overhead in runtime reports
- Push data to bencher using BMF json file

### DIFF
--- a/.github/workflows/bencher-cleanup.yml
+++ b/.github/workflows/bencher-cleanup.yml
@@ -1,0 +1,19 @@
+on:
+    pull_request:
+      types: [closed]
+
+jobs:
+    archive_pr_branch:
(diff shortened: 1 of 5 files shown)


msimberg:
@iomaganaris do you think it's worth keeping this branch around? Do you have any plans to revive it? I like the idea, but if there's no plan on continuing this particular PR/branch, I'd close it?

iomaganaris:
This was a PoC and since there was no interest to add this we can close this

PR #677, opened 2025-02-20
Make stencil test for solve_tridiagonal_matrix_for_w_forward_sweep

edopao:
During testing of the dace backend, it was found that the test for `solve_tridiagonal_matrix_for_w_forward_sweep` was using a fixed backend. This PR converts the existing test into a `StencilTest` so that it can run with a parameterized backend.

### DIFF
--- a/model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
+++ b/model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
@@ -7,108 +7,92 @@
 # SPDX-License-Identifier: BSD-3-Clause
 import gt4py.next as gtx
 import numpy as np
-from gt4py.next.program_processors.runners.gtfn import run_gtfn
+import pytest
 
 from icon4py.model.atmosphere.dycore.stencils.solve_tridiagonal_matrix_for_w_forward_sweep import (
     solve_tridiagonal_matrix_for_w_forward_sweep,
 )
 from icon4py.model.common import dimension as dims
-from icon4py.model.common.grid.simple import SimpleGrid
 from icon4py.model.common.type_alias import vpfloat, wpfloat
 from icon4py.model.common.utils.data_allocation import random_field
+from icon4py.model.testing.helpers import StencilTest
 
 
-def solve_tridiagonal_matrix_for_w_forward_sweep_numpy(
-    vwind_impl_wgt: np.array,
-    theta_v_ic: np.array,
-    ddqz_z_half: np.array,
-    z_alpha: np.array,
-    z_beta: np.array,
-    z_exner_expl: np.array,
-    z_w_expl: np.array,
-    z_q_ref: np.array,
-    w_ref: np.array,
-    dtime,
-    cpd,
-) -> tuple[np.array]:
-    z_q = np.copy(z_q_ref)
-    w = np.copy(w_ref)
-    vwind_impl_wgt = np.expand_dims(vwind_impl_wgt, axis=-1)
+class TestSolveTridiagonalMatrixForWForwardSweep(StencilTest):
+    PROGRAM = solve_tridiagonal_matrix_for_w_forward_sweep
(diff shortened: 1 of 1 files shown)


edopao:
cscs-ci run default

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
```diff
@@ -7,108 +7,92 @@
...
 
-    z_gamma = dtime * cpd * vwind_impl_wgt * theta_v_ic / ddqz_z_half
-    z_a = np.zeros_like(z_gamma)
-    z_b = np.zeros_like(z_gamma)
-    z_c = np.zeros_like(z_gamma)
-    z_g = np.zeros_like(z_gamma)
+    @staticmethod
+    def reference(
+        grid,
+        vwind_impl_wgt: np.array,
```
nfarabullini, line 28: we use officially `np.ndarray`
```suggestion
        vwind_impl_wgt: np.ndarray,
```

nfarabullini (changes requested):
It's good that you are converting this into a proper stencil test, however please keep the numpy version out of the class and call it directly in the reference as it might be used later on into a bigger test for combined stencils

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
edopao: You are right. I searched online and found this:
> numpy.array is a function that returns a numpy.ndarray object.
> There is no object of type numpy.array.

I copied this type annotation from another place where it was wrong. I will search and replace in all files.

edopao: Maybe better to only fix this file, in this PR, and do the search and replace in all files in a separate PR.

edopao:
cscs-ci run default

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
nfarabullini: Yes focus only on this. We have been fixing these data types everywhere but there are so mane tests that we always end up missing some... sorry about the fact that another one was wrong

edopao:
cscs-ci run default

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
```diff
@@ -54,61 +55,72 @@ def solve_tridiagonal_matrix_for_w_forward_sweep_numpy(
...
-    assert np.allclose(
-        z_q_ref[h_start:h_end, v_start:v_end], z_q.asnumpy()[h_start:h_end, v_start:v_end]
-    )
-    assert np.allclose(
-        w_ref[h_start:h_end, v_start:v_end], w.asnumpy()[h_start:h_end, v_start:v_end]
-    )
+        h_start = 0
+        h_end = gtx.int32(grid.num_cells)
+        v_start = 1
+        v_end = gtx.int32(grid.num_levels)
```
nfarabullini, line 108: I don't think that we need the int32 type cast but I might be wrong

nfarabullini (approved):

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_forward_sweep.py
edopao: We are enforcing the integer representation on 32 bits. The grid attributes have `int` type, so they are platform dependent. I guess this is why we do the cast here, most stencil tests do it.

PR #679, opened 2025-02-21
Remove jabw init savepoint

halungge:
(Cleanup) This savepoint only contains 
 - constant values (`PARAMETER`s from `mo_nh_jabw_exp.f90`) 
 - fields that are initialized with a constant value (surface pressure) or zero

There is no need to keep it in the serialization and read it in via serialbox.


halungge:
cscs-ci run default

OngChia (approved):

PR #683, opened 2025-02-26
Substepping outside dycore, refactor grid_wrapper, add mean_cell_area and vol_flx_ic

muellch:
Model:
- Pull substepping outside of the dycore
- fix bounds in predictor
Fortran wrappers:
- Extract grid_wrapper instead of one grid_init per diffusion/dycore granule
- Add mean_cell_are and vol_flx_ic to interface
- Change name `max_nudge_coeff` to indicate it's the ICON scaled one that is passed to the bindings
- Adjust nflat_gradp to Python indexing in the bindings.


model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -213,7 +213,10 @@ def start_index(self, domain: h_grid.Domain) -> gtx.int32:
         if domain.local:
             # special treatment because this value is not set properly in the underlying data.
             return gtx.int32(0)
-        return gtx.int32(self._start_indices[domain.dim][domain()])
+        if hasattr(self._end_indices[domain.dim][domain()], "get"):
```
havogt, line 216: this is removed again in #685 which is open against this PR.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -856,12 +853,6 @@ def time_step(
                 offset_provider={},
             )
 
-        self.update_time_levels_for_velocity_tendencies(
```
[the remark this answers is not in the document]
havogt: Ah this is the swapping.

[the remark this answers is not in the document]
muellch: Yes. :)

model/common/src/icon4py/model/common/grid/icon.py
muellch: Perfect.

havogt:
cscs-ci run default

model/driver/src/icon4py/model/driver/icon4py_driver.py
```diff
@@ -226,6 +226,11 @@ def _do_dyn_substepping(
                 f"simulation date : {self._simulation_date} substep / n_substeps : {dyn_substep} / "
                 f"{self.n_substeps_var} , is_first_step_in_simulation : {self._is_first_step_in_simulation}"
             )
+            self.solve_nonhydro.update_time_levels_for_velocity_tendencies(
```
halungge, line 229: Can you move that function out  `solve_nonhydro`? As the swapping is no longer in the dycore this seems to be the wrong place. It is part of the driving code. So rather belongs here and not anymore in the dycore.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -1289,19 +1280,20 @@ def run_predictor_step(
...
-                c_lin_e=self._interpolation_state.c_lin_e,
-                z_dexner_dz_c_1=self.z_dexner_dz_c_1,
-                z_gradh_exner=z_fields.z_gradh_exner,
-                horizontal_start=self._start_edge_nudging_level_2,
-                horizontal_end=self._end_edge_local,
-                vertical_start=self._vertical_params.nflatlev,
-                vertical_end=gtx.int32(self._vertical_params.nflat_gradp + 1),
-                offset_provider=self._grid.offset_providers,
-            )
+            if self._vertical_params.nflatlev < gtx.int32(self._vertical_params.nflat_gradp + 1):
```
halungge, line 1283: Can it happen that `nflatlev>= nflat_gradp + 1` ? Shouldn't the condition rather be checked when computing these values at initialization?  The other questions is does gt4py crash for a "negative" or "empty" domain?

tools/tests/py2fgen/test_dycore_wrapper.py
```diff
@@ -256,6 +245,7 @@ def test_dycore_wrapper_granule_inputs(
 
     # PrepAdvection
     vn_traj = sp.vn_traj()
+    vol_flx_ic = data_alloc.zero_field(icon_grid, dims.CellDim, dims.KDim)  # TODO sp.vol_flx_ic()
```
halungge, line 248: hm, you wanted the `vol_flx_ic` in the init savepoint? I think I only added it to the exit... shame on me...

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
havogt: Yes, it crashed for an empty domain. We should probably fix that on the GridTools side.

model/driver/src/icon4py/model/driver/icon4py_driver.py
havogt: Should we extract the `do_dyn_substep` from the driver? At least in tests we duplicate the swapping logic when we test the full timestep? Maybe in a later cleanup?

model/atmosphere/dycore/tests/dycore_tests/test_solve_nonhydro.py
```diff
@@ -932,6 +932,10 @@ def test_run_solve_nonhydro_multi_step(
         at_first_substep = i_substep == 0
         at_last_substep = i_substep == (ndyn_substeps - 1)
 
+        solve_nonhydro.update_time_levels_for_velocity_tendencies(
```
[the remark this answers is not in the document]
havogt: done, had to re-introduce the swapping depending on the linit from the serialization (for the second parametrization)

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
halungge: Yes, I would suggest so. Anyway, we can leave this as a workaround for now. Can you open an issue in GT4Py?

havogt: On a second thought, I am not sure if it should be legal, because that could also hide user errors. At least GT4Py should warn that it doesn't execute a program. Here is the issue https://github.com/GridTools/gt4py/issues/1891

havogt:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
halungge: Thanks.  I agree it could hide errors, but there are valid cases for it: for example if you run something like APE every layer is flat, so if you try to compute something from some layer up to the first flat layer, that might collapse (what probably happens in the example here) or in the horizontal if you compute something only on the boundaries, but you have a global model, then there are no boundaries. The user would then have to always do an extra
`if domain-is-not-empty` check what the if condition above essentially is.

havogt:
tests passed in https://github.com/C2SM/icon4py/pull/687

PR #685, opened 2025-02-27
Dycore granule multinode

havogt:


tools/src/icon4py/tools/py2fgen/wrappers/dycore_wrapper.py
```diff
@@ -236,11 +238,11 @@ def solve_nh_init(
     )
 
     # datatest config CellParams
-    cell_geometry = grid_states.CellParams.from_global_num_cells(
+    cell_geometry = grid_states.CellParams(
         cell_center_lat=cell_center_lat,
         cell_center_lon=cell_center_lon,
         area=cell_areas,
-        global_num_cells=dycore_wrapper_state["grid"].global_properties.num_cells,
+        mean_cell_area=mean_cell_area,
```
havogt, line 245: @halungge Before this change, mean_cell_area was computed, I guess it's safer to use this approach.

tools/src/icon4py/tools/py2fgen/wrappers/dycore_wrapper.py
```diff
@@ -464,34 +467,4 @@ def grid_init(
...
-        e2c=e2c.ndarray,
-        c2e2c=c2e2c.ndarray,
-        e2c2e=e2c2e.ndarray,
-        e2v=e2v.ndarray,
-        v2e=v2e.ndarray,
-        v2c=v2c.ndarray,
-        e2c2v=e2c2v.ndarray,
-        c2v=c2v.ndarray,
-    )
+    dycore_wrapper_state["grid"] = diffusion_wrapper.diffusion_wrapper_state["grid"]
```
havogt, line 470: this is a temporary workaround to just use the diffusion grid, which is the same except it can already do multinode.

havogt:
cscs-ci run default

tools/src/icon4py/tools/py2fgen/wrappers/dycore_wrapper.py
halungge: yes. ICON does several things for this quantity: if you run with a DWD grid file (icosahedron) mean_cell_area is computed from the R and B parameter of the grid as we do it in icon4py. If you use a MPI-M generated grid, it is computed in their grid_generator and they do a global reduction over the cell_areas for the icosahedron. (If you run on a torus it is analytically computed. ) So you don't know what you get and for limited_area models there might be a difference between the two approaches.

havogt:
cscs-ci run default

PR #687, opened 2025-03-03
fix max_nudge_coeff and nflat_grad

havogt:
I propose to merge into #529. Compatible with https://github.com/C2SM/icon-exclaim/pull/324


havogt:
cscs-ci run default

PR #691, opened 2025-03-05
CI: cleanup base_image gitlab config

edopao:
This cleanup is made possible by the change in CI-Ext: switched the container-builder from the compute nodes on the clusters to the build-farm.


edopao:
cscs-ci run default

edopao:
cscs-ci run default

havogt (approved):

PR #693, opened 2025-03-07
remove duplicate connectivity read from grid_manager.py

halungge:
(fix) two connectivity fields were read twice


halungge:
cscs-ci run default

iomaganaris (approved):
LGTM

PR #669, opened 2025-02-14
Bencher: feature branches

kotsaloscv:
Compare the performance results of the current branch (PR's branch) with Bencher's baseline (main branch) (https://github.com/C2SM/icon4py/pull/666 & https://github.com/C2SM/icon4py/pull/695)

This pipeline compares the performance of the feature branch with the baseline found in Bencher cloud (not the self-hosted one -for simplicity and not any other reason-)

Before merging any feature branch to main, the author should run `cscs-ci run benchmark-bencher` (before it was just `cscs-ci run benchmark`). If there is performance regression, the pipeline fails and the PR author can inspect which tests are failing.

PR #695, opened 2025-03-11
Bencher Baseline: Enable Self-Hosting

kotsaloscv:
This PR simply adds the `--host {os.environ['BENCHER_HOST']}` option to `bencher run`.

To use Bencher cloud, `BENCHER_HOST` needs to be `https://api.bencher.dev`.
To use the self-hosted Bencher, `BENCHER_HOST` needs to be `https://bencher-api.tds.cscs.ch` (this link will change in the future as we currently use a testing server).

The idea is to use both Bencher Cloud and self-hosting (until we decide how to proceed). Therefore, in the `https://cicd-ext-mw.cscs.ch`, there are two pipelines `benchmark-bencher-baseline` & `benchmark-bencher-baseline-self-hosted` where the only difference is the value of `BENCHER_HOST` (and the bencher api token ofc).

You can access the self-hosted Bencher in this address (specifically for the icon4py project) -> `https://bencher.tds.cscs.ch/perf/c2sm-icon4py`


kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

noxfile.py
```diff
@@ -67,48 +67,50 @@ def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> No
...
-        --file {bencher_json_file_name}".split(),
-        external=True,
-    )
+    bencher_json_file_name = f"results_{session.python}_{subpackage.replace('/', '_')}.json"
+    if valid_results_file(bencher_json_file_name):
+        session.run(
+            *f"bencher run \
+            --project {os.environ['BENCHER_PROJECT']} \
+            --token {os.environ['BENCHER_API_TOKEN']} \
+            --host {os.environ['BENCHER_HOST']} \
```
havogt, line 101: I would expose the `BENCHER_HOST` in the CI pipeline file to make it more transparent where data goes. Also the `BENCHER_PROJECT` could go there. I assume currently their are set in the ci interface?

noxfile.py
```diff
@@ -67,48 +67,50 @@ def benchmark_model(session: nox.Session, subpackage: ModelSubpackagePath) -> No
             success_codes=[0, NO_TESTS_COLLECTED_EXIT_CODE],
         )
 
-def merge_pytest_benchmark_results(bencher_json_file_name: str) -> None:
-    """Gather all benchmark results files and merge them into a single file."""
-    merged_results = {"benchmarks": []}
-    files = glob.glob("results_*.json")
-    for file in files:
+def valid_results_file(file_name: str) -> bool:
```
havogt, line 70: Why is this needed? It seems currently it is not an error if the file is e.g. corrupted.

noxfile.py
kotsaloscv: Indeed this is a bit more clear. Although, if we keep both options for now (i.e. cloud & self-hosting), this means that we will have two yml files where we set the above vars. With the current solution (set them in cicd-ext.cscs.ch), we can have just one pipeline file.

noxfile.py
kotsaloscv: This is needed because some of the sub packages do not have any benchmark tests, but they generate empty json files. For this reason, we need this function to check these cases.

noxfile.py
havogt: Ah I understand now how the setup works.

noxfile.py
kotsaloscv: Indeed this is what I still prefer. Ideally, I would like to delete the subpackage parametrization and run pytest-benchmarks in the model folder. In this way, we will upload way less stuff in Bencher cloud. Regarding your question, by merging the different json files, I had some failures to push the reports to Bencher (weirdly enough only for the self-hosted version). This is way I proceeded with the subpackage by subpackage approach.

Although, let me know what you think about the  subpackage parametrization and I could change it.

kotsaloscv: @havogt
FYI, I have applied the change, where I remove the subpackages. I prefer way more this version. It is simpler and very straightforward. Shall we proceed like this?

kotsaloscv:
@havogt I see that I have some issues when uploading large files to the self-hosted server (this is why the pipeline fails). I will have to ask the experts on kubernetes clusters on how to avoid these issues (possibly tune the settings).

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

noxfile.py
```diff
@@ -49,54 +49,33 @@
...
-
+    """
+    Run pytest benchmarks and upload them using Bencher (https://bencher.dev/) (cloud or self-hosted).
+    This session is used only on the main branch to create the historical baseline.
+    The historical baseline is used to compare the performance of the code in the PRs.
+    """
     session.run(
         *f"bencher run \
         --project {os.environ['BENCHER_PROJECT']} \
         --token {os.environ['BENCHER_API_TOKEN']} \
```
iomaganaris, line 77: I've noticed here that `nox` output in the CI leaks the token of the self hosted bencher. Probably we should find a way to keep this secret

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
@havogt 
FYI, I have changed the kubernetes cluster settings and now we can upload large multiple files without any issues.
Any input on removing the subpackages approach and simply running pytest-benchmark on the model folder as a whole?

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
> Let's not forget to update the automatic comment from `github-actions` regarding how to launch the various pipelines

@iomaganaris this happens in this PR about feature branches -> https://github.com/C2SM/icon4py/pull/669

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

iomaganaris:
> > Let's not forget to update the automatic comment from `github-actions` regarding how to launch the various pipelines
> 
> @iomaganaris this happens in this PR about feature branches -> #669

Ah you're right. `benchmark-bencher-baseline-self-hosted` is not supposed to be launched with a comment 🙈 Thanks for the info

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

noxfile.py
kotsaloscv: I found a way to protect the secret env vars -> https://github.com/C2SM/icon4py/blob/506a183dc935033edbb6dba6db3ddde28229ab8f/noxfile.py#L83

Bencher uses these env vars if they exist. So, if we do not pass them explicitly in the bencher run command, then they are not exposed and bencher uses the ones that are already exported in the env.

kotsaloscv: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504674/-/jobs/9415342313#L1680

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
FYI,
@havogt @egparedes @iomaganaris 
As discussed yesterday with Enrique, I have implemented the framework to add more tests in benchmarks. For now, I have added only one from the diffusion granule in order to discuss the approach. Once we converge to the approach, it will be very easy to add many more tests.

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run benchmark-bencher-baseline-self-hosted

kotsaloscv:
cscs-ci run default

kotsaloscv:
launch jenkins spack

havogt:
cscs-ci run default

havogt (approved):
lgtm

PR #697, opened 2025-03-12
Fix types dycore stencil tests

halungge:
Fixing typing issues in `dycore` `StencilTest`s


halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_back_substitution.py
```diff
@@ -23,7 +26,12 @@ class TestSolveTridiagonalMatrixForWBackSubstitution(StencilTest):
     OUTPUTS = ("w",)
 
     @staticmethod
-    def reference(grid, z_q: np.array, w: np.array, **kwargs) -> dict:
+    def reference(
+        connectivities: dict[gtx.Dimension, np.ndarray],
+        z_q: np.ndarray,
+        w: np.ndarray,
+        **kwargs: Any,
+    ) -> dict:
```
egparedes, line 34: Just a question (which applies to all files): shouldn't the return type of the reference method return something like `dict[str, Field | scalar]`?

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_back_substitution.py
```diff
@@ -34,7 +42,7 @@ def reference(grid, z_q: np.array, w: np.array, **kwargs) -> dict:
         return dict(w=w_new)
 
     @pytest.fixture
-    def input_data(self, grid):
+    def input_data(self, grid: base.BaseGrid) -> dict:
```
egparedes, line 45: Same as above, shouldn't the return type of the `input_data` method return something like `dict[str, Field | scalar]`?

egparedes (approved):
I think there are two return type annotations that could be more restricted, but it's not really a blocker for merging this PR, which is already quite large and it clearly improves the current state.

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_back_substitution.py
halungge: I think that is correct. Don't know whether I want to touch it again though...

halungge:
cscs-ci run default

model/atmosphere/dycore/tests/dycore_stencil_tests/test_solve_tridiagonal_matrix_for_w_back_substitution.py
halungge: cscs-ci run default

halungge:
cscs-ci run default

PR #703, opened 2025-03-25
Add more benchmark tests

kotsaloscv:


kotsaloscv:
@egparedes any feedback on these changes?
Thanks!

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -171,4 +171,4 @@ def pytest_benchmark_update_json(output_json):
     "Replace 'fullname' of pytest benchmarks with a shorter name for better readability in bencher."
     for bench in output_json["benchmarks"]:
         # Replace fullname with name and filter unnecessary prefix and suffix
-        bench["fullname"] = bench["name"].replace("test_", "").replace("_benchmark", "")
+        bench["fullname"] = bench["name"].replace("test_", "")
```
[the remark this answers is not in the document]
kotsaloscv: it is still needed because `fullname` (which is used by bencher) includes the full path of the test, while the `name` is just the name of the test.

model/atmosphere/diffusion/tests/diffusion_tests/test_diffusion.py
```diff
@@ -504,13 +509,17 @@ def test_run_diffusion_single_step(
...
-    diffusion_granule.run(
+    benchmark(
+        diffusion_granule.run,
         diagnostic_state=diagnostic_state,
         prognostic_state=prognostic_state,
         dtime=dtime,
     )
 
-    verify_diffusion_fields(config, diagnostic_state, prognostic_state, savepoint_diffusion_exit)
+    if benchmark.disabled:
```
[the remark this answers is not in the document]
kotsaloscv: How about the new iteration?
I think that it addresses all your comments (on which I fully agree).

egparedes (commented):
I like the current approach and have a couple of suggestions to improve it further.

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -110,28 +104,37 @@ class Output:
...
-):
-    if self.MARKERS is not None:
-        apply_markers(self.MARKERS, grid, backend)
+def run_validate_and_benchmark(
+    func: Callable,
+    benchmark_fixture: pytest.FixtureRequest,
+    verification_func: Callable,
+    verify_args: dict[str, Any],
+    *args: Any,
+    **kwargs: Any,
```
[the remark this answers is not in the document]
kotsaloscv: while I like this order, this will not be a valid Python signature.
Any idea how to make it work?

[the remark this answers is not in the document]
kotsaloscv: @egparedes 
How about now?
Once we fix the framework, I will apply it to other granule functions.

egparedes (commented):
Minor style-related comments. Additionally, it would be good to add some tests for the most critical/complex functions here.

kotsaloscv:
> Minor style-related comments. Additionally, it would be good to add some tests for the most critical/complex functions here.

FYI, @egparedes 

Once you are happy with the verification/benchmarking infrastructure, I will add more granule tests (I just wanted to stabilize it first and then simply apply it to other tests -easy part ;) -)

egparedes:
> FYI, @egparedes
> 
> Once you are happy with the verification/benchmarking infrastructure, I will add more granule tests (I just wanted to stabilize it first and then simply apply it to other tests -easy part ;) -)

I'm sorry if my comment about adding tests was misleading, but I meant to add unit tests for the benchmarking infrastructure enhanced in this PR.

kotsaloscv:
> > FYI, @egparedes
> > Once you are happy with the verification/benchmarking infrastructure, I will add more granule tests (I just wanted to stabilize it first and then simply apply it to other tests -easy part ;) -)
> 
> I'm sorry if my comment about adding tests was misleading, but I meant to add unit tests for the benchmarking infrastructure enhanced in this PR.

How about now?

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -143,32 +144,43 @@ def _test_validation(
...
-        if self.MARKERS is not None:
-            apply_markers(self.MARKERS, grid, backend)
+def _test_and_benchmark(
+    self,
+    grid: base.BaseGrid,
+    backend: gtx_backend.Backend,
+    connectivities_as_numpy: dict[str, np.ndarray],
+    input_data: dict[str, gtx.Field],
+    benchmark,  # benchmark fixture
+):
```
egparedes, line 158: Still missing type hints

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+
+def verify_field(
+    field: NDArray[base_type],
+    increment: int,
+    base_value: int,
+):
+    all_correct = np.all(field == (base_value + increment))
+    
+    assert type(all_correct) is np.bool_
+    assert all_correct, f"Field verification failed"
```
egparedes, line 32:
```suggestion
    np.testing.assert_all_close(field == (base_value + increment))
```

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+def test_verification_benchmarking_infrastructure(benchmark):
+    base_value = 1
+    field = base_value*np.ones((1000, 1000), dtype=base_type)
```
egparedes, line 37: No need to use an array, a simple 0d field should be ok for this function
```suggestion
    field = base_value*np.ones((), dtype=base_type)
```

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+        functools.partial(incr_func, field=field, increment=increment),
+        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
+        benchmark,
+    )
+
+    current_base_value = np.random.choice(field.flat)
+    assert current_base_value != base_value, f"Base values should not be equal. Otherwise, the test did not go through incr_func and/or verify_field functions."
+    
+    incr_func(field, increment)
+    verify_field(field, increment, current_base_value)
```
egparedes, line 51: What is this testing?

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+    
+    increment = 6
+    
+    helpers.run_verify_and_benchmark(
+        functools.partial(incr_func, field=field, increment=increment),
+        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
+        benchmark,
+    )
+
+    current_base_value = np.random.choice(field.flat)
```
egparedes, line 47: For 0d fields:
```suggestion
    current_base_value = field[()]
```

egparedes (commented):
More comments.

model/testing/tests/test_verification_benchmarking.py
kotsaloscv: Just to make sure that python passes correctly through the `run_verify_and_benchmark` function and increments the field. Therefore, the second time that we call these functions, the base value is already altered.

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,51 @@
...
+    base_value = 1
+    field = base_value*np.ones((1000, 1000), dtype=base_type)
+    
+    increment = 6
+    
+    helpers.run_verify_and_benchmark(
+        functools.partial(incr_func, field=field, increment=increment),
+        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
+        benchmark,
+    )
```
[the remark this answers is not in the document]
kotsaloscv: I am not sure how to test if verify fails.
In this case, the test simply fails as the verification function hits assertion errors.
Isn't that enough?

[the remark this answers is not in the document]
kotsaloscv: Any idea on the above @egparedes ? Thanks!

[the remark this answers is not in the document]
kotsaloscv: FYI @egparedes 
I now fixed the testing of the non-happy path as well.
Let me know if that's ok.

kotsaloscv:
@halungge do you think that we can add more granule tests for benchmarking, other than the ones that I have already included? Thanks!

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,61 @@
...
+    base_value = 1
+    field = np.array((base_value * np.ones((), dtype=BASE_DTYPE)))
+
+    increment = 6
+
+    helpers.run_verify_and_benchmark(
+        functools.partial(incr_func, field=field, increment=increment),
+        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
+        benchmark_fixture=None,  # no need to benchmark this test
+    )
```
egparedes, line 46: As I said in a previous review, instead of passing the real `benchmark` fixture you should pass a mock (please read the documentation at https://devdocs.io/python~3.13/library/unittest.mock#unittest.mock.Mock) here with the `enabled` attribute set to True/False and then assert if it has been called or not. Something like:

```suggestion
    benchmark = mock.Mock(enabled=False)
    
    helpers.run_verify_and_benchmark(
        functools.partial(incr_func, field=field, increment=increment),
        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
        benchmark_fixture=benchmark,
    )
    assert not benchmark.called

    benchmark.enabled = True
    helpers.run_verify_and_benchmark(
        functools.partial(incr_func, field=field, increment=increment),
        functools.partial(verify_field, field=field, increment=increment, base_value=base_value),
        benchmark_fixture=benchmark,
    )
    assert benchmark.called
```
This is only an example to show how to use it, don't take it as a real suggestion. Try to fit something like that in the test in a way that makes sense and tests all relevant code paths.

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -143,32 +144,43 @@ def _test_validation(
...
-        if self.MARKERS is not None:
-            apply_markers(self.MARKERS, grid, backend)
+def _test_and_benchmark(
+    self,
+    grid: base.BaseGrid,
+    backend: gtx_backend.Backend,
+    connectivities_as_numpy: dict[str, np.ndarray],
+    input_data: dict[str, gtx.Field],
+    benchmark: pytest.FixtureRequest,
+):
```
egparedes, line 158: Missing return type hint

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -199,8 +211,7 @@ def __init_subclass__(cls, **kwargs):
         # reflect the name of the test we do this dynamically here instead of using regular
         # inheritance.
         super().__init_subclass__(**kwargs)
-        setattr(cls, f"test_{cls.__name__}", _test_validation)
-        setattr(cls, f"test_{cls.__name__}_benchmark", _test_execution_benchmark)
+        setattr(cls, f"test_{cls.__name__}", _test_and_benchmark)
 
 
 def reshape(arr: np.ndarray, shape: tuple[int, ...]):
```
egparedes, line 217: Not related to your PR, but since you're touching already this file, maybe you could the proper return type hint here.

egparedes (commented):
Only small things from previous rounds which are still unaddressed.

model/atmosphere/diffusion/tests/diffusion_tests/test_diffusion.py
```diff
@@ -428,11 +430,17 @@ def test_run_diffusion_single_step(
...
     orchestration,
+    benchmark,
 ):
     if orchestration and not helpers.is_dace(backend):
         pytest.skip("Orchestration test requires a dace backend.")
     if orchestration and data_alloc.is_cupy_device(backend):
         pytest.xfail("GPU compilation fails.")
+
+    if experiment == dt_utils.REGIONAL_EXPERIMENT:
+        # Skip benchmarks for this experiment
```
halungge, line 441: why? (This experiment is not "final" in the sense that we a) want to switch to a MCH production setup instead, which has a smaller (coarser) grid (my understanding), but it might be useful to benchmark a LAM setup as well, because there are quite a lot of boundary stencils that only run in this mode. Of course, it can always be added later...

model/testing/tests/test_verification_benchmarking.py
halungge, line 1: I would call this file `test_helpers.py`. It is not always like that in `icon4py` but the tests for something in a module `foo.py` should go into `test_foo.py` for easier overview.

model/testing/tests/test_verification_benchmarking.py
```diff
@@ -0,0 +1,65 @@
...
+
+    benchmark.assert_not_called()
+
+    current_base_value = field[()]
+    assert (
+        current_base_value != base_value
+    ), "Base values should not be equal. Otherwise, the test did not go through incr_func/ verify_field functions."
+
+    # Expect AssertionError
+    with pytest.raises(AssertionError):
```
halungge, line 58: I don't really understand this test. Why should it raise? the function under test `run_verify_and_benchmark` does not raise, does it?  And shouldn't there be a test for the case that the benchmark si called, and one for the case that it isn't?

halungge (changes requested):

model/atmosphere/advection/tests/advection_tests/test_advection.py
```diff
@@ -103,6 +106,11 @@ def test_advection_run_single_step(
         pytest.xfail(
             "This test is skipped until the cause of nonzero horizontal advection if revealed."
         )
+
+    if ntracer == 1:
+        # Skip benchmarks for this experiment
```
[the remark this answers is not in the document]
kotsaloscv: Since we pass the benchmark fixture by default, if we do not set this condition , then it will benchmark the test for multiple parametrizations. By setting this, it benchmarks only the case with 4 tracers. But this is just a suggestion from my side. If want otherwise, please let me know.

model/atmosphere/diffusion/tests/diffusion_tests/test_diffusion.py
kotsaloscv: It is just a suggestion, and a way to show how to skip benchmarking for a specific parameter set. I do not think that we need more than one benchmarking per test. Although, if you think that we need to do otherwise, I can change it.

model/testing/tests/test_verification_benchmarking.py
kotsaloscv: It raises an error because of this:
```
            functools.partial(
                verify_field, field=field, increment=increment, base_value=base_value
            ),  # base_value should be current_base_value
```
i.e. to be correct we need `base_value= current_base_value`.  It is there just to test if `run_verify_and_benchmark` raises errors correctly.

kotsaloscv: Regarding your comment about testing the benchmark branch, we implicitly test it with the mock object `benchmark = mock.Mock(enabled=False)`, i.e. we pass the benchmark fixture (not None), but with the `enabled` parameter to False.

kotsaloscv:
FYI @egparedes @halungge 
is it ok now?
If yes, we can merge it.

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -139,32 +140,43 @@ def _test_validation(
...
-    def _test_execution_benchmark(self, pytestconfig, grid, backend, input_data, benchmark):
-        if self.MARKERS is not None:
-            apply_markers(self.MARKERS, grid, backend)
+def _test_and_benchmark(
+    self,
+    grid: base.BaseGrid,
+    backend: gtx_backend.Backend,
+    connectivities_as_numpy: dict[str, np.ndarray],
+    input_data: dict[str, gtx.Field],
+    benchmark: pytest.FixtureRequest,
```
edopao, line 153: Just for me to understand how this works: where do we pass the `benchmark` fixture?

kotsaloscv: The benchmark fixture is passed to the test and the test passes it to the helper function, just like here:
https://github.com/C2SM/icon4py/blob/963e9b29e821a08efea584c14ee8d1f7bd74b9ca/model/atmosphere/diffusion/tests/diffusion_tests/test_diffusion.py#L517

halungge: In the tests: when a stencil  `StencilTest`s class is created it is passed through the `init_subclass` that registers this function.  For granule you pass it to the test function like in the example in `def test_run_diffusion_single_step` in `test_diffusion.py`. Then at runtime it is captured and provided by pytest.

model/testing/tests/test_helpers.py
halungge, line 1: It is still hard to understand what you want to test here. You can use mocks not only for the benchmark but also for the functions, since all you want check is whether they are called or not. I would do something like that: 
```python
pytest.mark.parametrize("benchmark_enabled", [True, False])
def test_verification_benchmarking_infrastructure(benchmark_enabled):
    base_value = 1
    field = np.array((base_value * np.ones((), dtype=BASE_DTYPE)))

    increment = 6

    benchmark = mock.Mock(enabled=benchmark_enabled)
    test_func = mock.Mock(wraps=incr_func)
    verification_func = mock.Mock(wraps=verify_field)

    helpers.run_verify_and_benchmark(
        functools.partial(test_func, field=field, increment=increment),
        functools.partial(verification_func, field=field, increment=increment, base_value=base_value),
        benchmark_fixture=benchmark,
    )
    test_func.assert_called_once()
    verification_func.assert_called_once()
    benchmark.assert_called() if benchmark_enabled else benchmark.assert_not_called()
```

(using your functions, you could even use empty functions). I think that tests all variants except for the `benchmark = None`, which I don't know whether you can mock that

halungge: Not the current version tests all branches, if it does it does it in a very obfuscated way imho. It took me way too long to understand what you were trying to test....

model/testing/src/icon4py/model/testing/helpers.py
egparedes: @edopao the "benchmark" fixture is injected by pytest at runtime as any other fixture, so there is no particular magic there. The only "magic" is in `StencilTest.__init_subclass__ ` where the `_test_and_benchmark` method definition is injected into every new subclass with a customized name.

model/testing/src/icon4py/model/testing/helpers.py
```diff
@@ -106,28 +101,34 @@ class Output:
...
-        **{k: v.asnumpy() if isinstance(v, gtx.Field) else v for k, v in input_data.items()},
-    )
+    Args:
+        test_func: Function to be ran, verified and benchmarked.
+        verification_func: Function to be used for verification of test_func.
+        benchmark_fixture: pytest-benchmark fixture.
 
-    input_data = allocate_data(backend, input_data)
+    Note:
+        - test_func and verification_func should be provided with binded arguments, i.e. with functools.partial.
```
egparedes, line 118:
```suggestion
        - test_func and verification_func should be provided with bound arguments, i.e. with functools.partial.
```

model/testing/tests/test_helpers.py
```diff
@@ -0,0 +1,45 @@
...
+
+from icon4py.model.testing import helpers
+
+
+def required_args_func(req_arg):
+    pass
+
+
+@pytest.mark.parametrize("benchmark_enabled", [True, False])
+def test_verification_benchmarking_infrastructure(benchmark_enabled):
```
egparedes, line 21:
```suggestion
def test_run_verify_and_benchmark(benchmark_enabled):
```

model/testing/tests/test_helpers.py
```diff
@@ -0,0 +1,45 @@
...
+    else:
+        benchmark.assert_not_called()
+
+    failing_verification_func = mock.Mock(side_effect=AssertionError("Verification failed."))
+    with pytest.raises(AssertionError):
+        helpers.run_verify_and_benchmark(
+            functools.partial(test_func, req_arg=mock.Mock()),
+            failing_verification_func,
+            benchmark_fixture=None,
+        )
```
egparedes, line 45: This is not really testing anything new because we already tested earlier that `verification_func` is always called. 

However there is  still one test case missing: calling the function with `benchmark_fixture = None` which should still call the verification function.

It should be very simple moving this part to a new function with fresh mocks so we can simply assert:
```suggestion
    helpers.run_verify_and_benchmark(
        functools.partial(test_func, req_arg=mock.Mock()),
        verification_func,
        benchmark_fixture=None,
    )

    test_func.assert_called_once()
    verification_func.assert_called_once()
```

egparedes (commented):
Final changes. After fixing these changes it should be fine for me so I don't need another round and I'll leave the final approval to @halungge

model/testing/src/icon4py/model/testing/helpers.py
edopao: @egparedes You are right. What I didn't understand before is that `_test_and_benchmark` is injected with a customized name, and this name adheres to the naming convention for `test_` functions. So when you say that _the fixture is injected by pytest at runtime_, that is because the test function is called by pytest and the `benchmark` fixture is probably passed as a keyword argument.. but I don't see this, therefore it seemed magic to me. I (probably) see now how it all works. Thanks.

model/testing/tests/test_helpers.py
```diff
@@ -36,10 +36,16 @@ def test_verification_benchmarking_infrastructure(benchmark_enabled):
...
 
-    failing_verification_func = mock.Mock(side_effect=AssertionError("Verification failed."))
-    with pytest.raises(AssertionError):
-        helpers.run_verify_and_benchmark(
-            functools.partial(test_func, req_arg=mock.Mock()),
-            failing_verification_func,
-            benchmark_fixture=None,
-        )
+
+def test_run_and_verify():
```
egparedes, line 40:
```suggestion
def test_run_verify_and_benchmark_no_fixture():
```

halungge (approved):
LGTM

egparedes:
@kotsaloscv you forgot to add the PR description before merging. Could you add it now so it would be at least visible following the PR number in the commit log?

kotsaloscv:
> @kotsaloscv you forgot to add the PR description before merging. Could you add it now so it would be at least visible following the PR number in the commit log?

Indeed, my bad ;) Thanks for reminding me.

PR #709, opened 2025-04-04
RBF interpolation stencils improvement

halungge:
Adds computation of RBF interpolation coefficient fields for cells, edges, and vertices. Adds a few missing cartesian coordinate fields used as inputs for the computation. The R04B09 vertex and cell fields are quite far off from the reference fields (~1e-2) but this seems to be inherent to the method. 

This does not port the gradient RBF field.

This is using numpy/cupy for the computation and is mostly vectorized. The cholesky decomposition/solve is always done on CPU, and is not vectorized/parallelized. The RBF matrices are small (n=4-9) and don't offer enough parallelism. I haven't made any attempts at parallelizing this or profiling in general, so this may be suboptimal, but I would leave potential optimizations for a follow-up, if needed.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/icon.py
+++ b/model/common/src/icon4py/model/common/grid/icon.py
@@ -44,14 +44,16 @@ class GlobalGridParams:
     radius: float = constants.EARTH_RADIUS
 
     @classmethod
-    def from_cells(
+    def from_mean_cell_area(
         cls,
-        num_cells: int,
         mean_cell_area: float,
+        root: Optional[int] = None,
+        level: Optional[int] = None,
+        num_cells: Optional[int] = None,
         geometry_type: Final[base.GeometryType] = base.GeometryType.ICOSAHEDRON,
         radius: float = constants.EARTH_RADIUS,
     ):
-        return cls(None, None, num_cells, mean_cell_area, geometry_type)
+        return cls(root, level, num_cells, mean_cell_area, geometry_type)
 
     @functools.cached_property
     def num_cells(self):
--- a/tools/src/icon4py/tools/py2fgen/wrappers/grid_wrapper.py
+++ b/tools/src/icon4py/tools/py2fgen/wrappers/grid_wrapper.py
@@ -138,7 +138,7 @@ def grid_init(
         limited_area=limited_area,
         on_gpu=on_gpu,
     )
-    grid.set_global_params(icon_grid.GlobalGridParams.from_cells(num_cells, mean_cell_area))
+    grid.set_global_params(icon_grid.GlobalGridParams.from_mean_cell_area(mean_cell_area))
 
     # Edge geometry
     edge_params = grid_states.EdgeParams(
--- a/model/common/src/icon4py/model/common/grid/grid_manager.py
+++ b/model/common/src/icon4py/model/common/grid/grid_manager.py
@@ -393,9 +393,9 @@ class GridManager:
         num_edges = self._reader.dimension(gridfile.DimensionName.EDGE_NAME)
         num_vertices = self._reader.dimension(gridfile.DimensionName.VERTEX_NAME)
         uuid = self._reader.attribute(gridfile.MandatoryPropertyName.GRID_UUID)
-        grid_level = self._reader.attribute(gridfile.MandatoryPropertyName.LEVEL)
         grid_root = self._reader.attribute(gridfile.MandatoryPropertyName.ROOT)
-        global_params = icon.GlobalGridParams(level=grid_level, root=grid_root)
+        grid_level = self._reader.attribute(gridfile.MandatoryPropertyName.LEVEL)
+        global_params = icon.GlobalGridParams(root=grid_root, level=grid_level)
         grid_size = base.HorizontalGridSize(
             num_vertices=num_vertices, num_edges=num_edges, num_cells=num_cells
         )
--- a/model/common/tests/grid_tests/test_icon.py
+++ b/model/common/tests/grid_tests/test_icon.py
@@ -216,25 +216,37 @@ def test_when_replace_skip_values_then_only_pentagon_points_remain(grid_file, di
 
 
 @pytest.mark.parametrize(
-    "grid_root,grid_level,num_cells,mean_cell_area,expected",
+    "grid_root,grid_level,num_cells,mean_cell_area,expected_num_cells,expected_mean_cell_area",
     [
-        (2, 4, None, None, 24907282236.708576),
-        (4, 9, None, None, 6080879.45232143),
-        (2, 4, 42, 123.456, 123.456),
-        (4, 9, None, 123.456, 123.456),
+        (2, 4, None, None, 20480, 24907282236.708576),
+        (4, 9, None, None, 83886080, 6080879.45232143),
+        (2, 4, 42, 123.456, 42, 123.456),
+        (4, 9, None, 123.456, 83886080, 123.456),
+        (4, 9, 42, None, 42, 12145265243042.658),
     ],
 )
-def test_mean_cell_area_calculation(grid_root, grid_level, num_cells, mean_cell_area, expected):
+def test_global_grid_params(
+    grid_root, grid_level, num_cells, mean_cell_area, expected_num_cells, expected_mean_cell_area
+):
     params = icon.GlobalGridParams(grid_root, grid_level, num_cells, mean_cell_area)
-    assert expected == params.mean_cell_area
+    assert expected_num_cells == params.num_cells
+    assert expected_mean_cell_area == params.mean_cell_area
 
 
 @pytest.mark.parametrize(
-    "num_cells,mean_cell_area",
+    "grid_root,grid_level,num_cells,mean_cell_area,expected_num_cells,expected_mean_cell_area",
     [
-        (42, 123.456),
+        (2, 4, 42, 123.456, 42, 123.456),
+        (4, 9, None, 123.456, 83886080, 123.456),
+        (0, 0, 42, 123.456, 42, 123.456),
+        (None, None, 42, 123.456, 42, 123.456),
     ],
 )
-def test_mean_cell_area_from_cells(num_cells, mean_cell_area):
-    params = icon.GlobalGridParams.from_cells(num_cells, mean_cell_area)
-    assert mean_cell_area == params.mean_cell_area
+def test_global_grid_params_from_mean_cell_area(
+    grid_root, grid_level, num_cells, mean_cell_area, expected_num_cells, expected_mean_cell_area
+):
+    params = icon.GlobalGridParams.from_mean_cell_area(
+        mean_cell_area, num_cells=num_cells, root=grid_root, level=grid_level
+    )
+    assert expected_num_cells == params.num_cells
+    assert expected_mean_cell_area == params.mean_cell_area
--- a/model/testing/src/icon4py/model/testing/serialbox.py
+++ b/model/testing/src/icon4py/model/testing/serialbox.py
@@ -149,8 +149,8 @@ class IconGridSavepoint(IconSavepoint):
     ):
         super().__init__(sp, ser, size, backend)
         self._grid_id = grid_id
-        self.global_grid_params = icon.GlobalGridParams(
-            root, level, self.num(dims.CellDim), self.mean_cell_area()
+        self.global_grid_params = icon.GlobalGridParams.from_mean_cell_area(
+            self.mean_cell_area(), root=root, level=level
         )
 
     def verts_vertex_lat(self):


model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -464,6 +464,49 @@ def _register_computed_fields(self):
             pairs=(("u_cell_1", "u_cell_2"), ("v_cell_1", "v_cell_2")),
         )
         self.register_provider(tangent_cell_wrapper)
+        cartesian_vertex_centers = factory.FieldOperatorProvider(
+            # TODO: ..._on_vertices?
+            func=math_helpers.geographical_to_cartesian_on_vertex.with_backend(self.backend),
```
msimberg, line 469: Should this be
```suggestion
            func=math_helpers.geographical_to_cartesian_on_vertices.with_backend(self.backend),
```
? The edge and cell functions are called `on_edges` and `on_cells`, respectively.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
launch jenkins spack

msimberg:
cscs-ci run default

model/common/tests/grid_tests/test_geometry.py
```diff
@@ -308,6 +309,28 @@ def test_dual_normal_vert(backend, grid_savepoint, grid_file, experiment):
...
+    x = grid_geometry.get(attrs.EDGE_CENTER_X)
+    y = grid_geometry.get(attrs.EDGE_CENTER_Y)
+    z = grid_geometry.get(attrs.EDGE_CENTER_Z)
+    assert x.ndarray.shape == (grid.num_edges,)
+    assert y.ndarray.shape == (grid.num_edges,)
+    assert z.ndarray.shape == (grid.num_edges,)
+    # those are coordinates on the unit sphere: hence norm should be 1
+    norm = data_alloc.zero_field(grid, dims.EdgeDim, dtype=x.dtype, backend=backend)
+    math_helpers.norm2_on_edges(x, z, y, out=norm, offset_provider={})
+    assert helpers.dallclose(norm.asnumpy(), 1.0)
```
msimberg, line 331: To do: add tests for other added fields:
- vertex centers
- cell centers

model/common/tests/interpolation_tests/test_rbf_interpolation.py
```diff
@@ -0,0 +1,240 @@
...
+    )
+    assert np.max(offset_table) == grid.num_edges - 1
+    v2e = grid.connectivities[dims.V2EDim]
+    # for i in range(offset_table.shape[0]):
+    assert (offset_table == v2e).all()
+
+
+# TODO: make cupy ready
+# TODO: grid_file here only for comparison?
+# TODO: more experiments? at least one regional (with missing neighbors)
```
msimberg, line 75:
```suggestion
```

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+            rbf_vec_coeff_np[j][i, :num_neighbors] = sla.cho_solve(
+                z_diag_np, np.nan_to_num(rhs_np[j][i, :num_neighbors])
+            )
+    rbf_vec_coeff = [array_ns.asarray(x) for x in rbf_vec_coeff_np]
+
+    # Normalize coefficients
+    for j in range(len(u)):
+        rbf_vec_coeff[j] /= array_ns.sum(nxnx[j] * rbf_vec_coeff[j], axis=1)[:, array_ns.newaxis]
+
+    dim2 = gtx.Dimension("What")  # TODO
```
msimberg, line 393: To do.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+    for i in range(z_rbfmat.shape[0]):
+        invalid_neighbors = np.where(rbf_offset[i, :] < 0)[0]
+        num_neighbors = rbf_offset_np.shape[1] - invalid_neighbors.size
+        rbfmat_np = z_rbfmat_np[i, :num_neighbors, :num_neighbors]
+        z_diag_np = sla.cho_factor(rbfmat_np)
+        for j in range(len(u)):
+            rbf_vec_coeff_np[j][i, :num_neighbors] = sla.cho_solve(
+                z_diag_np, np.nan_to_num(rhs_np[j][i, :num_neighbors])
+            )
+    rbf_vec_coeff = [array_ns.asarray(x) for x in rbf_vec_coeff_np]
```
msimberg, line 387: Note, with a GPU backend this part is still done on CPU (with scipy) because cupy doesn't have the `cho_solve` API exposed, though one is proposed in a PR. Unless this becomes a performance bottleneck, I would keep this on CPU for now, and potentially move it over to the GPU if/when cupy gets the same `cho_solve` API. One can compose the functionality of `cho_solve` from other parts in the cupy API, but I don't see it as worth the effort right now.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+
+    # Normalize coefficients
+    for j in range(len(u)):
+        rbf_vec_coeff[j] /= array_ns.sum(nxnx[j] * rbf_vec_coeff[j], axis=1)[:, array_ns.newaxis]
+
+    dim2 = gtx.Dimension("What")  # TODO
+    return tuple(gtx.as_field([dim, dim2], c) for c in rbf_vec_coeff)
```
msimberg, line 394: IMO readability does suffer from mixing numpy/cupy arrays/operations and gt4py fields/programs, but curious to hear how others find this.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+    arc_lengths = _dot_product(v1, v2, array_ns=array_ns)
+    v1_norm = array_ns.linalg.norm(v1, axis=-1)
+    v2_norm = array_ns.linalg.norm(v2, axis=-1)
+    # Divide the dot products by the broadcasted norms
+    arc_lengths = array_ns.divide(arc_lengths, v1_norm[:, :, array_ns.newaxis])
+    arc_lengths = array_ns.divide(arc_lengths, v2_norm[:, array_ns.newaxis, :])
+    # Ensure all points are within [-1.0, 1.0] (may be outside due to numerical
+    # inaccuracies)
+    arc_lengths = array_ns.clip(arc_lengths, -1.0, 1.0)
+    return array_ns.squeeze(array_ns.arccos(arc_lengths), axis=1)
```
msimberg, line 187: To do: rename/clean up.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+    # and the default value is resolution dependent, according to the namelist
+    # documentation in ICON
+    rbf_vector_scale: dict[RBFDimension, ta.wpfloat] = MappingProxyType(
+        {
+            RBFDimension.CELL: 1.0,
+            RBFDimension.EDGE: 1.0,
+            RBFDimension.VERTEX: 1.0,
+            RBFDimension.GRADIENT: 1.0,
+        }
+    )
```
msimberg, line 65: Remove in favour of `compute_rbf_scale`.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
halungge: I agree. I think we mostly do either numpy/cupy or GT4Py computations, only then converting the result, this of course as a drawback sometimes means to either duplicate some functionality or storing some intermediate results/fields that are only used as inputs to some other computations.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+    resol = mean_characteristic_length / 1000.0
+    scale = 0.5 / (1.0 + c1 * math.log(threshold / resol) ** c2) if resol < threshold else 0.5
+    return scale * (resol / 0.125) ** c3 if resol <= 0.125 else scale
+
+
+def construct_rbf_matrix_offsets_tables_for_cells(
+    grid: base_grid.BaseGrid,
+) -> data_alloc.NDArray:
+    """Compute the neighbor tables for the cell RBF matrix: rbf_vec_index_c"""
+    c2e2c = grid.connectivities[dims.C2E2CDim]
```
halungge, line 103: I think this connectivity is available on the ICON grid: `icon_grid.connectivities[dims.C2E2C2EDim]`. There is afunction in `grid_manager.py` called `_construct_triangle_edges` that does this.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
halungge: You could even return  `ndarrays` here, those FieldProviders that we register in the factories do a conversion to fields in the end for numpy/cupy compuations. Such that from the outside you only get `gtx.Fields`

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,520 @@
...
+    edge_center_z,
+    edge_normal_x,
+    edge_normal_y,
+    edge_normal_z,
+    u,
+    v,
+    rbf_offset,
+    rbf_kernel: InterpolationKernel,
+    scale_factor: ta.wpfloat,
+    backend: gtx.backend.Backend | None = None,
```
halungge, line 259: you could use the same pattern as in the other functions, that use numpy, pass it as argument. The argument can then be bound by `functools.partial`, you would then have to return `ndarray` s from the function.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: I updated the whole thing to just use numpy/cupy.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: Not using fields/field operators anymore, just plain numpy/cupy. This part is gone.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: Updated to only use numpy/cupy and thus pass `array_ns` into the rbf function; previously I had to pass the backend in to get access to some gt4py stuff based on the backend but that's gone now.

I haven't registered the field yet so not doing any `functools.partial` stuff yet, looking at that next.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run extra

model/common/src/icon4py/model/common/grid/geometry_attributes.py
```diff
@@ -169,6 +205,30 @@
...
+        standard_name=CELL_CENTER_Y,
+        long_name="y-component of cell center",
+        units="",
+        dims=(dims.CellDim,),
+        icon_var_name="t_grid_cells%%cartesian_center%x(2)",
+        dtype=ta.wpfloat,
+    ),
+    CELL_CENTER_Z: dict(
+        standard_name=CELL_CENTER_Z,
+        long_name="z-component of cell center",
```
msimberg, line 226:
```suggestion
        long_name="y component of cell center",
        units="",
        dims=(dims.CellDim,),
        icon_var_name="t_grid_cells%%cartesian_center%x(2)",
        dtype=ta.wpfloat,
    ),
    CELL_CENTER_Z: dict(
        standard_name=CELL_CENTER_Z,
        long_name="z component of cell center",
```

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
```diff
@@ -25,6 +25,11 @@
 POS_ON_TPLANE_E_X: Final[str] = "pos_on_tplane_e_x"
 POS_ON_TPLANE_E_Y: Final[str] = "pos_on_tplane_e_y"
 CELL_AW_VERTS: Final[str] = "cell_to_vertex_interpolation_factor_by_area_weighting"
+RBF_VEC_COEFF_C1: Final[str] = "rbf_interpolation_coefficient_cell_1"  # TODO: long name?
+RBF_VEC_COEFF_C2: Final[str] = "rbf_interpolation_coefficient_cell_2"  # TODO: long name?
```
msimberg, line 29: I _think_ these could be called (sem for vertices):
```suggestion
RBF_VEC_COEFF_C1: Final[str] = "rbf_interpolation_coefficient_cell_zonal"
RBF_VEC_COEFF_C2: Final[str] = "rbf_interpolation_coefficient_cell_meridional"
```
since the zonal/meridional components of the cell centers are used for the right-hand side of the linear solve. I don't know if `zonal` and `meridional` make sense physically in the names. Perhaps @vectorflux has an opinion on naming?

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
```diff
@@ -131,4 +136,44 @@
         icon_var_name="cells_aw_verts",
         dtype=ta.wpfloat,
     ),
+    RBF_VEC_COEFF_C1: dict(
+        standard_name=RBF_VEC_COEFF_C1,
+        long_name="rbf interpolation coefficient from cells to edges 1",  # TODO: name
```
msimberg, line 141: I suppose this `long_name` has the same questions as in https://github.com/C2SM/icon4py/pull/709/files#r2111186708.

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -52,8 +56,22 @@ def __init__(
         self._attrs = metadata
         self._providers: dict[str, factory.FieldProvider] = {}
         self._geometry = geometry_source
+        mean_characteristic_length = math.sqrt(
+            mean_cell_area(constants.EARTH_RADIUS, grid.global_num_cells)
+        )
```
msimberg, line 61: I had to recreate this computation for here (using an existing helper function, but still...). `CellParams` has a member function `characteristic_length`/`mean_cell_area` with this value already computed. `IconSavepoint` also has a `mean_cell_area` member that I use in the tests. Does it make sense to add a `mean_cell_area` member to `IconGrid` which I could use to access it directly from `self.grid`?

model/common/src/icon4py/model/common/states/prognostic_state.py
```diff
@@ -6,13 +6,12 @@
 # Please, refer to the LICENSE file in the root directory.
```
msimberg, line 6: I can't comment on these changes. @halungge are they useful to keep?

model/common/src/icon4py/model/common/math/helpers.py
```diff
@@ -98,7 +98,7 @@ def _grad_fd_tang(
     return grad_tang_psi_e
 
 
-@gtx.field_operator
+@gtx.field_operator(grid_type=gtx.GridType.UNSTRUCTURED)
```
msimberg, line 101: I ended up adding the `grid_type` while I was using field operators for the RBF computation but ended up removing it in the end. Technically this isn't needed anymore for this PR, but probably belongs here anyway. Keep?

model/common/src/icon4py/model/common/math/helpers.py
```diff
@@ -193,6 +193,19 @@ def dot_product_on_cells(
...
+    x2: fa.VertexField[ta.wpfloat],
+    y1: fa.VertexField[ta.wpfloat],
+    y2: fa.VertexField[ta.wpfloat],
+    z1: fa.VertexField[ta.wpfloat],
+    z2: fa.VertexField[ta.wpfloat],
+) -> fa.VertexField[ta.wpfloat]:
+    """Compute dot product of cartesian vectors (x1, y1, z1) * (x2, y2, z2)"""
+    return x1 * x2 + y1 * y2 + z1 * z2
+
+
```
msimberg, line 208: I ended up adding this operator and the few next ones while I was using field operators for the RBF computation but ended up removing it in the end. Technically this isn't needed anymore for this PR, but probably belongs here anyway. Keep?

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
vectorflux: Whoa, I take it back.  We have to make sure whether the components are really zonal/meridional, or rather tangential/normal.  I suspect they are the latter.

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
vectorflux: Again, it would be valuable to know if 1 is cell-tangential and 2 is cell-normal, or possibly the other way around.

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
```diff
@@ -131,4 +136,44 @@
...
+        units="",
+        dims=(dims.CellDim, dims.KDim),  # TODO KDim?
```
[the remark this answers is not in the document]
vectorflux: It is a 3-D field (over the horizontal and vertical), right. So I think this is correct.

model/common/src/icon4py/model/common/math/helpers.py
vectorflux: Question for Magda.

model/common/src/icon4py/model/common/math/helpers.py
vectorflux: Personally I am against adding code which is not needed, just for the case that maybe some day it might be useful.  This is the reason ICON-F90 is so bloated.

vectorflux (commented):
As is clear from my comments, additional thoughts from Magda would be helpful.

model/common/src/icon4py/model/common/interpolation/interpolation_attributes.py
msimberg: Something like that, yeah. I'll keep the `1`/`2` suffixes, for lack of a (clearly) better alternative, and I don't want to cause extra confusion if zonal/meridional aren't precise enough. I'll remove the TODOs.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -0,0 +1,491 @@
...
+        InterpolationKernel(rbf_kernel),
+        scale_factor,
+        horizontal_start,
+        array_ns=array_ns,
+    )
+    assert len(coeffs) == 2
+    return coeffs
+
+
+def compute_rbf_interpolation_coeffs_edge(
```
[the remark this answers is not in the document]
msimberg: I'll leave it as it is now, until there's proper concensus (if ever).

model/common/src/icon4py/model/common/math/helpers.py
msimberg: I'll leave the grid types as it seemed a bit arbitrary where they were added and where not. I'll remove the added operators though.

model/common/src/icon4py/model/common/math/helpers.py
msimberg: Removed the added operators.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
> I ll have a look. Generally, for the icosahedron, if the grid is properly read from a grid file the global number of cells is correct because it is computed from the R and B parameters which are read from the file that is independent of the distribution. For the serialization data it is not the case, because we do not serialize those but parse them from the filename, which is doubtful. The torus is not implemented at all, there even in the grid file they set the R and B to some arbitrary values (2, 0) which have no meaning afaik instead of hinting at "this is a torus" (which is also directly set in the netcdf attributes, so it is not necessary). These (2,0) values are used somewhere in the driver code for I don't know what but there is even a bug in this.

Thanks for checking. I noticed the (2,0) for the torus and didn't know what to make of it. It's obviously very wrong if you try to compute the global number of cells... based on an icosahedron. As far as I could tell, there's no detection of the torus to set the type to torus in the global parameters and I ended up in the wrong branches with zero global cells. This is why I went in the direction of just getting the mean cell area from the serialized data in that case instead, which is hopefully more reliable.

msimberg:
cscs-ci run default

halungge (commented):
Only some small things, technically, I think there is one test that you should move because you moved the function. And in the wrapper I think you pass the wrong value as the `num_cell` to the `GlobalGridParams`.

All the other things are maybe more philosophical I am not sure my opinion is consistent or clear so feel free to ask if you don't understand my comments. It is a long standing topic.

tools/src/icon4py/tools/py2fgen/wrappers/grid_wrapper.py
```diff
@@ -138,6 +138,7 @@ def grid_init(
         limited_area=limited_area,
         on_gpu=on_gpu,
     )
+    grid.set_global_params(icon_grid.GlobalGridParams.from_cells(num_cells, mean_cell_area))
```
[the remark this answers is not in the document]
msimberg: > Now this is probably wrong: I expect `num_cells` to refer to the number of cells in the local patch, so it would not the right value for a distributed grid nor for a single node LAM model.
> 
> There is a `num_cell` defined on the `IconGrid` itself, that is the number of cells in the local patch, the `num_cell` in the `GlobalGridParams` really is the number of cells in the global grid (before distributed and before a region was cut out for LAM) we should maybe rename the field there to `global_num_cells` or `num_cells_on_global_grid` or something.

Thanks. Yeah, I think that matches my understanding as well and what I was worried about. I think it makes sense to keep the name as `num_cells`, given that it's inside `GlobalGridParams`, so it should be implicit from that. It's also exposed through https://github.com/C2SM/icon4py/blob/4566d532501d7ee9eb33cb99a5919e194112fc22/model/common/src/icon4py/model/common/grid/icon.py#L158-L166.

Do you think it'd be acceptable to make the global number of cells an optional, and return `None` in the cases that we don't know how many there are? In other words:
- pass root + level: can derive global number of cells and mean cell area
- pass mean cell area (e.g. for torus): only return mean cell area, the rest is `None`
- (maybe) pass mean cell area and num cells: unsure if we can guarantee that num cells is the global number?

> It is maybe a bit inconsistent, with what I said above, if this value means the number of cells on the global grid, but the `mean_cell_area` means the "mean cell area of the cells in the patch" or "mean cell area inside the LAM region of the grid used". I am sure about the meaning of the `num_cell` but not about the meaning of the `mean_cell_area`, and I am not sure whether there is consensus about this in the ICON community.

One would hope that these two values are relatively close... but I guess small differences may have a big enough effect.

msimberg:
I pushed some minor changes in https://github.com/C2SM/icon4py/pull/709/commits/3b04eace258ae7191e039d2a4171d004027c6170, including more tests to describe what I'm (attempting) to do in `GlobalGridParams`. The `from_mean_cell_area` constructor does essentially the same as the regular constructor, it just allows you to pass the `mean_cell_area` as the first parameter and the rest are optional (defaulting to `None` if not passed). This signals "prefer the explicit value that I'm passing to you".

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -35,29 +36,72 @@
...
-            case base.GeometryType.ICOSAHEDRON:
-                return compute_icosahedron_num_cells(self.root, self.level)
-            case base.GeometryType.TORUS:
-                return compute_torus_num_cells(1000, 1000)
-            case _:
-                NotImplementedError(f"Unknown gemoetry type {self.geometry_type}")
+        if self._num_cells is None:
+            match self.geometry_type:
+                case base.GeometryType.ICOSAHEDRON:
+                    assert self.root is not None and self.level is not None
```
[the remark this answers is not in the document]
halungge: I think the `root=0` value is an undocumented convention. There is a `grid_geometry` as well which that encodes that. But I am fine with leaving it as is, and we think about it again when we properly support the torus.

tools/src/icon4py/tools/py2fgen/wrappers/grid_wrapper.py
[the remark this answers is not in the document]
halungge: I like you list above. 

in case of the wrapper it is ok to not set the num_cells, because it is only used to compute the `mean_cell_area` which in the wrapper (Fortran integration case) we want to be passed in from Fortran.

halungge:
LGTM, I cannot approve. You will have to approve yourself...

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -35,29 +36,72 @@
...
+
+        return self._num_cells
+
+    @functools.cached_property
+    def characteristic_length(self):
+        return math.sqrt(self.mean_cell_area)
+
+    @functools.cached_property
+    def mean_cell_area(self):
+        if self._mean_cell_area is None:
```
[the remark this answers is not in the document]
msimberg: Yeah, fully agree on computing from basic properties what we can. I hope we can do that more easily when considering the torus properly.

PR #714, opened 2025-04-16
Separate datatests

halungge:
- add pytest option `--datatest-only` which runs tests marked with `datatest` only 
- use this option in the `nox` datatest session
- add a job in `default.yml` that runs the other (`basic`) `nox` session

**additional changes**
- deleted unused functions in `diffusion_utils.py`
- move a `StencilTest` class into `stencil_test` subfolder for test discovery.


halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -161,10 +166,23 @@ def pytest_runtest_setup(item):
...
-        is_datatest=item.config.getoption("--datatest"),
     )
 
 
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--datatest-only"):
+        # --datatest-only given in cli: skip all tests that are not marked with @pytest.mark.datatest
+        for item in items:
+            if not [mark.name for mark in item.own_markers if mark.name == "datatest"]:
+                item.add_marker(
```
egparedes, line 177: Isn't this the same as `-k 'not datatest'`?

egparedes (commented):
It looks good, only have questions, mostly for my own understanding.

model/common/tests/metric_tests/metric_stencil_tests/test_metric_stencils.py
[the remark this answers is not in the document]
halungge: yes, we currently select the stencil tests base on the folder. That is maybe something we should change, we could also select all `StencilTest` implementations or base on some decorator. The reason for the folder structure was that we used it to hide whatever else there is for the spack package of icon4py.

model/testing/src/icon4py/model/testing/datatest_fixtures.py
```diff
@@ -36,8 +36,8 @@ def download_ser_data(request, processor_props, ranked_data_path, experiment, py
     Session scoped fixture which is a prerequisite of all the other fixtures in this file.
```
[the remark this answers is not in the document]
halungge: True. That might have been the case before the we introduced the `experiment` dependency.

model/testing/src/icon4py/model/testing/datatest_fixtures.py
```diff
@@ -36,8 +36,8 @@ def download_ser_data(request, processor_props, ranked_data_path, experiment, py
...
     """
     try:
-        if not request.config.getoption("datatest"):
-            pytest.skip("not running datatest marked test")
+        if not (request.config.getoption("datatest") or request.config.getoption("datatest-only")):
```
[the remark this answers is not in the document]
halungge: They could be true at the same time: if you run with `--datatest-only --datatest` but reduces to running with  `--datatest-only` . The `--datatest` would better be called something like `--with-datatest` imho but I kept it the same for backward compatibiliy.

The fixture body should be skipped if you are not running any datatest. But it is not very nice to do this like that, so if there is a better way....

model/testing/src/icon4py/model/testing/pytest_config.py
halungge: hm, you are right.

halungge:
cscs-ci default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

model/testing/src/icon4py/model/testing/datatest_fixtures.py
```diff
@@ -33,11 +33,12 @@ def download_ser_data(request, processor_props, ranked_data_path, experiment, py
     """
     Get the binary ICON data from a remote server.
 
-    Session scoped fixture which is a prerequisite of all the other fixtures in this file.
+    Fixture which is a prerequisite of all the other fixtures in this file.
     """
     try:
-        if not request.config.getoption("datatest"):
-            pytest.skip("not running datatest marked test")
+        # we don't want to run this ever if we are not running datatests
```
halungge, line 39: Not sure this is correct: My reasoning was: if we run with `--datatest-skip` we set the "not datatest" in the config, and then we don't need to download data, so return and skip the body.

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -78,20 +78,40 @@ def pytest_configure(config):
     if config.getoption("--backend"):
         backend_option = config.getoption("--backend")
         _check_backend_validity(backend_option)
+    if config.getoption("--datatest-only"):
+        k_option = config.getoption("-k")
+        if not k_option:
+            config.option.keyword = "datatest"
+        else:
+            new_option = f"({k_option}) and datatest"
+            config.option.keyword = new_option
```
egparedes, line 87: Optional suggestion:

```suggestion
        if (k_option := config.getoption("-k", [])):
            k_option = [f"({k_option})"]   # add parenthesis around original k_option just in case
        new_option = " and ".join(["datatest"] + k_option)
        config.option.keyword = new_option
```

model/testing/src/icon4py/model/testing/pytest_config.py
```diff
@@ -78,20 +78,40 @@ def pytest_configure(config):
...
+    if config.getoption("--datatest-skip"):
+        k_option = config.getoption("-k")
+        if not k_option:
+            config.option.keyword = "not datatest"
+        else:
+            new_option = f"({k_option}) and not datatest"
+            config.option.keyword = new_option
```
egparedes, line 94: If you like the previous suggestion, the same pattern can be applied here.

model/testing/src/icon4py/model/testing/datatest_fixtures.py
egparedes: I think it makes sense, but wouldn't it be simpler to avoid the `try-except` block and provide a falsy value as default to `request.config.getoption()`:
```python
if "not datatest" in request.config.getoption("-k", ""):
    return
```

egparedes (commented):
It looks good. I only have minor comments about style.

model/testing/src/icon4py/model/testing/datatest_fixtures.py
```diff
@@ -33,11 +33,12 @@ def download_ser_data(request, processor_props, ranked_data_path, experiment, py
     """
     Get the binary ICON data from a remote server.
 
-    Session scoped fixture which is a prerequisite of all the other fixtures in this file.
+    Fixture which is a prerequisite of all the other fixtures in this file.
```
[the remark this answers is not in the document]
halungge: The problem is that it depends on fixtures that are in function scope in particular the `experiment` which is whether we run on global grid or LAM model. This one is not session scoped but parametrized by test function.

halungge:
cscs-ci run default

halungge:
cscs-ci run default

egparedes (approved):
LGTM

PR #716, opened 2025-04-17
remove unused arguments (concat_where)

halungge:
Cleanups in metric fields:
- removes unused arguments from function signatures where `concat_where` was introduced
- delete a duplicated function
- use concat_where in some more places


halungge:
cscs-ci run default

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -774,6 +771,18 @@ def compute_flat_idx(
     )
 
 
+def compute_max_index(
+    flat_idx: data_alloc.NDArray, array_ns: ModuleType = np
```
edopao, line 775: It could be error-prone to set a default `array_ns`, in case one forgets to pass it. I would suggest to not set a default.

edopao (approved):

model/common/src/icon4py/model/common/metrics/metric_fields.py
halungge: True, but we set this default everywhere.

halungge:
cscs-ci run default

halungge:
cscs-ci run default

halungge:
cscs-ci run default

PR #723, opened 2025-04-24
Run benchmarks per subpackage, merge and upload with Bencher

kotsaloscv:


kotsaloscv:
cscs-ci run benchmark-bencher

PR #724, opened 2025-04-25
CI: remove boost dependency

edopao:
With gridtools-cpp 2.3.9 the relevant boost dependencies are bundled, therefore GT4Py does not depend on user-installed boost anymore.

Removes the boost installation from CI.

See the GT4Py PR https://github.com/GridTools/gt4py/pull/1971


edopao:
cscs-ci run default

PR #726, opened 2025-04-25
GitHub runner for unit tests

halungge:


.github/workflows/icon4py-test-tools.yml
```diff
@@ -0,0 +1,43 @@
...
+              run: |
+                sudo apt-get update
+                sudo apt-get install libboost-all-dev
+            # we need to install python with setup-python action and cannot rely on the setup-uv:
+            # we need to link to the python library in the CFFI test in tools and  LD_LIBRARY_PATH
+            # is not set correctly in the setup-uv action
+            #- name: Set up Python
+            #  uses: actions/setup-python@v4
+            #  with:
+            #     python-version: ${{ matrix.python-version }}
```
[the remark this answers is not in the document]
halungge: yes.

halungge:
cscs-ci run default

.github/workflows/icon4py-test-model.yml
```diff
@@ -0,0 +1,71 @@
...
+        types: [opened, reopened, synchronize]
+jobs:
+    test-model:
+        runs-on: ubuntu-latest
+        strategy:
+            matrix:
+                python-version: ["3.10", "3.11"]
+                backend: ["embedded", "gtfn_cpu"] # TODO add dace-cpu?
+                # TODO: advection tests are skipped, re-enable them by adding 'advection' to the 'COMPONENT' list.
+                #  The problem is that GTFN backend times out on the advection stencil tests.
```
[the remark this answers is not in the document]
edopao: I think the problem was for both backends. Anyway, the advection tests will be re-enabled in #738 once we use a newer gt4py version.

.github/validation_grids.py
```diff
@@ -0,0 +1,45 @@
+import enum
```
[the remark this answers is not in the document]
halungge: nice

tools/tests/py2fgen/test_cli.py
```diff
@@ -83,7 +84,8 @@ def run_test_case(
 
 
 def invoke_cli(cli, module, function, library_name):
-    cli_args = [module, function, library_name]
+    rpath = f"{sys.base_prefix}/lib"
```
havogt, line 87: I am wondering if there is a scenario where you want to use a different libpython at runtime than the one that was used when executing py2fgen. If not, we could unconditionally set the `rpath` and remove the cli parameter.

halungge: This is what in fact seems to happen when we don't set it: My system python is 3.10. I could run 
` uv run nox -s "test_tools-3.10(unittest)" ` but `uv run nox -s "test_tools-3.11(unittest)" ` failed when not passing the rpath because it linked the system library. I think for that example that is _not_ what you want because it is undefined. But I also don't know for other scenarios.

.github/validation_grid_utils.py
```diff
@@ -0,0 +1,70 @@
...
+        return 0
+    except ValueError as e:
+        message, exit_code = e.args
+        print(f"[ERROR] {message}")
+        return exit_code
+
+
+
+if __name__=="__main__":
+    sys.exit(app())
```
egparedes, line 70: Now that I learned a bit more about Typer, I would just use 2 different app commands:
```suggestion

app = typer.Typer()

@app.command(name="cache-key")
def cache_key() -> None:
    """Generate a cache key for the GitHub cache action based on the download URI and the grid name."""
    d = "_".join(grid.name + grid.uri for grid in VALIDATION_GRIDS)
    hexdigest = hashlib.md5(d.encode()).hexdigest()
    print(hexdigest)

@app.command
def download() -> None:
    """Effectively download the grid files."""
    for grid in VALIDATION_GRIDS:
        print(f"Downloading and unpacking {grid.name}...")
        fname = grid_utils._download_grid_file(grid.name)
        print(f"done - downloaded {fname}")

if __name__=="__main__":
    app()
```

egparedes (commented):
I think the typer script can be simplified. Other than that, LGTM

.github/validation_grid_utils.py
halungge: I used the @app.command in a a first try, but I really don't remember now why I reverted back to the option.  Nothing

.github/validation_grid_utils.py
```diff
@@ -0,0 +1,43 @@
...
+import typer
+import sys
+
+from icon4py.model.testing import grid_utils
+from icon4py.model.testing import definitions
+import hashlib
+
+VALIDATION_GRIDS = (definitions.Grids.R02B04_GLOBAL, definitions.Grids.MCH_CH_R04B09_DSL, definitions.Grids.MCH_OPR_R04B07_DOMAIN01) # change to MCH_OPR_R04B07_DOMAIN01
+app = typer.Typer()
+@app.command(name="cache-key")
```
egparedes, line 23: It seems that pre-commit / ruff-format is not processing this file. Do you know why? Could you add it?

halungge:
cscs-ci run default

.github/workflows/icon4py-test-tools.yml
```diff
@@ -0,0 +1,36 @@
...
+                  version: "0.6.16"
+                  enable-cache: true
+                  cache-dependency-glob: "uv.lock"
+                  python-version: ${{ matrix.python-version }}
+            - name: run nox session - tools
+              env:
+                NUM_PROCESSES: auto
+              shell: bash
+              run: |
+                uv run nox -s "test_tools-${{matrix.python-version}}(unittest)" --
```
egparedes, line 36: Same as above.

egparedes (approved):
Two final comments but I won't need another round of review because they are not important. Approving the PR now.

.github/workflows/icon4py-test-model.yml
```diff
@@ -0,0 +1,65 @@
...
+                NUM_PROCESSES: auto
+              run: |
+                uv run nox -s "test_model-${{matrix.python-version}}(basic, ${{ matrix.component }})" -- --backend=${{ matrix.backend }}
+
+            - name: run nox session - stencil tests
+              shell: bash
+              env:
+                  NUM_PROCESSES: auto
+              run: |
+                uv run nox -s "test_model-${{matrix.python-version}}(stencils, ${{ matrix.component }})" -- --backend=${{ matrix.backend }} --grid=icon_grid
```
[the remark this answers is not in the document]
halungge: makes lot of sense..

.github/validation_grid_utils.py
halungge: I guess it is this in the `.pre-commit-config.yaml`:
```
minimum_pre_commit_version: 3.8.0
files: |
  (?x)^(
      model/.*|
      tools/.*
  )$

```

.pre-commit-config.yaml
```diff
@@ -9,7 +9,8 @@ minimum_pre_commit_version: 3.8.0
 files: |
   (?x)^(
       model/.*|
-      tools/.*
+      tools/.*|
+      .github/.*.py
   )$
```
egparedes, line 14: Hmm, I haven't realized about this, but why do we have a `files:` filter here? Shouldn't every hook have a specific filter, if needed, and otherwise apply hooks to the whole repo? That's what we do in gt4py at least...

halungge: I have no idea, why it is like that. Probably for no real reason. I suggest, that we leave it like this here and adress it in a different PR.

.github/validation_grid_utils.py
```diff
@@ -0,0 +1,58 @@
...
+# Copyright (c) 2022-2024, ETH Zurich and MeteoSwiss
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+# ruff: noqa: ERA001
+
+# /// script
+# requires-python = ">=3.10"
```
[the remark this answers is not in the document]
halungge: That does not seem to help. It does not complain about the entire comment it really does not like the lines 
```
# dependencies = [
# "typer>=0.10",
# "icon4py-testing",
# ]

```
If I put noqa locally there I fall into the trap that the get removed as unnecessary and then in the next run the ERA001 shows up again..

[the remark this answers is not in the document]
halungge: It might be a ruff version thingy, maybe updating ruff might fix it. I will check...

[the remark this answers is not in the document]
egparedes: Let's merge it now with the noqa and address this later after updating hooks. I would like to implement the same approach used in the latest gt4py (https://github.com/GridTools/gt4py/pull/2086)

PR #728, opened 2025-04-28
don't merge: hack skip values

havogt:


model/common/src/icon4py/model/common/grid/base.py
```diff
@@ -132,7 +133,24 @@ def offset_providers(self):
         for key, value in self.offset_provider_mapping.items():
             try:
                 method, *args = value
-                offset_providers[key] = method(*args) if args else method()
+                offset_provider = method(*args) if args else method()
+                if gtx_common.is_neighbor_connectivity(offset_provider):
+                    offset_provider_type = offset_provider.__gt_type__()
+                    if True: #not offset_provider_type.has_skip_values:
```
havogt, line 139: I would expect that removing `-1`s for fields that don't have skip_values set would be enough, because all that have skip_values should be protected by a `can_deref`.

halungge: According to Till the can_deref only protects the neighbor sums, if there is a direct access like CONN[0] this does not help.

PR #729, opened 2025-04-28
Make vn incr optional

havogt:


jcanton:
there are other locations in the dycore where vn_incr (and others) should be optional, e.g.:
https://github.com/C2SM/icon4py/blob/008015f4a63af9740eeebd16458caa84339db032/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py#L988C1-L988C87

see also: https://github.com/C2SM/icon4py/pull/739

PR #743, opened 2025-05-13
Run pre-commit with `--all-files` flag in CI

msimberg:
`pre-commit` was not doing any checks in CI, because it was run without `--all-files`, which normally means running on all changed files about to be committed. Since there are no staged changes in the CI run, no files were checked.

This silences mypy type warnings and formatting errors that snuck in while the checks were disabled. This also updates the actions used in the QA workflow and uses `setup-python`, `setup-uv`, and `pre-commit` actions instead of manually setting up the tools. The `.python-version` file is no longer ignored in `.gitignore` as it's actually checked into the repo and used by CI and environments.


msimberg:
mypy checks failing. Will see if I can disable only that check in CI.

.github/workflows/icon4py-qa.yml
```diff
@@ -33,4 +33,4 @@ jobs:
             - name: Run checks
               run: |
                   source .venv/bin/activate
-                  pre-commit run
+                  SKIP=mypy pre-commit run --all-files
```
halungge, line 36: There are two packages where we fixed `mypy` (see `.pre-commit-config.yml`): the stencil tests in the dycore and tools. So rather than globally skipping mypy it would be good to run it there otherwise those will rot again.

(In fact they already do, because I get some errors for the `tools` package. I suggest to do the following:
remove the skip here: 
```suggestion
                  pre-commit run --all-files
```
and fix the issues in the `tools` package. 
You can fix all of them with a `# type: ignore[<rule>]` comment. 

(All except for one are in "deprecated" code that will be deleted eventually because it is part of Fortran to GT4Py integration methods that we do not use any longer. The one that is not, already has such a comment but the wrong rule, if I see that correctly.

See patch file: 
[mypy_fixes_in_tools.txt](https://github.com/user-attachments/files/20202842/mypy_fixes_in_tools.txt)

@egparedes does that make sense?

halungge (commented):
Thanks. I suggest to remove the `mypy` skip (see comment below).

.github/workflows/icon4py-qa.yml
msimberg: Thanks @halungge, that's definitely better! I applied your diff and:

- ignored warnings locally for `icon4py.model.atmosphere.dycore.dycore_states` since there were only three of them (an empty `py.typed` file also works for this, is that nicer?)
- added the empty `py.typed` file for testing modules

With this it all the checks pass for me locally. I hope CI will agree...

muellch:
I am not the best to understand what the `pre-commit` changes mean.
I asked Enrique to take a look at the PR. Thanks again.
I will exit myself as a reviewer.

.github/workflows/icon4py-qa.yml
egparedes, line 1: Since we are touching this github workflow, I'd use this opportunity to update it to the same style as gt4py (
https://github.com/GridTools/gt4py/blob/main/.github/workflows/code-quality.yml)
using the uv and pre-commit github actions (note that the pre-commit action runs the correct `--all-files` by default).

model/atmosphere/dycore/tests/dycore_stencil_tests/test_apply_divergence_damping_and_update_vn.py
```diff
@@ -12,7 +12,9 @@
 
 import icon4py.model.common.type_alias as ta
 import icon4py.model.testing.helpers as test_helpers
-from icon4py.model.atmosphere.dycore.dycore_states import DivergenceDampingOrder
+from icon4py.model.atmosphere.dycore.dycore_states import (  # type: ignore[import-untyped]
```
egparedes, line 15: Why is this needed? Shouldn't all `icon4py` subpackages be typed (even if they are incomplete or partially wrong for now)? Do all subpackages contain the compulsory `py.typed` file?

This comment applies to all `ignore[import-untyped]` comments from icon4py in this PR

tools/src/icon4py/tools/py2fgen/_definitions.py
```diff
@@ -94,7 +94,7 @@ class ScalarParamDescriptor(eve.Node):
     ArrayInfo: TypeAlias = tuple[Any, tuple[int, ...], bool, bool]
 
 if TYPE_CHECKING:
-    import cupy as cp  # type: ignore[import-untyped]
+    import cupy as cp  # type: ignore[import-not-found]
```
egparedes, line 97: The new type ignore kind of makes more sense than the previous one, I would explicitly explain why. IIUC the reason for this might be:
```suggestion
    import cupy as cp  # type: ignore[import-not-found]  # cupy might not be always available at type-checking 
```

egparedes (commented):
Sorry for the late review. Here are my suggestions.

.github/workflows/icon4py-qa.yml
egparedes: I also forgot to mention that `boost` is probably not longer needed  since we are using a modern version of gt4py which requires a `gridtools-cpp` where boost has been dropped (thanks to @msimberg :stuck_out_tongue_winking_eye: ).

And, since we now have per-commit versions in gt4py, we should start updating the minimum required version of gt4py when we start using new features. Ideally, though, we move first gt4py to EffVer and release  a new minor or patch version...

msimberg: Yeah, removing boost was another thing I noticed could be done, but I struggled to find anywhere a mention of which version of gridtools-cpp was being used... I'd leave this for a separate PR. It looks like `libboost-all-dev` may pull in other libraries transitively that may or may not be needed?

model/atmosphere/dycore/tests/dycore_stencil_tests/test_apply_divergence_damping_and_update_vn.py
msimberg: > Why is this needed?

I wish I knew.

> Shouldn't all icon4py subpackages be typed (even if they are incomplete or partially wrong for now)?

No idea :stuck_out_tongue: 

> Do all subpackages contain the compulsory py.typed file?
I'll check.

.github/workflows/icon4py-qa.yml
egparedes: After talking to @halungge we realized that boost is probably needed by serialbox anyway, so it's better to keep it for now in the repo, although it's definitely not needed for running pre-commit.

model/atmosphere/dycore/tests/dycore_stencil_tests/test_apply_divergence_damping_and_update_vn.py
egparedes: I've quickly checked and it seems that there are other packages besides `icon4py.model.testing` (which is being fixed in this PR) missing the `py.typed` file:

- icon4py.model.atmosphere.dycore
- icon4py.model.atmosphere.subgrid_scale_physics.microphysics
- icon4py.model.atmosphere.driver

Could you also add the `py.typed` files into their `src` folders and then removed the `type-ignore`?

msimberg: https://github.com/C2SM/icon4py/pull/743/commits/33b4ec8500dcccd4273808f7e2851d6ab62badaa

I added one to `model/driver/src/icon4py/model/driver`. Is that what you meant for icon4py.model.atmosphere.driver?

.github/workflows/icon4py-qa.yml
msimberg: > Since we are touching this github workflow, I'd use this opportunity to update it to the same style as gt4py (
https://github.com/GridTools/gt4py/blob/main/.github/workflows/code-quality.yml)
using the uv and pre-commit github actions (note that the pre-commit action runs with the correct --all-files flag by default).

With the latest commits the workflow now uses the `setup-python`, `setup-uv`, and `pre-commit` actions. I still don't quite understand how the gt4py workflow doesn't need to explicitly install the environment to get access to mypy; if someone else knows please let me know. In any case, installing the dependencies explicitly with `uv sync --extra all` works, and also seems to be what the `setup-uv` action should require (https://github.com/astral-sh/setup-uv?tab=readme-ov-file#does-setup-uv-also-install-my-project-or-its-dependencies-automatically).

Note that I also bumped the actions versions at the same time.

msimberg:
The test reminder workflow probably fails because the PR is from my fork? Do you mostly not work with forks or is that supposed to work?

model/atmosphere/dycore/tests/dycore_stencil_tests/test_apply_divergence_damping_and_update_vn.py
egparedes: Oops, yes, it was a typo from my side.

.default-python-version
egparedes, line 1: This is not needed since there is `.python-version` in the repo. GT4Py uses a slightly more advanced configuration, we could adopt it here also, but it'd involved more changes and I wouldn't do it in this PR

.github/workflows/icon4py-qa.yml
```diff
@@ -13,24 +13,27 @@ jobs:
...
-                    **/pyproject.toml
-                    uv.lock
-            - name: Install venv and icon4py namespace packages
-              run: |
-                  uv sync --extra all
-            - name: Run checks
-              run: |
-                  source .venv/bin/activate
-                  pre-commit run
+                python-version-file: ".default-python-version"
```
egparedes, line 26: As mentioned above, let's use the existing `.python-version` for now:
```suggestion
                python-version-file: ".python-version"
```

.github/workflows/icon4py-qa.yml
egparedes: > I still don't quite understand how the gt4py workflow doesn't need to explicitly install the environment to get access to mypy; if someone else knows please let me know

The answer is here: https://github.com/GridTools/gt4py/blob/main/.pre-commit-config.yaml#L80
pre-commit uses `uv` to run `mypy` which will create the required project environment before running `mypy`. Note that `uv run` runs commands or scripts inside a project environment, which is different that `uv tool run` (AKA `uvx`, like `pipx`) which runs a command from a python distribution package by creating a throw-away venv.

egparedes (commented):
Just one more change and it should be ready to go.

egparedes:
> The test reminder workflow probably fails because the PR is from my fork? Do you mostly not work with forks or is that supposed to work?

In the past `icon4py` was a closed repo and forks were never used. Now it's open and we should move to forks, but I guess there is a tendency for many people to keep working directly on this repo.

.github/workflows/icon4py-qa.yml
msimberg: Ah, thanks, I completely missed that file! It wasn't showing up in my file searches because it's in `.gitignore`. Is the idea that it's common to change the version locally but that change should not be propagated into the repo, so it's better to have it in `.gitignore` to avoid accidental checkins?

.github/workflows/icon4py-qa.yml
msimberg: Aha, thanks! TIL.

msimberg:
Thanks for the reviews @egparedes and @halungge. This should be ready to go now.

.github/workflows/icon4py-qa.yml
egparedes: Hmm, you're right. I had missed completely that `.python-version` is in the `.gitignore`, which doesn't make sense. I propose to rename it then to `.default-python-version`, check that everything still works and then merge.

egparedes (approved):
LGTM

halungge:
If it is just the workflow thingy, we will force merge.

msimberg:
mypy checks failed again after merging. I'll try to fix them today.

msimberg:
@nfarabullini @OngChia could you just confirm that the changes in https://github.com/C2SM/icon4py/pull/743/commits/5b304669495bf8254cc0a0e4f37166b5aee6329a and https://github.com/C2SM/icon4py/pull/743/commits/d7041175a51e59947e4b336bae2ae36a7e13e3f7 look fine to you?

msimberg:
cscs-ci run default

nfarabullini (approved):

PR #750, opened 2025-05-16
fix divdamp_type == 32

halungge:
(FIX) fixes for `DivergenceDampingType.COMBINED` (32):
 - rename backend
 - use only first element in return tuple

**additional changes**:
use Enum types instead of ints in configuration


huppd (approved):
Thank you for this fix. It problems a problem in the blue line.

huppd:
cscs-ci run default

PR #753, opened 2025-05-20
Remove vertical nesting check

iomaganaris:
For vertical nesting there is a variable `jk_start` set. However since we don't use vertical nesting we can remove anything related to this variable to simplify the code


iomaganaris:
cscs-ci run default

OngChia (approved):

muellch (approved):
Looks good! Thanks.

iomaganaris:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -497,14 +495,6 @@ def __init__(
         )
         self._allocate_local_fields()
         self._determine_local_domains()
-        # TODO (magdalena) vertical nesting is only relevant in the context of
-        #      horizontal nesting, since we don't support this we should remove this option
-        self.l_vert_nested: bool = False
```
iomaganaris, line 502: @muellch It seems like `l_vert_nested` is passed through the `py2f` wrappers. Should we keep this in the signatures for backward compatibility or shall I try to remove this from everywhere?

iomaganaris:
cscs-ci run default

iomaganaris:
cscs-ci run default

iomaganaris:
cscs-ci run default

iomaganaris:
I've fixed some issues with the code and made some small change in the tests to pass because of small differences near 0. Would you like to have a second look @OngChia @muellch ? Thank you 🙏

OngChia:
I think the stencil test passed not because of the random fields that you use for those output variables. It just so happens that no small values are produced in the second CI run. It is anyway fine for me to set the initial values of those output variables to random numbers.
In addition, there is no division at all in `compute_derived_horizontal_winds_and_ke_and_horizontal_advection_of_w_and_contravariant_correction`. So, it is different from my case where there is a division by zero or very small number.

iomaganaris:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
muellch: No we should remove it everywhere. We do not care about vertical nesting.
If you need help removing it on the icon-exclaim side, I can help.

iomaganaris: After discussion with Christoph we decided to keep it in the wrapper because it needs to be passed by the icon Fortran world. If it's passed as `True` then `NonHydrostaticConfig` returns an error because it's not supported by `icon4py`

muellch: Perfect.

muellch (approved):
Nice.
Finally it is gone. :)

PR #756, opened 2025-05-21
Fix lowering issue of DaCe with `n_lev - 1`

iomaganaris:
Currently there is a restriction in DaCe lowering/optimization passes when there is an expression of scalar values inside a `concat_where` like
```
n_lev - 1
```

This PR avoids this issue by passing the value of `n_lev - 1` as argument of the gt4py field operator.
Fixes the issue that https://github.com/GridTools/gt4py/pull/2036 tries to solve IIUC.


iomaganaris:
cscs-ci run default

iomaganaris:
cscs-ci run dace

edopao (commented):
I would prefer to only apply the fix in gt4py. In my opinion, this PR makes the icon4py code more difficult to read.

PR #762, opened 2025-05-27
Another attempt at 1 to 13

muellch:
Re-writing the stencil 1_to_13 to better suit dace optimization pipeline, while keeping it as readable as possible for scientists.
Understanding all of the changes in this PR requires good understanding of the ICON Fortran code.


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -52,22 +52,38 @@
...
+)
 from icon4py.model.common.math.derivative import _compute_first_vertical_derivative_at_cells
 from icon4py.model.common.type_alias import vpfloat, wpfloat
 
 
 horzpres_discr_type: Final = HorizontalPressureDiscretizationType()
 
 
+@field_operator
+def _initialize_zero_if_limited_area(
```
OngChia, line 66: Do we actually need this to set these two variables to zero from `(start_cell_lateral_boundary, start_cell_lateral_boundary_level_3)`? They are initialized to be zero fields when instantiating the granule, and the cells in `(start_cell_lateral_boundary, start_cell_lateral_boundary_level_3)` is never recomputed?

model/atmosphere/dycore/tests/dycore_stencil_tests/test_compute_perturbed_quantities_and_interpolation.py
```diff
@@ -450,11 +448,8 @@ def input_data(self, grid: base.BaseGrid) -> dict[str, gtx.Field | state_utils.S
             start_cell_lateral_boundary=start_cell_lateral_boundary,
             start_cell_lateral_boundary_level_3=start_cell_lateral_boundary_level_3,
             start_cell_halo_level_2=start_cell_halo_level_2,
-            end_cell_end=end_cell_end,
             end_cell_halo=end_cell_halo,
             end_cell_halo_level_2=end_cell_halo_level_2,
-            horizontal_start=0,
```
OngChia, line 456: I quite like the removal of `horizontal_start`, `horizontal_end`, and if possible `vertical_start` and `vertical_end`. Otherwise, it is very confusing with the combination of many starting/ending index names with `horizontal_start` and `horizontal_end`. I would suggest we also rename `vertical_start` and `vertical_end` to `model_top_index` and `surface_level_index` or `num_levels_plus_one` (you can think of a nicer name), respectively. And maybe we should adopt it in other stencils as well.

(just putting a reminder here that the docstring also needs to be edited whenever other things are finished)

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -523,6 +483,21 @@ def compute_perturbed_quantities_and_interpolation(
         },
     )
 
+    _compute_perturbation_of_rho_and_theta(
```
OngChia, line 486: Small point. Maybe I would prefer to put this field operator call directly right after the field operator `_compute_perturbed_quantities_and_interpolation` for better readability instead of here.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
havogt: This is an old comment and we discussed at some point. Just want to repeat here for others to see: for granule local temporaries, we want to be able to allocate and free the fields to decrease memory footprint when the granule is inactive (or even within the granule). Therefore we should not assume initialization of the fields. See also https://github.com/C2SM/icon4py/pull/811

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -52,22 +52,38 @@
...
+    limited_area: bool,
+) -> tuple[
+    fa.CellKField[ta.vpfloat],
+    fa.CellKField[ta.vpfloat],
+]:
+    (perturbed_rho_at_cells_on_model_levels, perturbed_theta_v_at_cells_on_model_levels) = (
+        _init_two_cell_kdim_fields_with_zero_vp()
+        if limited_area
+        else _init_two_cell_kdim_fields_with_zero_vp()
+    )
```
havogt, line 76: ... but what is this actually doing?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -127,63 +121,44 @@ def _compute_perturbed_quantities_and_interpolation(
     (
         perturbed_rho_at_cells_on_model_levels,
         perturbed_theta_v_at_cells_on_model_levels,
+    ) = _compute_perturbation_of_rho_and_theta(
```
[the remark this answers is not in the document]
muellch: no, I do not think it should be split.
What happens in 1 to 13 is that ro and theta_v are the two tracers that are advected.
If we had higher level datastructures, we could put the two of them into the same container and then advect them together.
I would say the best version of this code takes as an input a variable called `rhotheta` and perturbs it by perturbing both components.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
muellch: The renaming can be done, no problem. The position I would rather not change, I intentionally put it at the very end.
Right now the program looks like this.

```
field_operator (start_lateral_boundary, start_lateral_boundary_3),
field_operator (start_lateral_boundary_3, end_halo_line_1),
field_operator (start_lateral_boundary_3, end_halo_line_1),
field_operator (start_lateral_boundary_3, end_halo_line_1),
field_operator (start_lateral_boundary_3, end_halo_line_1),

OngChia (approved):
LGTM. 
Nice, with much better readability to me!

muellch:
cscs-ci run default

muellch:
cscs-ci run default

PR #763, opened 2025-05-28
Fix concat_where range syntax

edopao:
The latest gt4py will not allow the range syntax `start <= dim < end`, instead we should use `(start <= dim) & (dim  < end)`.


edopao:
cscs-ci run default

PR #767, opened 2025-05-29
ci: use all available CUDA devices for parallel tests

egparedes:
Enhance pytest and CSCS-CI configuration settings to use all CUDA devices during parallel tests runs.
It works by adding code in the `pytest_configure()` hook, which is executed by every pytest worker, to set the environment variable `CUDA_VISIBLE_DEVICES` to a different device for each worker id. The list of available devices needs to be explicitly defined in the custom `PYTEST_XDIST_SPLIT_CUDA_VISIBLE_DEVICES` environment variable as a comma separated list.

Additionally, replace the custom `NUM_PROCESSES` env variable by the standard `PYTEST_XDIST_AUTO_NUM_WORKER` to control the number or pytest-xdist workers.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -87,6 +87,8 @@ build_image_aarch64:
     # Grace-Hopper gpu architecture is not enabled by default in CUDA build
     CUDAARCHS: "90"
     # Limit test parallelism to avoid "OSError: too many open files" in the gt4py build stage.
-    PYTEST_XDIST_AUTO_NUM_WORKERS: 16
+    PYTEST_XDIST_AUTO_NUM_WORKERS: 16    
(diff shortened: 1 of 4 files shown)


egparedes:
cscs-ci run default

egparedes:
cscs-ci run default

egparedes:
cscs-ci run default

egparedes:
cscs-ci run default

msimberg:
@egparedes I'll close this tentatively, since we're now running most tests on the shared partition on a single GPU, so we sort of get "use all available GPUs" through separate slurm jobs. Reopen if you think it's worth revisiting.

PR #791, opened 2025-07-02
Restrict connectivities to actual sizes of the dimension

havogt:
In Fortran, all connectivities have `nproma` size.
Here we are restricting the connectivity to the actual size as this is currently used for sizing GT4Py temporaries.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/base.py
+++ b/model/common/src/icon4py/model/common/grid/base.py
@@ -185,11 +185,13 @@ class BaseGrid(ABC):
             neighbor_table = self._neighbor_tables[dim]
 
         if (dimension_size := self.size[from_dim]) <= neighbor_table.shape[0]:
-            # TODO(havogt): Introduce a layer that isolates ICON Fortran related hacks. 
+            # TODO(havogt): Introduce a layer that isolates ICON Fortran related hacks.
             # In Fortran, all connectivities have `nproma` size.
             # Here we are restricting the connectivity to the actual size as this is currently used
             # for sizing GT4Py temporaries.
-            _log.info(f"Restricting connectivity for {dim} to size of {from_dim} ({dimension_size}).")
+            _log.info(
+                f"Restricting connectivity for {dim} to size of {from_dim} ({dimension_size})."
+            )
             neighbor_table = neighbor_table[:dimension_size, :]
 
         connectivity = gtx.as_connectivity(
@@ -236,6 +238,7 @@ class BaseGrid(ABC):
             self._neighbor_tables[dim].shape,
             from_dim,
             to_dim,
+            self.size[from_dim],
             has_skip_values=False,
             array_ns=xp,
         )
--- a/model/common/src/icon4py/model/common/grid/utils.py
+++ b/model/common/src/icon4py/model/common/grid/utils.py
@@ -5,10 +5,14 @@
 #
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
+import logging
 from types import ModuleType
 
 import gt4py.next as gtx
 import numpy as np
+
+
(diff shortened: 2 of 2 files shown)


model/common/src/icon4py/model/common/grid/base.py
```diff
@@ -183,6 +183,14 @@ def _construct_connectivity(self, dim, from_dim, to_dim):
             )
         else:
             neighbor_table = self._neighbor_tables[dim]
+
+        if (dimension_size := self.size[to_dim]) <= neighbor_table.shape[0]:
```
havogt, line 187: please spend a minute in verifying that `to_dim` is correct. (but I guess tests would catch it anyway)

havogt:
cscs-ci run default

model/common/src/icon4py/model/common/grid/base.py
halungge: Is it not the other way round? for C2E: `neighbor_table.shape[0] == num_cells` and we use:
```
 "C2E": (self._construct_connectivity, dims.C2EDim, dims.CellDim, dims.EdgeDim),
``` 
so `CellDim == from_dim`
The naming reflects the mapping of the indices not the fields.

model/common/src/icon4py/model/common/grid/base.py
```diff
@@ -183,6 +183,14 @@ def _construct_connectivity(self, dim, from_dim, to_dim):
...
+            # In Fortran, all connectivities have `nproma` size.
+            # Here we are restricting the connectivity to the actual size as this is currently used
+            # for sizing GT4Py temporaries.
+            _log.info(f"Restricting connectivity for {dim} to size of {to_dim} ({dimension_size}).")
+            neighbor_table = neighbor_table[:dimension_size, :]
```
halungge, line 192: For me is fine to do it unconditionally, the if is never reached in greenline. What I mostly dislike is the comment, because I really don't want to see `nproma` mentioned anywhere in icon4py... :-/

model/common/src/icon4py/model/common/grid/base.py
halungge: tests seem to say to support that you need to use `from_dim`
```suggestion
        if (dimension_size := self.size[from_dim]) <= neighbor_table.shape[0]:
```
Other thought, could it cause problems with the 1d connectivities?

havogt:
> > This doesn't work, but I am not yet sure why. Need to think about the 1d sparse fields as @halungge suggested.
> 
> Can it simply be a lazy evaluation issue? we only register the `_construct_connectivity` or `_get_connectivity_sparse_fields` functions for each dimension at setup and they get evaluated the first time the connectivity is requested. Both of these functions operate on `self._neighbor_tables` but only `_construct_connectivity` reduces the size of the neighbor table. So depending on whether I call `grid.get_connectivity("E2C")` before or after ` grid.get_connectivity("E2EC")` the resulting sparse table looks different. (has the original size or the reduced size)
> 
> if this is the issue: something like that might help:
> 
> ```python
>      def _get_connectivity_sparse_fields(self, dim, from_dim, to_dim):
>         if dim not in self._neighbor_tables:
>             raise MissingConnectivity(f"No neighbor table for dimension {dim}.")
>         xp = data_alloc.array_ns(self.config.on_gpu)
>        # make sure the connectivity array is reduced
>         shp = self.get_connectivity(dim.value).ndarray.shape
>         return grid_utils.connectivity_for_1d_sparse_fields(
>             dim,
>             shp,
>             from_dim,
>             to_dim,
>             has_skip_values=False,
>             array_ns=xp,
>         )
> ```
> 
> or get rid of this lazy evaluation all together or do the size reduction when setting up the original `_neighbor_tables` dict.

I double-checked, I am not actually updating the `self._neighbor_tables`, just constructing the connectivities with a view of them.

PR #793, opened 2025-07-04
fix bug in computation of dwdz for 3d divergence damping 

OngChia:
`starting_vertical_index_for_3d_divdamp` is wrong when `divdamp_type = 32`. 
`starting_vertical_index_for_3d_divdamp` is used as a manual optimization in the original ICON. It is a k starting index for computation of `dwdz_at_cells_on_model_levels` (declared as `z_dwdz_dd` in ICON) which is used later in the computation of 3D divergence damping. 
We have decided to remove this index in icon4py. The purpose of this PR is to 
(1) remove `starting_vertical_index_for_3d_divdamp` and
(2) add verification for `dwdz_at_cells_on_model_levels` in `test_vertically_implicit_solver_at_predictor_step` that was missing before.

### DIFF
--- a/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
+++ b/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
@@ -14,7 +14,6 @@ from typing import Final, Optional
 import gt4py.next as gtx
 from gt4py.next import backend as gtx_backend
 
-from icon4py.model.atmosphere.dycore import dycore_states
 import icon4py.model.atmosphere.dycore.solve_nonhydro_stencils as nhsolve_stencils
 import icon4py.model.common.grid.states as grid_states
 import icon4py.model.common.utils as common_utils
@@ -22,14 +21,13 @@ from icon4py.model.common.utils import data_allocation as data_alloc
 
 from icon4py.model.common import constants
 from icon4py.model.atmosphere.dycore.stencils import (
+    compute_cell_diagnostics_for_dycore,
     compute_edge_diagnostics_for_dycore_and_update_vn,
-    compute_results_for_thermodynamic_variables,
+    vertically_implicit_dycore_solver,
 )
 from icon4py.model.atmosphere.dycore.stencils.init_cell_kdim_field_with_zero_wp import (
     init_cell_kdim_field_with_zero_wp,
 )
-
-from icon4py.model.atmosphere.dycore.stencils import compute_cell_diagnostics_for_dycore
 from icon4py.model.atmosphere.dycore.stencils.accumulate_prep_adv_fields import (
     accumulate_prep_adv_fields,
 )
@@ -40,7 +38,6 @@ from icon4py.model.atmosphere.dycore.stencils.compute_avg_vn import compute_avg_
 from icon4py.model.atmosphere.dycore.stencils.compute_avg_vn_and_graddiv_vn_and_vt import (
     compute_avg_vn_and_graddiv_vn_and_vt,
 )
-from icon4py.model.atmosphere.dycore.stencils import vertically_implicit_dycore_solver
 from icon4py.model.atmosphere.dycore.stencils.compute_dwdz_for_divergence_damping import (
     compute_dwdz_for_divergence_damping,
 )
@@ -407,20 +404,6 @@ class SolveNonhydro:
         self._edge_geometry = edge_geometry
         self._cell_params = cell_geometry
 
-        #:  start level for 3D divergence damping terms
-        #: this is only different from 0 if divdamp_type == 32: calculation done in mo_vertical_grid.f90
-        xp = data_alloc.import_array_ns(self._backend)
-        self._starting_vertical_index_for_3d_divdamp = (
-            xp.min(
-                xp.where(self._metric_state_nonhydro.scaling_factor_for_3d_divdamp.ndarray > 0.0)
-            )[0]
-            if self._config.divdamp_type == 32
-            else 0
-        )
-        """
-        Declared as kstart_dd3d in ICON.
-        """
-
         self._compute_theta_and_exner = compute_theta_and_exner.with_backend(self._backend).compile(
             vertical_start=[gtx.int32(0)],
             vertical_end=[gtx.int32(self._grid.num_levels)],
@@ -482,7 +465,6 @@ class SolveNonhydro:
             is_iau_active=[self._config.is_iau_active],
             limited_area=[self._grid.limited_area],
             divdamp_order=[self._config.divdamp_order],
-            starting_vertical_index_for_3d_divdamp=[self._starting_vertical_index_for_3d_divdamp],
             vertical_start=[gtx.int32(0)],
             vertical_end=[gtx.int32(self._grid.num_levels)],
             offset_provider=self._grid.connectivities,
@@ -514,7 +496,6 @@ class SolveNonhydro:
             rayleigh_type=[self._config.rayleigh_type],
             divdamp_type=[self._config.divdamp_type],
             end_index_of_damping_layer=[self._vertical_params.end_index_of_damping_layer],
-            starting_vertical_index_for_3d_divdamp=[self._starting_vertical_index_for_3d_divdamp],
             kstart_moist=[self._vertical_params.kstart_moist],
             flat_level_index_plus1=[gtx.int32(self._vertical_params.nflatlev + 1)],
             vertical_start_index_model_top=[gtx.int32(0)],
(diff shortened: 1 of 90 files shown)


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -352,10 +353,14 @@ def _validate(self):
 class NonHydrostaticParams:
     """Calculates derived quantities depending on the NonHydrostaticConfig."""
 
-    def __init__(self, config: NonHydrostaticConfig):
+    def __init__(self, config: NonHydrostaticConfig, scaling_factor_for_3d_divdamp: np.ndarray):
         #:  start level for 3D divergence damping terms
         #: this is only different from 0 if divdamp_type == 32: calculation done in mo_vertical_grid.f90
-        self.starting_vertical_index_for_3d_divdamp: Final[int] = 0
+        self.starting_vertical_index_for_3d_divdamp: Final[int] = (
```
halungge, line 359: This should for sure not be using numpy.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -609,11 +614,6 @@ def __init__(
         self._init_test_fields = nhsolve_stencils.init_test_fields.with_backend(
             self._backend
         ).compile(offset_provider={})
-        if self._config.divdamp_type == 32:
```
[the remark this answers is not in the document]
havogt: Ah, I think it pre-compiled the wrong program, but then we executed the correct one, right?

halungge, line 612: What is the problem with this? I remember now, that we looked at an issue with @huppd in May (one of the things in the "Hannes handover" document, we needed to add the [0] in order to get the first element of the tuple that where returns)

I think, after we fixed that @huppd could run the MCH experiments? So I am wondering whether there is an issue at all.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -352,10 +353,14 @@ def _validate(self):
 class NonHydrostaticParams:
     """Calculates derived quantities depending on the NonHydrostaticConfig."""
 
-    def __init__(self, config: NonHydrostaticConfig):
+    def __init__(self, config: NonHydrostaticConfig, scaling_factor_for_3d_divdamp: np.ndarray):
```
halungge, line 356: also I don't like this random array being passed here. The `NonHydrostaticConfig` contains namelist parameters and the `NonHydrostaticParams` quantities derived from them. No fields should be passed in here.

halungge (changes requested):

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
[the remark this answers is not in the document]
OngChia: Ahh, good point! I now just realized that `starting_vertical_index_for_3d_divdamp` is only for optimization because `scaling_factor_for_3d_divdamp` is zero when `k < starting_vertical_index_for_3d_divdamp`. The final result should not change if we erroneously set `starting_vertical_index_for_3d_divdamp`  tou zero!
Please correct me if I am wrong @muellch

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: Agreed. I think if we need to keep `starting_vertical_index_for_3d_divdamp`, we keep this here and use it in the relevant stencils. We remove the same variable in `NonHydrostaticParams`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
OngChia: Yes. Me either... I will keep the one in `SolveNonhydro` and remove this one here.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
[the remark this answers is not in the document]
havogt: My understanding is `inv_ddqz_z_full_wp` is zero, therefore `dwdz_at_cells_on_model_levels` is zero there as well. I would be interested what the value of `starting_vertical_index_for_3d_divdamp` is for typical simulations. In mch_ch2 it's `4` (Fortran indexing). I doubt this will have measurable impact on performance and would vote for getting rid of it.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -609,11 +612,6 @@ def __init__(
         self._init_test_fields = nhsolve_stencils.init_test_fields.with_backend(
             self._backend
         ).compile(offset_provider={})
-        if self._config.divdamp_type == 32:
-            xp = data_alloc.import_array_ns(self._backend)
-            self.starting_vertical_index_for_3d_divdamp = xp.min(
```
havogt, line 614: Ah, because this line should be `self._params.....` to be correct, right?

OngChia: Yes.

OngChia:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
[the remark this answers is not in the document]
muellch: So if I see this correctly `scalfac_dd3d/scaling_factor_for_3d_divdamp` is 0 below `kstart_dd3d/starting_vertical_index_for_3d_divdamp`, so we do not make a mistake in the computation of `z_graddiv_vn`.

Additionally, the `z_dwdz_dd` we compute with the same `starting_vertical_index_for_3d_divdamp` here:
https://github.com/C2SM/icon-exclaim/blob/f51df5591cdd1059adbdf912d35eeffaa2e88c05/src/atm_dyn_iconam/mo_solve_nonhydro.f90#L3050 
is only used in the computation of `z_graddiv_vn` and hence also multiplied by zero by the `scaling_factor_for_3d_divdamp`.

So then yes, `starting_vertical_index_for_3d_divdamp` is an optimization and can be dropped. This needs to be well documented though.

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
```diff
@@ -331,25 +331,24 @@ def _apply_divergence_damping_and_update_vn(
     is_iau_active: bool,
     limited_area: bool,
     divdamp_order: gtx.int32,
-    starting_vertical_index_for_3d_divdamp: gtx.int32,
     end_edge_halo_level_2: gtx.int32,
     start_edge_lateral_boundary_level_7: gtx.int32,
     start_edge_nudging_level_2: gtx.int32,
     end_edge_local: gtx.int32,
 ) -> fa.EdgeKField[ta.wpfloat]:
+    # add dw/dz for divergence damping term. In ICON, this stencil starts from k = kstart_dd3d until k = nlev - 1.
```
muellch, line 339: Nice comments!

muellch (approved):
Looks good, I left 2 comments.

model/atmosphere/dycore/tests/dycore_tests/test_solve_nonhydro.py
```diff
@@ -1843,6 +1846,9 @@ def test_vertically_implicit_solver_at_predictor_step(
     theta_v_ref = sp_nh_exit.theta_v_new()
     exner_dyn_incr_ref = sp_nh_exit.exner_dyn_incr()
 
+    z_dwdz_dd_ref_with_zero_in_2d_divdamp_layers = sp_nh_exit.z_dwdz_dd().asnumpy()
+    z_dwdz_dd_ref_with_zero_in_2d_divdamp_layers[0:starting_vertical_index_for_3d_divdamp] = 0.0
```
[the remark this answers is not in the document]
OngChia: My understanding is that the serialized data of `dwdz` from ICON openacc run will contain garbage value when `k = [0,starting_vertical_index_for_3d_divdamp-1]` (inclusive) if `divdamp_type == 32`. So, I force the reference equal to zero when `k < starting_vertical_index_for_3d_divdamp`. I agree the first option is the cleanest way. But probably we need to modify how the data is serialized. I would opt for the second option and also write a comment there to explain why we need this extra computation.

PR #798, opened 2025-07-15
Integrate combined program apply_diffusion_to_theta_and_exner

nfarabullini:
Include `apply_diffusion_to_theta_and_exner` into diffusion model and remove stencils already included within


nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -461,28 +455,14 @@ def __init__(
                 offset_provider=self._grid.connectivities,
             )
         )
-        self.calculate_nabla2_for_theta = calculate_nabla2_for_theta.with_backend(
+        self.apply_diffusion_to_theta_and_exner = apply_diffusion_to_theta_and_exner.with_backend(
             self._backend
         ).compile(
             enable_jit=False,
```
havogt, line 461:
```suggestion
            enable_jit=False,
            apply_zdiffusion_t=self.config.apply_zdiffusion_t,
```
I think this should be a compile time option

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -896,57 +876,30 @@ def _do_diffusion_step(
...
                 rd_o_cvd=self.rd_o_cvd,
+                apply_zdiffusion_t=self.config.apply_zdiffusion_t,
                 horizontal_start=self._cell_start_nudging,
                 horizontal_end=self._cell_end_local,
                 vertical_start=0,
                 vertical_end=self._grid.num_levels,
+                offset_provider=self._grid.connectivities,
             )
-            log.debug("running stencil 16 (update_theta_and_exner): end")
+            log.debug("running stencil 13 and 16 (apply_diffusion_to_theta_and_exner): end")
```
havogt, line 902:
```suggestion
            log.debug("running stencil 13 to 16 (apply_diffusion_to_theta_and_exner): end")
```

havogt: I think we could also drop the old stencil names, i.e.
```suggestion
            log.debug("running apply_diffusion_to_theta_and_exner: end")
```
but that should be applied to the rest of the file, and can be done in a separate PR.

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -824,7 +825,7 @@ def _linearize_first_2dims(
             allocator=self.backend,
         )
 
-    @IconSavepoint.optionally_registered()
+    @IconSavepoint.optionally_registered(dims.CECDim, dims.KDim)
```
havogt, line 828:
```suggestion
    @IconSavepoint.optionally_registered(dims=(dims.CECDim, dims.KDim))
```
```suggestion
    @IconSavepoint.optionally_registered(like="zd_vertoffset")
```

havogt:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -461,28 +455,14 @@ def __init__(
                 offset_provider=self._grid.connectivities,
             )
         )
-        self.calculate_nabla2_for_theta = calculate_nabla2_for_theta.with_backend(
+        self.apply_diffusion_to_theta_and_exner = apply_diffusion_to_theta_and_exner.with_backend(
             self._backend
         ).compile(
```
edopao, line 460: Please set `enable_jit=False` (it was removed for testing at some point).

havogt:
blocked by https://github.com/C2SM/icon-exclaim/pull/361

model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -69,8 +69,12 @@ def wrapper(self, *args, **kwargs):
                         f"{name}: field not registered in savepoint {self.savepoint.metainfo}"
                     )
                     if dims:
-                        shp = tuple(self.sizes[d] for d in dims)
-                        return gtx.as_field(dims, np.zeros(shp), allocator=self.backend)
+                        # We allocate a dummy field with size 1 in each dimension
```
havogt, line 72: what do you think about this?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -896,57 +877,34 @@ def _do_diffusion_step(
...
                 area=self._cell_params.area,
                 theta_v=prognostic_state.theta_v,
                 exner=prognostic_state.exner,
                 rd_o_cvd=self.rd_o_cvd,
+                apply_zdiffusion_t=self.config.apply_zdiffusion_t,
                 horizontal_start=self._cell_start_nudging,
                 horizontal_end=self._cell_end_local,
                 vertical_start=0,
                 vertical_end=self._grid.num_levels,
+                offset_provider=self._grid.connectivities,
```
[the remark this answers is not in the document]
havogt: Currently the semantics is: on `compile` compile a variant with these compile time values and offset_provider. If the user calls with a different argument, we will jit with this argument (making it compile time as well) unless `enable_jit=False`.
We are not sure what the right interface is, we could also re-introduce a `bind` which binds an argument and removes the parameter, but not sure if that's what we want in all cases.

model/testing/src/icon4py/model/testing/serialbox.py
havogt: Yes, I checked that if the user disables the code-path that uses these fields, they are not accessed in the generated code.

Longer term we want to have optional fields, but we didn't think of an interface. In the example of this PR, the field would be only optional if `apply_zdiffusion_t==False`. So maybe we want syntax like `OptionalIf[Field[...], condition]`. Then, should the condition be required to be compile-time etc...

havogt:
cscs-ci run default

PR #806, opened 2025-07-21
fix: nudge_max_coeff in solve_nonhydro

havogt:
In ICON Fortran there is an inconsistency between the `nudge_max_coeff` in the namelist and at runtime, see https://gitlab.dkrz.de/icon/icon-model/-/blob/289490d6c166f27c280df246c6df2da9446c8c9b/src/namelists/mo_interpol_nml.f90#L257

Here we introduce `max_nudging_coefficient` as the properly scaled one (the one that ICON uses at runtime) but keep `_nudge_max_config` for initialization with the ICON namelist value. ICON4Py users should ideally only work with `max_nudging_coefficient`.

### DIFF
--- a/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
+++ b/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
@@ -219,8 +219,8 @@ class NonHydrostaticConfig:
         l_vert_nested: bool = False,
         rhotheta_offctr: float = -0.1,
         veladv_offctr: float = 0.25,
-        _max_nudging_coefficient: float = None,  # default is set in __init__
-        scaled_max_nudging_coefficient: float = None,  # default is set in __init__
+        _nudge_max_coeff: float = None,  # default is set in __init__
+        max_nudging_coefficient: float = None,  # default is set in __init__
         fourth_order_divdamp_factor: float = 0.0025,
         fourth_order_divdamp_factor2: float = 0.004,
         fourth_order_divdamp_factor3: float = 0.004,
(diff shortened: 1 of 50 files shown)


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -1335,7 +1335,7 @@ def run_corrector_step(
             gtx.int32(self._config.divdamp_order),
             self._grid.global_properties.mean_cell_area,
             second_order_divdamp_factor,
-            self._config.nudge_max_coeff,
+            self._config.nudge_max_coeff * 5.0,
```
havogt, line 1338: We need to fix this in a way that doesn't hurt us in the future...

havogt: Consider using `max_nudging_coeff` like in diffusion. And apply same pattern as in diffusion.

model/common/src/icon4py/model/common/constants.py
```diff
@@ -79,7 +79,9 @@
 # Implementation constants
 #: default physics to dynamics time step ratio
 # TODO (magdalena) not a constant, this is a default config parameter
-DEFAULT_PHYSICS_DYNAMICS_TIMESTEP_RATIO: Final[float] = 5.0
+DEFAULT_PHYSICS_DYNAMICS_TIMESTEP_RATIO: Final[
+    float
+] = 5.0  # TODO isn't this the inverse i.e. should it be called DEFAULT_DYNAMICS_TO_PHYSICS_TIMESTEP_RATIO
```
halungge, line 84: Agree with the name change.... 

Also: can you remove the `# TODO ` above. I don't know why I put that here. It is the default not the dynamic substep ratio, so it _is_ a constant after all.

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         self._config = {
             "divavg_cntrwgt": 0.5,
             "weighting_factor": 0.0,
-            "nudge_max_coeffs": 0.375,
+            "nudge_max_coeffs": 0.375,  # TODO why the extra `s` at the end? is this a typo?
```
halungge, line 63: I would say so. It is called exactly like this in the function that uses it in `compute_nudgecoeffs.py` (which computes a field...)

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -311,7 +312,26 @@ def __init__(
         #: parameters from other namelists:
 
         #: from mo_interpol_nml.f90
-        self.nudge_max_coeff: float = max_nudging_coeff
+
```
halungge, line 315: Can you add a `# TODO` that this should be handled in the configuration utility, once we get to this, such that we can get rid of the duplication.

halungge (commented):
Looks good, only some small things mostly naming and documentation

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
havogt: you agree it's a typo?

havogt: I renamed it

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -60,7 +60,7 @@ def __init__(
         self._config = {
             "divavg_cntrwgt": 0.5,
             "weighting_factor": 0.0,
-            "nudge_max_coeffs": 0.375,
+            "nudge_max_coeff": 0.375,  # TODO why the extra `s` at the end? is this a typo?
```
havogt, line 63: this is the scaled one, look again once we know how to call it

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -82,7 +83,7 @@ def _mch_ch_r04b09_diffusion_config():
             smagorinski_scaling_factor=0.025,
             zdiffu_t=True,
             velocity_boundary_diffusion_denom=150.0,
-            max_nudging_coeff=0.075,
+            max_nudging_coefficient=0.075 * constants.DEFAULT_DYNAMICS_TO_PHYSICS_TIMESTEP_RATIO,
```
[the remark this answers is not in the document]
OngChia: Good catch here. 
For the r04b09 experiment, it should be 0.075 in both cases. The default value is 0.02. This coefficient does nothing when doing global simulations or on torus grid. Moreover, I think we should remove manually setting `max_nudging_coefficient` in the `_jabw_nonhydro_config()` and `_gauss_nonhydro_config()` and `_jabw_diffusion_config()` to avoid any confusion.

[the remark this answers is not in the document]
havogt: Thanks! fixed.

havogt:
cscs-ci run default

havogt:
cscs-ci run default

model/driver/src/icon4py/model/driver/icon4py_configuration.py
```diff
@@ -168,7 +168,7 @@ def _gauss3d_nonhydro_config(n_substeps: int):
         return solve_nh.NonHydrostaticConfig(
             igradp_method=3,
             ndyn_substeps_var=n_substeps,
-            max_nudging_coeff=0.02,
+            max_nudging_coefficient=0.02 * constants.DEFAULT_DYNAMICS_TO_PHYSICS_TIMESTEP_RATIO,
```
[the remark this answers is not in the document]
havogt: done

havogt:
cscs-ci run default

PR #811, opened 2025-07-23
Empty allocate local fields

havogt:
This would allow in a next step to temporarily request fields from a pool (or alloc via numpy/cupy from a pool).


msimberg:
Is this still something you want to try?

PR #821, opened 2025-07-31
Fix missing raises on exception statements

msimberg:
Fixes a couple I introduced in #709 and another one that existed from before. These are triggered by PLW0133 for the linter (https://docs.astral.sh/ruff/rules/useless-exception-statement/, equivalent to https://pylint.readthedocs.io/en/latest/user_guide/messages/warning/pointless-exception-statement.html from pylint). PLW will be enabled by #842.


pyproject.toml
```diff
@@ -206,7 +206,10 @@ docstring-code-format = true
 # YTT: flake8-2020
 exclude = ["**/tests/**"]
 explicit-preview-rules = true
-extend-select = ["F822"]  # TODO(egparedes): remove when not longer in preview
+extend-select = [
+  'F822',  # TODO(egparedes): remove when not longer in preview
+  'PLW0133'  # [useless-exception-statement]
+]
```
msimberg, line 212: Is there a convention for what goes in `extend-select` and what goes in `select`? I think I understand the difference between the two, but I don't understand e.g. why F822 is in `extend-select` and why FA100 is in `select`...

msimberg:
> If you want an answer to your question before merging, you probably need to wait for Enrique to be back.

Doesn't hurt to wait, this isn't urgent.

msimberg:
Note that https://github.com/C2SM/icon4py/pull/829 also fixes the missing raises in `.../grid/icon.py`, but this PR is still needed for the linter warnings and the other missing raise.

halungge (approved):

pyproject.toml
egparedes: The reason why `F822` goes in `extend-select` is a bit hacky and arbitrary. This rule comes from an active ruleset (`F`) but  still needs to be explicitly activated because it is a rule in `preview` mode and `explicit-preview-rules=true` (to avoid loading all preview rules by default). Since the explicit activation of this rule will become unnecessary as soon as it is removed from preview mode, adding it to `extend-select` was a hack to try to make more visible that it was a temporary activation.

In #839 , this rule activation has been moved to `select` as expected.

egparedes: I would avoid keep this PR as a simple fix for the missing `raise` statements, and activate more ruff rulesets in other PRs like #839

pyproject.toml
```diff
@@ -226,6 +226,7 @@ select = [
   # 'PLE', # Pylint-Error
   # 'PLR', # Pylint-Refactor
   # 'PLW', # Pylint-Warning
+  'PLW0133',  # useless-exception-statement
```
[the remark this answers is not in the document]
egparedes: Yes, please. PR #842 should be merged today / tomorrow, so I would leave out the rule from this PR

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

PR #823, opened 2025-07-31
Run CI test reminder workflow on `pull_request_target` events

msimberg:
This ensures it runs in the context of the target branch (main icon4py repository) and has sufficient permissions to write the comment on a PR even when the PR comes from a fork.

### DIFF
--- a/.github/workflows/mandatory_and_optional_test_reminder.yml
+++ b/.github/workflows/mandatory_and_optional_test_reminder.yml
@@ -1,5 +1,5 @@
 on:
-  pull_request:
+  pull_request_target:
 
 jobs:
   test:


msimberg:
I'm 99% percent sure this is correct, but I can't test it on the PR itself. Changing the even type to `pull_request_trigger` means that it also runs the workflow defined on the `main` branch, rather than the on from this PR, for security reasons. The main concern with `pull_request_trigger` is that the workflow has privileged access to the repo, but the workflow can only be changed through PRs _merged to main_, i.e. further changes are also only picked up only when merged (so PRs that try to exploit the workflow can't do so).

msimberg:
cscs-ci run default

msimberg:
Just to follow up: this seems to be working now on a PR from a fork: https://github.com/C2SM/icon4py/pull/829#issuecomment-3164858208.

PR #824, opened 2025-08-04
add version in muphys pyproject.toml

OngChia:
Add version in pyproject.toml of muphy package

### DIFF
--- a/model/atmosphere/subgrid_scale_physics/muphys/pyproject.toml
+++ b/model/atmosphere/subgrid_scale_physics/muphys/pyproject.toml
@@ -1,7 +1,10 @@
+# -- Build system requirements (PEP 518) --
+
 [build-system]
 build-backend = "setuptools.build_meta"
 requires = ["setuptools>=61.0", "wheel>=0.40.0"]
 
+# -- Standard project description options (PEP 621) --
 [project]
 authors = [
   {email = "gridtools@cscs.ch"},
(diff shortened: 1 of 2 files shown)


model/atmosphere/subgrid_scale_physics/muphys/pyproject.toml
```diff
@@ -22,16 +25,19 @@ classifiers = [
...
 dependencies = [
-  "gt4py==1.0.6",
+  # workspace members
   "icon4py-common[io]>=0.0.6",
+  # external dependencies
+  "gt4py==1.0.6",
   "packaging>=20.0"
 ]
 description = "ICON subgrid scale muphys parameterization."
-dynamic = ['version']
```
OngChia, line 30: I removed this line because it seems to conflict with the version at line 40 below. The error is 
```
configuration error: You cannot provide a value for `project.version` and list it under `project.dynamic` at the same time
DESCRIPTION:
    According to PEP 621:  Build back-ends MUST raise an error if the metadata
    specifies a field statically as well as being listed in dynamic.

GIVEN VALUE:
    {
        "version": "0.0.6",
        "...": " # ...",
        "dynamic": [
            "version"
        ]
    }

OFFENDING RULE: 'PEP 621'

DEFINITION:
    {
        "see": "https://packaging.python.org/en/latest/specifications/pyproject-toml/#dynamic"
    }

```

uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
```
OngChia, line 2: This automatically becomes 3. Is this correct?

havogt (approved):
lgtm

PR #827, opened 2025-08-06
Return max_vertical_cfl to Fortran

havogt:
The only mechanism we currently have in py2fgen to pass information out of Python is via arrays. Therefore we receive the max_vcfl from Fortran as an array with a single element. Then we forward it to the granule as an integer and pass it back via the array.
On the Fortran side the same mechanism is implemented: `scalar -> array` -> `timestep` -> `array -> scalar`.


havogt:
cscs-ci run default

havogt:
cscs-ci run default

PR #845, opened 2025-08-22
Compute domain bounds from grid refinement

halungge:
Compute start and end index of grid domain from the refinement control fields.

### Additions:
- GridManager does not any longer read start_index , end_index from the grid file, instead it computes those values from the grid refinement fields.
- `start_index`, `end_index` functions are directly passed to the `Grid` instead of lookup arrays, they can be initialized from different sources (ICON arrays or grid refinement, ...)
### Cleanups:
- removes lot of constants initially introduced to track the ICON behavior in `horizontal.py`. The translation to ICON should be possible through documentation.


model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -312,31 +312,48 @@ def edge_cell_length(self):
...
-        return self._read_int32_shift1("c_start_index")
+        start_idx = self._read_int32("c_start_index")
+        return np.where(start_idx == 0, start_idx, start_idx - 1)
 
     def cells_end_index(self):
         return self._read_int32("c_end_index")
 
     def vertex_start_index(self):
-        return self._read_int32_shift1("v_start_index")
+        start_idx = self._read_int32("v_start_index")
```
halungge, line 322: ICON sets some of these values to 0. I suspect these are values that are not used otherwise ICON would crash since it has 1 based fields (?)/ But this might be an issue in the wrapper.

tools/src/icon4py/tools/py2fgen/wrappers/common.py
```diff
@@ -165,14 +165,14 @@ def construct_icon_grid(
...
     xp = data_alloc.import_array_ns(backend)
-
-    cells_start_index = adjust_fortran_indices(cell_starts)
-    vertex_start_index = adjust_fortran_indices(vertex_starts)
-    edge_start_index = adjust_fortran_indices(edge_starts)
-
-    cells_end_index = cell_ends
-    vertex_end_index = vertex_ends
-    edge_end_index = edge_ends
+    # TODO (halungge): icon has 0 values in these arrays in some places possibly where they don't use them in case
```
halungge, line 168: See my comment above in `serialbox.py`. Whenever the values are `> 0` the denote real indices, but there are 0 values which probably denote "invalid". Generally adjusting them to 0 based Python indices might be a source of errors.

halungge:
cscs-ci run default

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+    dims.CellDim: 14,
+    dims.EdgeDim: 28,
+    dims.VertexDim: 14,
+}
+"""
+Grid points in the grid refinement fields are labeled with their distance to the lateral boundary.
+The distance is counted in "rows" and distances _MAX_BOUNDARY_DISTANCE are computed.
+
+**Rows** can be defined like:
+### CELLS:
```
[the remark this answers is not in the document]
havogt: I gave up on making the Edge, Vertex pictures with Claude, but I think they would still be useful to have in the codebase (or maybe we add them to docs and link from here).

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+Grid points can be ordered (moved to the beginning of an field array) up to any distance <= MAX_BOUNDARY_DISTANCE in the grid-generators, the exact number of
+ordered rows is a parameter to the grid generation.
+"""
+
+_GRID_REFINEMENT_BOUNDARY_WIDTH: Final[dict[gtx.Dimension, int]] = {
+    dims.CellDim: 4,
+    dims.EdgeDim: 9,
+    dims.VertexDim: 4,
+}
+"""Number of cell rows used in ICON grid refinement: meaning there are for lateral boundary cell-rows.
```
havogt, line 72: > meaning there are for lateral boundary cell-rows.

?

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+"""
+
+_UNORDERED: Final[dict[gtx.Dimension, tuple[int, int]]] = {
+    dims.CellDim: (0, -4),
+    dims.EdgeDim: (0, -8),
+    dims.VertexDim: (0, -4),
+}
+"""Refinement value indicating a point is in the unordered interior (fully prognostic) of the grid: this is encoded by 0 or -4 in coarser parent grid."""
```
havogt, line 80: It mean the tuple is to be read as: either `0` or `-4`? (not range or something else)

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+def is_unordered_field(
+    field: data_alloc.NDArray, dim: gtx.Dimension, array_ns: ModuleType = np
+) -> data_alloc.NDArray:
+    assert field.dtype in (gtx.int32, gtx.int64), f"not an integer type {field.dtype}"  # type: ignore [attr-defined]
+    return array_ns.where(
+        field == _UNORDERED[dim][0], True, array_ns.where(field == _UNORDERED[dim][1], True, False)
+    )
+
+
+def convert_to_unnested_refinement_values(
```
havogt, line 162:
```suggestion
def convert_to_non_nested_refinement_values(
```
does it mean that, or is `unnesting` and operation?

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+                if found.size > 0
+                else (0, 0)
+            )
+        elif my_zone is h_grid.Zone.INTERIOR:
+            # for the Vertex and Edges the level after the nudging zones are not ordered anymore, so
+            # we rely on using the end index of the nudging zone for INTERIOR
+            value = get_nudging_refinement_value(dim)
+            found = array_ns.where(refinement_ctrl == value)[0]
+            start_index = array_ns.max(found).item() + 1 if found.size > 0 else 0
+            end_index = refinement_ctrl.shape[0]
```
havogt, line 143: What happens if we fall through? Then
```
        start_index = 0
        end_index = refinement_ctrl.shape[0]
```
is what we use, aka all-the-points, right?

havogt: I.e. this algorithms is: we take the whole domain, then for same cases we shrink it?

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+    return _LAST_BOUNDARY[dim].level + _LAST_NUDGING[dim].level
+
+
+def is_unordered_field(
+    field: data_alloc.NDArray, dim: gtx.Dimension, array_ns: ModuleType = np
+) -> data_alloc.NDArray:
+    assert field.dtype in (gtx.int32, gtx.int64), f"not an integer type {field.dtype}"  # type: ignore [attr-defined]
+    return array_ns.where(
+        field == _UNORDERED[dim][0], True, array_ns.where(field == _UNORDERED[dim][1], True, False)
+    )
```
havogt, line 159: Is this
```suggestion
    return array_ns.isin(field, _UNORDERED[dim])
```

model/common/src/icon4py/model/common/grid/horizontal.py
```diff
@@ -46,199 +46,135 @@
...
 }
 
-_BOUNDS = {
+_ICON_CONSTANTS_BOUNDS = {
     dims.CellDim: (0, _CELL_GRF - 1),
     dims.EdgeDim: (0, _EDGE_GRF - 1),
     dims.VertexDim: (0, _VERTEX_GRF - 1),
 }
+"""
+Indices used to look up start index and end index in the arrays int `start_idx_[c,e,v]` from ICON.
```
havogt, line 90: Is this docstring for all or only the last?

havogt (commented):
I didn't go through all of it in detail. Are there specific places where it would be good if I go deep?

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+Refinement control for ICON grid.
+
+Grid refinement is used in the context of
+- local area grids to determine the type of a grid point: its distance from the boundary
+- nested horizontal grids to determine the nested overlap regions for feedback
+- domain decomposition to order grid points
+
+See Zaengl et al. Grid Refinement in ICON v2.6.4 (Geosci. Model Dev., 15, 7153-7176, 202)
+
+This module only contains functionality related to grid refinement as we use it in ICON4Py.
```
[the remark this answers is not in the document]
halungge: I tried, I can also otherwise just remove it.

model/common/src/icon4py/model/common/grid/grid_refinement.py
halungge: yes, kind of.  For a non limited area grid there is nothing other than "all points" except for halos, but the halos are not yet taken into account here. (See TODO above) The problem is that due to the ordering that they do or don't in the limited area case it is far less standard than I originall thought, or I did not get the right hook to implement it in a decent way.

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+_MAX_BOUNDARY_DISTANCE: Final[dict[gtx.Dimension, int]] = {
+    dims.CellDim: 14,
+    dims.EdgeDim: 28,
+    dims.VertexDim: 14,
+}
+"""
+Grid points in the grid refinement fields are labeled with their distance to the lateral boundary.
+The distance is counted in "rows" and distances _MAX_BOUNDARY_DISTANCE are computed.
```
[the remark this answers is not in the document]
halungge: updated documentation. I hope it becomes more clear.

model/common/src/icon4py/model/common/grid/horizontal.py
halungge: For all

halungge:
> I didn't go through all of it in detail. Are there specific places where it would be good if I go deep?

I tried to remove as much as possible of all these constants ported from ICON, we still need the original ones though to match the values from serialbox.

The computation itself in grid_refinement.py is not very standardized or homogenous among grid zones and dimension, I did not find a better way, as the data in the grid file actually is less generic than I originally thought.

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,242 @@
...
+ /  \  /  \  /  \  /
+/____\/____\/____\/
+
+
+### VERTEX:
+1. row: vertex on the lateral boundary
+2. row: shared between cell of 1. row and 2. row cell.
+...
+
+         1 xxxx 1 xxxxx 1 xxxxx  ← Lateral boundary, row 1 VERTICES
```
havogt, line 71: Nice!

model/common/src/icon4py/model/common/grid/grid_refinement.py
```diff
@@ -0,0 +1,242 @@
...
+
+_MAX_BOUNDARY_DISTANCE: Final[dict[gtx.Dimension, int]] = {
+    dims.CellDim: 14,
+    dims.EdgeDim: 28,
+    dims.VertexDim: 14,
+}
+"""
+Grid points in the grid refinement fields are labeled with their distance to the lateral boundary.
+The distance is counted in "rows" and distances up to _MAX_BOUNDARY_DISTANCE are computed, elements
+a distances larger than _MAX_BOUNDARY_DISTANCE are labeled as _UNORDERED
```
havogt, line 46:
```suggestion
at distances larger than _MAX_BOUNDARY_DISTANCE are labeled as _UNORDERED
```

havogt (approved):
Nice documentation!

halungge:
cscs-ci run default

halungge:
cscs-ci run extra

halungge:
cscs-ci run default

PR #847, opened 2025-08-27
More torus support cleanup

msimberg:
This is on top of https://github.com/C2SM/icon4py/pull/829. Diff: https://github.com/msimberg/icon4py/compare/better-torus-support...msimberg:icon4py:better-torus-support-2.

This only removes the `from_mean_cell_area` constructor from `GridParams` as a cleanup. I realized it doesn't add much value since it doesn't do any extra logic or validation, all it did was provide a convenient way to pass the `mean_cell_area` as the first argument. I've made the `GridParams` constructor take keyword-only parameters so it's easy to pass only the `mean_cell_area` with the regular constructor.


egparedes (approved):
LGTM, just a type hinting change and a question.

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -79,61 +79,57 @@ def __init__(
...
-        num_cells: int | None = None,
         radius: float = constants.EARTH_RADIUS,
-    ):
-        return cls(
-            grid_shape,
-            num_cells,
-            mean_cell_area,
-            radius,
-        )
+        num_cells: int | None = None,
```
halungge, line 91: Do this two parameters even make sense? What about these this test?
```
def test_num_cell_arg():
    shape = GridShape(geometry_type=ICOSAHEDRON), GridSubdivison(root = 4, level=9))
    params = GlobalGridParams(grid_shape=shape, num_cells = 2)
    assert 2 == params.num_cells
```

msimberg: Yeah, I suppose this particular combination isn't super realistic, but it just shows that it gives priority to the explicit `num_cells` parameter. The more meaningful changes are coming in https://github.com/C2SM/icon4py/pull/848, so may be worth waiting until that one for a longer discussion on what should be preferred/allowed and what not.

PR #852, opened 2025-08-29
Customize backend

nfarabullini:


nfarabullini:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -396,6 +396,7 @@ def __init__(
         owner_mask: fa.CellField[bool],
         backend: gtx_backend.Backend | None,
         exchange: decomposition.ExchangeRuntime = decomposition.SingleNodeExchange(),
+        backend_options: dict | None = None,
```
havogt, line 399:
```suggestion
        backend: Backend | BackendDescriptor | None = None,
        exchange: decomposition.ExchangeRuntime = decomposition.SingleNodeExchange(),
```

havogt: class BackendDescriptor:
    backend_kind: Literal["gtfn", "dace"]
    device: Literal["gpu", "cpu"]

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -358,7 +358,7 @@ def __init__(
         interpolation_state: diffusion_states.DiffusionInterpolationState,
         edge_params: grid_states.EdgeParams,
         cell_params: grid_states.CellParams,
-        backend: gtx_backend.Backend | None,
+        backend: gtx.backend.Backend | Literal["gpu", "cpu"] | dict[str, Any] | None = None,
```
havogt, line 361: - we should use an enum for gpu/cpu or something similar -> see slack discussion
- I would propose to introduce an type alias for `dict[str, Any]` to help the reader, e.g. `BackendDescription: TypeAlias = dict[str, Any]`

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -397,7 +397,7 @@ def __init__(
         self._determine_horizontal_domains()
 
         self.mo_intp_rbf_rbf_vec_interpol_vertex = setup_program(
-            backend=self._backend,
+            backend=backend,
```
havogt, line 400: after this change, do we need to store `backend` in the instance or can we drop it?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity_advection.py
```diff
@@ -44,10 +45,12 @@ def __init__(
...
         edge_params: grid_states.EdgeParams,
         owner_mask: fa.CellField[bool],
-        backend: gtx_backend.Backend | None,
+        backend: gtx.backend.Backend | Literal["gpu", "cpu"] | dict[str, Any] | None = None,
     ):
         self.grid: icon_grid.IconGrid = grid
-        self._backend = backend
+        self._backend = (
+            backend if isinstance(backend, gtx.backend.Backend) else customize_backend(**backend)
+        )
```
havogt, line 53: Why this different pattern here compared to diffusion/solve_nonhydro? Is it a refactoring artifact?

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -50,8 +75,14 @@ def setup_program(
     vertical_sizes = {} if vertical_sizes is None else vertical_sizes
     offset_provider = {} if offset_provider is None else offset_provider
 
+    if isinstance(backend, gtx.backend.Backend):
+        custom_backend = backend
+    else:
+        # customized backend params
+        custom_backend = customize_backend(**backend)
```
havogt, line 82: Maybe just
```suggestion
    if not isinstance(backend, (gtx.backend.Backend, None)):
        backend = customize_backend(**backend)
```

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -13,18 +13,43 @@
...
+    if options["backend_kind"] == "dace":
+        backend_func = make_custom_dace_backend
+    elif options["backend_kind"] == "gtfn":
+        backend_func = make_custom_gtfn_backend
+    on_gpu = backend["device"] == "gpu"
+    custom_backend = backend_func(
+        on_gpu=on_gpu,
+        **options,
+    )
+    return custom_backend
```
havogt, line 44: A few more thoughts (I make one comment per idea, to be able to discuss them separately)

havogt: Maybe instead of `backend_kind` we could use `backend_factory` directly, i.e. the advanced user (aka performance engineer) instead of passing a str-identifier could directly pass the `make_custom_dace_backend` function.
It would look like:
```suggestion
def customize_backend(program: str = "", arch: str = "", **backend):
    program_id = str(program.past_stage.past_node.id) if program != "" else program
    options = get_options(program_id, arch, **backend)
    on_gpu = backend["device"] == "gpu"
    backend_factory = options.pop("backend_factory")
    custom_backend = backend_factory(
        on_gpu=on_gpu,
        **options,
    )
    return custom_backend
```
and
```python
def get_options(program_name: str, arch: str, **backend):
    backend_factory = backend.get("backend_factory", make_custom_dace_backend)
    return dict(backend_kind=backend_kind)
```

model/common/src/icon4py/model/common/model_options.py
havogt, line 82: We should inject the program name here, right?
```suggestion
        custom_backend = customize_backend(program = str(program.past_stage.past_node.id), **backend)
```
(or `program_name`?) and corresponding changes in `customize_backend`?

model/common/src/icon4py/model/common/model_options.py
havogt, line 44: I think we should change the makers to accept the device kind that we will introduce instead of processing and translating to `on_gpu`. That would also allow a backend maker to reject invalid devices.

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,35 @@
...
+    backend = "run_gtfn_cpu_cached"
+    match_backend = [
+        backend_str.split("_")
+        for backend_str in BACKENDS.keys()
+        if set(backend_str.split("_")).issubset(set(backend.split("_")))
+    ][0]
+    backend_options = {
+        "device": match_backend[1],
+        "backend_kind": match_backend[0],
+    }
```
havogt, line 28: for this test, why not just
```suggestion
    backend = "run_gtfn_cpu_cached"
    match_backend = [
        backend_str.split("_")
        for backend_str in BACKENDS.keys()
        if set(backend_str.split("_")).issubset(set(backend.split("_")))
    ][0]
    backend_options = {
        "device": "cpu", # or whatever it will be after the discussion
        "backend_kind": "gtfn", # or the corresponding maker if we go that way
    }
```

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,35 @@
...
+    backend = customize_backend(**backend_options)
+    field = data_alloc.zero_field(
+        simple_grid(), dims.EdgeDim, dims.KDim, backend=backend, dtype=vpfloat
+    )
```
havogt, line 32: would be nice if we find a way to test that we actually got the customized backend here (maybe we could just read it's properties to figure out that we do cpu or gpu)

havogt (commented):
We are still in the design phase. So I have some more idea.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity_advection.py
nfarabullini: left-over. thanks for pointing out

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
nfarabullini: I've reverted back to the original version as the backend argument is also used in `_allocate_local_fields`

model/common/src/icon4py/model/common/model_options.py
nfarabullini: where would you switch between `make_custom_dace_backend` and `make_custom_gtfn_backend`

model/common/src/icon4py/model/common/model_options.py
havogt: Because it's more concise and therefore should be easier to read. (Though the python type checker might disagree.)

model/common/src/icon4py/model/common/model_options.py
havogt: The user would pass one or the other and if nothing is passed the default will be used.

model/common/tests/common/states/unit_tests/test_model_options.py
nfarabullini: it just extrapolates info from backends that we already have. I thought it might be useful to keep it. Does it bother you?

model/common/src/icon4py/model/common/model_options.py
nfarabullini: I don't think that this is a user-friendly approach. They should be able to just say "gtfn" or "dace". We can make gtfn`backend_kind` 's default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
havogt: how would that help the reader?

model/common/src/icon4py/model/common/model_options.py
havogt: You mean you don't like the `not`?

model/common/src/icon4py/model/common/model_options.py
havogt: If it's just about the name you can make an alias from `gtfn` to `make_custom_gtfn_backend`.

havogt: Yes

model/common/tests/common/states/unit_tests/test_model_options.py
havogt: I don't exactly understand what you want to test, so my answer depends on your answer to that question.

model/common/src/icon4py/model/common/model_options.py
havogt: That makes sense to me. Especially, with the last statement. 
>If that was me, I would want to only specify "CPU" or "GPU" and then rely on the others to run whatever backend and backend customization is fastest for the given architecture.

Some observations:
- In this PR, so far we are discussing the granule interface, how we map the driver command line arguments was only partially touched in the discussion, but I agree it makes sense to think about that as well.
- passing a concrete backend to the driver might become a corner-case, because even if the user (= the one who runs the application) wants to say `dace` + `gpu`, they probably would like to run with the best dace variant for every gt4py program, not with the concrete backend that is constructed in icon4py (or gt4py).
- passing a factory can be implemented for the command line similar to how we added passing custom backends to the stencil_tests, there we allow the pattern `--backend my_package.subpackage:my_backend` (see also https://docs.python.org/3/library/pkgutil.html#pkgutil.resolve_name for the standard way to load an object from such a string). Additionally we could make one package `icon4py.model.common.model_backends` or similar the default backend, i.e. if the string doesn't contain a `:`, we could prefix with that package. We could than do `driver --device=gpu, --backend-factory=dace`, which would pass `icon4py.model.common.model_backends.make_custom_dace_backend` (after making an alias `dace` -> `make_custom_dace_backend`, but these are details)

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -11,14 +11,40 @@
...
     return {k: [v] for k, v in d.items()}
 
 
+def get_options(program_name: str, arch: str, **backend):
+    backend_kind = backend.get("backend_kind", "default")
+    return dict(backend_kind=backend_kind)
+
+
+def customize_backend(program_name: str = "", arch: str = "", **backend):
+    options = get_options(program_name, arch, **backend)
```
havogt, line 32:
```suggestion
    if isinstance(backend, DeviceType):
        backend = {"device": backend}
    options = get_options(program_name, arch, **backend)
```

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,34 @@
...
+
+@pytest.mark.parametrize("backend_factory", [make_custom_gtfn_backend, make_custom_dace_backend])
+@pytest.mark.parametrize("device", ["cpu", "gpu"])
+def test_custom_backend_backend_options(backend_factory, device):
+    backend_options = {
+        "backend_factory": backend_factory,
+        "device": device,
+    }
+    backend = customize_backend(backend_options)
+    default_backend = str(backend_factory).split("_")[2] + "_" + device
```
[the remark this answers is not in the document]
nfarabullini: I did it a bit differently, check if it's ok

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+    partial_program = setup_program(backend=backend, program=scale_k)
+    backend = model_backends.BACKENDS["gtfn_cpu"]
+    expected_partial = functools.partial(
+        scale_k.with_backend(backend).compile(
+            enable_jit=False,
+            offset_provider={},
+        ),
+        offset_provider={},
+    )
+    assert str(partial_program) == str(expected_partial)
```
havogt, line 62:
```suggestion
    # TODO: test should be improved to work without string comparison
    assert repr(partial_program) == repr(expected_partial)
```

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+    backend_options: dict = {
+        "backend_factory": backend_factory,
+        "device": model_backends.CPU,
+    }
+    backend = customize_backend(backend_options)
+    backend_name = expected_backend + "_cpu"
+    assert str(model_backends.BACKENDS[backend_name]) == str(backend)
+
+
+def test_custom_backend_device() -> None:
```
havogt, line 35: this test could probably be merged with the previous one

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+                "backend_factory": model_backends.make_custom_dace_backend,
+                "device": model_backends.GPU,
+            },
+            "dace_gpu",
+        ),
+        ({"backend_factory": model_backends.make_custom_dace_backend}, "dace_cpu"),
+        ({"device": model_backends.GPU}, "gtfn_gpu"),
+    ],
+)
+def test_setup_program_dace_gpu(backend_params: typing.Any, expected_backend: str) -> None:
```
havogt, line 80: the test name seems wrong

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+    partial_program = setup_program(backend=backend_params, program=scale_k)
+    backend = model_backends.BACKENDS[expected_backend]
+    expected_partial = functools.partial(
+        scale_k.with_backend(backend).compile(
+            enable_jit=False,
+            offset_provider={},
+        ),
+        offset_provider={},
+    )
+    assert str(partial_program) == str(expected_partial)
```
havogt, line 90:
```suggestion
     # TODO: test should be improved to work without string comparison
    assert repr(partial_program) == repr(expected_partial)
```

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+    ],
+)
+def test_custom_backend_options(backend_factory: typing.Callable, expected_backend: str) -> None:
+    backend_options: dict = {
+        "backend_factory": backend_factory,
+        "device": model_backends.CPU,
+    }
+    backend = customize_backend(backend_options)
+    backend_name = expected_backend + "_cpu"
+    assert str(model_backends.BACKENDS[backend_name]) == str(backend)
```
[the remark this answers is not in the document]
havogt: the above applies to all the tests here

model/common/tests/common/states/unit_tests/test_model_options.py
```diff
@@ -0,0 +1,90 @@
...
+    "backend",
+    [
+        model_backends.BACKENDS["gtfn_cpu"],
+        model_backends.CPU,
+        {"backend_factory": model_backends.make_custom_gtfn_backend, "device": model_backends.CPU},
+        {"backend_factory": model_backends.make_custom_gtfn_backend},
+        {"device": model_backends.CPU},
+    ],
+)
+def test_setup_program_defaults(backend: typing.Any) -> None:
```
havogt, line 52: please add the correct type annotation for backend

model/common/tests/common/states/unit_tests/test_model_options.py
havogt, line 80: please fix type annotation for `backend_params`

havogt (changes requested):
The tests are not ideal as string comparison is quite fragile. For now, it's ok for me if we add a TODO.

model/common/tests/common/states/unit_tests/test_model_options.py
nfarabullini: I prefer to keep them separate, but I have adjusted to the new assert format + TODO

nfarabullini:
cscs-ci run default

PR #856, opened 2025-09-02
Move nflat gradp to metrics

halungge:
Moves `nflat_gradp` out of the `VerticalGrid` placing it in the `NonHydroMetricState`, because it can only be computed together with `flat_idx` from the metrics fields.

Accordingly, `nflat_gradp` has been removed in in the `diffusion` since it is not used, then in `dycore` it has been moved to the metrics field. See also [interface change in icon-exclaim](https://github.com/C2SM/icon-exclaim/pull/372)

Additional change: delete an old unused dycore stencil (51).


halungge:
> `nflat_gradp` is computed via a global reduction over `flat_idx` and use only in the dycore. We had along discussion with @OngChia on whether we should to the reduction in the dycore.init or whether it is needed at all. Removing it would simplify the code but change the result because it would compute extra 2nd order derivative terms in some parts of the vertical grid.

Bottom line: 
- if we get rid of `nflat_gradp` we change the results (theoretically to better accuracy, but still change it) because we use the Taylor expansion to 2nd order (instead of only 1st order) in larger part of the grid. 
-  we cannot compute `nflat_gradp` inside `SolveNonHydro.__init__` because of Blueline:  would need the `flat_idx` for that and that one is local to `subroutine set_nh_metrics` in `mo_vertical_grid.f90`. We would have to do the entire compuation of all related fields `gradp` fields in `SolveNonHydro`

yiluchen1066 (approved):

PR #862, opened 2025-09-04
wip: use grid and experiment definitions

havogt:


havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

model/common/tests/common/interpolation/unit_tests/test_rbf_interpolation.py
```diff
@@ -209,16 +222,21 @@ def test_rbf_interpolation_coeffs_cell(
 @pytest.mark.level("unit")
 @pytest.mark.datatest
 @pytest.mark.parametrize(
-    "grid_file, experiment, atol",
+    "experiment, atol",
     [
-        (dt_utils.R02B04_GLOBAL, dt_utils.GLOBAL_EXPERIMENT, 3e-10),
-        (dt_utils.REGIONAL_EXPERIMENT, dt_utils.REGIONAL_EXPERIMENT, 3e-3),
+        (definitions.Experiments.EXCLAIM_APE, 3e-10),
+        (definitions.Experiments.MCH_CH_R04B09, 3e-3),
```
havogt, line 228: Is the reason for the bigger error the higher resolution? Wondering if it makes sense to maintain a different threshold? Or maybe we could have a threshold per RnBk?

halungge: It happens in quite some places that the regional MCH has a higher error, also in the dycore comparision. I think there are several possible reasons for this: 
- resolution (but the R04B09 is higher resolution than the R02B04) 
- boundary effects
- possibly mixed precision, (I do not even know whether the `mch_ch_r04b09` is mixed precision)


I don't know therefore whether a threshold per RnBk would fix it. Generally, there could be the same RnBk for a regional and global grid. 

Generally, if there is no special tol like this in the test we take the looser bound, unless it is really bad like this one.

model/testing/src/icon4py/model/testing/fixtures/datatest.py
```diff
@@ -61,7 +61,10 @@ def backend(request: pytest.FixtureRequest) -> gtx_backend.Backend:
 
 
 @pytest.fixture
-def experiment():
+def experiment() -> str:  # TODO(havogt): return definitions.Experiments.MCH_CH_R04B09
```
halungge, line 64: should not this one become the parametrized fixture that you define in some test files directly. 

Even now it gets overwritten by parametrization in the tests and it is probably only here because some other fixture depends on it and/or because I thought in the beginning that you need a fixture in order to overwrite it.

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -11,66 +11,24 @@
...
-GRID_URIS = {
-    REGIONAL_EXPERIMENT: MC_CH_R04B09_DSL_GRID_URI,
-    R02B04_GLOBAL: R02B04_GLOBAL_GRID_URI,
-    R02B07_GLOBAL: R02B07_GLOBAL_GRID_URI,
-    ICON_CH2_SMALL: MCH_OPR_R04B07_DOMAIN01_GRID_URI,
-    REGIONAL_BENCHMARK: DOMAIN01_GRID_URI,
-    WEISMAN_KLEMP_EXPERIMENT: TORUS_50000x5000_RES500_GRID_URI,
-}
 
+# TODO(havogt): is this still needed?
```
halungge, line 31: probably not. We changed the type of the grid.id to `str`, so where it is needed for the grid construction other than the `GridManager` that reads it from the file attributes, you could also use the `Grid.name`, if we want to be more accurate on what it is (and be able to really validate against Grids constructed from the file then we should add it to `definitions.testing.Grid`

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -49,19 +49,11 @@ class Grid:
     description: str
     kind: GridKind
     sizes: Mapping[Literal["cell", "edge", "vertex"], int]
-    file_name: str | None = None
-    uri: str | None = None
+    file_name: str
+    uri: str
 
 
 class Grids:
```
halungge, line 56: Are you removing this one, because it would not work with all the datatest machinery downloading files etc? I would kind of prefer to keep it here as a matter of documentation. But then again: I think the simple.py code is in the wrong place anyway. It should not be in `icon4py.model.common` but in `icon4py.model.testing`...

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -49,19 +49,11 @@ class Grid:
     description: str
     kind: GridKind
     sizes: Mapping[Literal["cell", "edge", "vertex"], int]
-    file_name: str | None = None
-    uri: str | None = None
```
halungge, line 53: would not `GridDescription` be a better name for this?

havogt: Absolutely! Will do the change.

model/testing/src/icon4py/model/testing/datatest_utils.py
havogt: I will think about it in a later PR, should I keep the TODO or remove it?

halungge: I would only remove the `TODO` together with the `GRID_IDS`...

model/common/tests/common/interpolation/unit_tests/test_rbf_interpolation.py
msimberg: > The mapping could be just:
> {(4,9): x, (2,4): y}, doesn't have to be the full function tol(r,b)

Fair enough. Keep in mind that there will be tests with torus grids soon which would require slightly different handling (no r/b parameters).

havogt:
I'll merge and continue the cleanups of test_rbf_interpolation.py in the next PR.

PR #864, opened 2025-09-05
fix format in debug_utils.py

halungge:
(FIX): fixing format in wrapper debug statements: 
do not request a fixed size buffer length which provokes a crash when the str is too long.


muellch (approved):
Looks good, thanks!

PR #866, opened 2025-09-08
fix a bug in grid-savepoint reading during driver initialization

OngChia:
Fix bug in the input arguments of `IconSerialDataProvider.from_grid_savepoint` in `initialization_utils.py` of the driver package after this PR https://github.com/C2SM/icon4py/pull/829 is merged.


model/driver/src/icon4py/model/driver/initialization_utils.py
```diff
@@ -60,36 +66,30 @@ class ExperimentType(str, enum.Enum):
 def read_icon_grid(
     path: pathlib.Path,
     backend: gtx_backend.Backend,
-    rank=0,
+    grid_shape: icon_grid.GridShape,
+    grid_uuid: uuid.UUID = GLOBAL_GRID_ID,
```
halungge, line 70: This should not have a default, as you read the grid file below there is no need for it.

model/driver/src/icon4py/model/driver/initialization_utils.py
```diff
@@ -559,3 +570,33 @@ def configure_logging(
...
+        grid_file: path of the grid file
+
+    Returns:
+        grid_shape: grid shape containing the root, level, and geometry type
+        grid_uuid: id (uuid) of the horizontal grid
+    """
+    grid = nc4.Dataset(grid_file, "r", format="NETCDF4")
+    grid_root = grid.getncattr("grid_root")
+    grid_level = grid.getncattr("grid_level")
+    grid_uuid = uuid.UUID(grid.getncattr("uuidOfHGrid"))
```
halungge, line 591: We changed the type of the id to `str` recently.
```suggestion
    grid_uuid = grid.getncattr("uuidOfHGrid")
```

model/driver/src/icon4py/model/driver/initialization_utils.py
```diff
@@ -559,3 +570,33 @@ def configure_logging(
     console_handler.setFormatter(formatter)
     console_handler.setLevel(logging_level)
     logging.getLogger("").addHandler(console_handler)
+
+
+def create_grid_info(
+    grid_file: str,
+) -> tuple[icon_grid.GridShape, uuid.UUID]:
```
halungge, line 577:
```suggestion
) -> tuple[icon_grid.GridShape, str]:
```

model/driver/src/icon4py/model/driver/initialization_utils.py
```diff
@@ -428,6 +437,8 @@ def read_static_fields(
      Args:
         grid: IconGrid
```
halungge, line 438: If you have the `grid` here already as an argument isnt the GridShape and the uuid available from it in the IconGrid.global_properties`?

model/driver/src/icon4py/model/driver/initialization_utils.py
```diff
@@ -559,3 +570,33 @@ def configure_logging(
     console_handler.setFormatter(formatter)
     console_handler.setLevel(logging_level)
     logging.getLogger("").addHandler(console_handler)
+
+
+def create_grid_info(
```
halungge, line 575: I think it s easiest to read this off the gridfile directly here. Keeping in mind that we really need to rework this driver. All grid file handling should eventually be done in the GridManager, but as we still use the serialized data here, we should keep constructing the grid from it as in some places ordering of local dimension is not identical.

halungge, line 575: maybe call it `create_grid_global_params` and return those directly ?

model/driver/src/icon4py/model/driver/initialization_utils.py
OngChia: I decided to passed in `grid_file` instead of these grid global params and id because they are only used internally by `from_savepoint_grid` in `initialization_utils.py` to generate `IconGrid`, `EdgeParams`, `CellParams`, and other static fields. So, I do not need to pass around these grid global params and id inside  `initialize` function in `icon4py_driver.py`

model/driver/src/icon4py/model/driver/initialization_utils.py
OngChia: Done.

model/driver/src/icon4py/model/driver/initialization_utils.py
OngChia: Okay. I return this global_grid_params and uuid from this function.

OngChia:
cscs-ci run default

PR #868, opened 2025-09-09
Fix exception handling in py2f wrapper

edopao:
Errors in  program build were not visible in the blue-line log because of this error:
```
Exception ignored from cffi callback <function diffusion_init_wrapper at 0x14f1960b2290>:
Traceback (most recent call last):
  File "<init code for 'icon4py_bindings'>", line 569, in diffusion_init_wrapper
NameError: name 'logger' is not defined
```


havogt (approved):
lgtm

PR #871, opened 2025-09-12
Scalar values from factory

halungge:
Adds the ability to produce scalar values from the factories. 
We need this in the metrics factory for the computation of `nflat_gradp` for the dynamical core and will need it as well for the computation of some mean properties for the grid (`mean_cell_area`...) as discussed in [PR-848](https://github.com/C2SM/icon4py/pull/848)

(Additional fixes): fixes a lot of typing issues in all three factory implementations.


model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -103,7 +101,7 @@ def __init__(
         self._grid = grid
         self._decomposition_info = decomposition_info
         self._attrs = metadata
-        self._geometry_type: base.GeometryType = grid.global_properties.geometry_type
+        self._geometry_type: base.GeometryType = base.GeometryType.ICOSAHEDRON
```
msimberg, line 104: Is this a necessary change?

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -64,6 +64,7 @@ def __call__(self, array: data_alloc.NDArray):
 
 
 CoordinateDict: TypeAlias = dict[gtx.Dimension, dict[Literal["lat", "lon"], gtx.Field]]
+# TODO (halungge): use a TypeDict for that
 GeometryDict: TypeAlias = dict[gridfile.GeometryName, gtx.Field]
```
msimberg, line 68: Is the TODO for later or this PR?

model/common/src/icon4py/model/common/grid/horizontal.py
```diff
@@ -420,6 +420,11 @@ def _domain(marker: Zone) -> Domain:
     return _domain
 
 
+vertex_domain = domain(dims.VertexDim)
+edge_domain = domain(dims.EdgeDim)
+cell_domain = domain(dims.CellDim)
+
+
```
msimberg, line 427: Are these used somewhere?

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -93,16 +95,20 @@ class FieldProvider(Protocol):
...
     @property
     def dependencies(self) -> Sequence[str]: ...
 
     @property
-    def fields(self) -> Mapping[str, Any]: ...
+    def fields(
+        self,
+    ) -> Mapping[
+        str, state_utils.FieldType | state_utils.ScalarType
+    ]: ...  # TODO(halungge): properly type value type state_utils.FieldType|state_utils.ScalarType?
```
msimberg, line 111: TODO still relevant?

model/common/tests/common/states/unit_tests/test_factory.py
```diff
@@ -286,3 +289,29 @@ def test_composite_field_source_raises_upon_get_unknown_field(
...
+    return ar.item()
+
+
+@pytest.mark.datatest
+def test_compute_scalar_value_from_numpy_provider(
+    height_coordinate_source: factory.FieldSource,
+    metrics_savepoint: serialbox.MetricSavepoint,
+    backend: gtx_typing.Backend,
+) -> None:
+    value_ref = np.min(np.min(metrics_savepoint.z_ifc()))
```
msimberg, line 306: Why are the two `np.min` needed?

model/common/tests/common/states/unit_tests/test_factory.py
```diff
@@ -286,3 +289,29 @@ def test_composite_field_source_raises_upon_get_unknown_field(
     with pytest.raises(ValueError) as err:
         composite.get("alice")
         assert "not provided by source " in err.value  # type: ignore[operator]
+
+
+def reduce_scalar_min(ar: data_alloc.NDArray, xp: ModuleType) -> gtx.float:
+    while ar.ndim > 0:
+        ar = xp.min(ar)
```
msimberg, line 296: Also here, why is the while loop needed? If I understand correctly, if you leave out the `axis` parameter it should take the minimum over the whole (flattened) array, even if it has more than one dimension.

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -611,13 +623,18 @@ def _compute(
...
         self._fields = {
-            k: gtx.as_field(tuple(self._dims), results[i], allocator=backend)
+            k: self._as_field(backend, results[i]) if self._dims else results[i]
             for i, k in enumerate(self.fields)
         }
 
+    def _as_field(
+        self, backend: gtx_typing.Backend | None, value: data_alloc.NDArray
+    ) -> state_utils.GTXFieldType:
+        return gtx.as_field(tuple(self._dims), value, allocator=backend)
```
msimberg, line 636: Are the RBF tests failing perhaps because I made them (at least the cell and vertex; edge is only one field) return lists instead of tuples (as the type hints would indicate)...? Or should lists of fields as return values be allowed? NB. I'm not asking for it to be allowed, I think restricting it to tuples and single arrays is sufficient.

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -611,13 +623,18 @@ def _compute(
         args.update(offsets)
         args.update(self._params)
         results = self._func(**args)
-        # TODO(): can the order of return values be checked?
-        results = (results,) if isinstance(results, data_alloc.NDArray) else results
+        # convert to tuple
+        results = (results,) if not isinstance(results, tuple) else results
         self._fields = {
-            k: gtx.as_field(tuple(self._dims), results[i], allocator=backend)
+            k: self._as_field(backend, results[i]) if self._dims else results[i]
```
msimberg, line 629: Just for my understanding: Is the idea even that a scalar value returned by `_func` will be converted to a "scalar"/0-d gt4py field? I.e. it does not need to be treated differently because gt4py supports scalar fields?

msimberg (commented):
Mostly questions/minor cleanup comments from my side. I think you'll unfortunately need to fix my RBF computations (the return types).

model/common/src/icon4py/model/common/grid/grid_manager.py
halungge: yes. I can also remove it.

model/common/src/icon4py/model/common/grid/horizontal.py
halungge: I import them `geometry.py`. We have been redefining them all over the place, and I got  a bit tired of that.

model/common/src/icon4py/model/common/states/factory.py
halungge: maybe. I dont know whether it matters at runtime. I never thougth about because we return tuples in gt4py and if you just do a 
```
def foo():
    ...
    return a, b, c
```
that will be a tuple.

model/common/src/icon4py/model/common/grid/geometry.py
msimberg: Yes, good point and fully agree.

I ask because this will clash with #853 which requires actually having this correctly set. I just wanted to understand the motivation, so let's go with this now and I'll get this cleaned up in #853.

halungge:
cscs-ci run default

halungge:
cscs-ci run default

model/common/src/icon4py/model/common/states/factory.py
halungge: I did a tuple([...]). The (...) will return a generator expression, not a tuple...

PR #880, opened 2025-09-22
Fix return types in RBF interpolation functions

msimberg:
https://github.com/C2SM/icon4py/pull/871 made a partial fix, but the types still ended up slightly inconsistent. The generic `_compute_rbf_interpolation_coeffs` returns a dynamically sized list of ndarrays, as it depends on a dynamically sized list as input. The return type was incorrectly specified as a two-element tuple (it can be one or two elements long). #871 correctly made sure the cell/vertex-specific functions return a tuple, but with `_compute_rbf_interpolation_coeffs` returning a list, they return values should be explicitly converted in those functions instead.

### DIFF
--- a/model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
+++ b/model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
@@ -227,7 +227,7 @@ def _compute_rbf_interpolation_coeffs(
     scale_factor: ta.wpfloat,
     horizontal_start: gtx.int32,
     array_ns: ModuleType = np,
-) -> tuple[data_alloc.NDArray, data_alloc.NDArray]:
+) -> list[data_alloc.NDArray]:
     rbf_offset_shape_full = rbf_offset.shape
     rbf_offset = rbf_offset[horizontal_start:]
     num_elements = rbf_offset.shape[0]
@@ -346,7 +346,7 @@ def _compute_rbf_interpolation_coeffs(
             rbf_vec_coeff_np[j][i + horizontal_start, valid_neighbors] = sla.cho_solve(
                 z_diag_np, rhs_np[j][i, valid_neighbors]
             )
-    rbf_vec_coeff = tuple([array_ns.asarray(x) for x in rbf_vec_coeff_np])
+    rbf_vec_coeff = [array_ns.asarray(x) for x in rbf_vec_coeff_np]
 
     # Normalize coefficients
     for j in range(num_zonal_meridional_components):
@@ -399,7 +399,7 @@ def compute_rbf_interpolation_coeffs_cell(
         array_ns=array_ns,
     )
     assert len(coeffs) == 2
-    return coeffs
+    return tuple(coeffs)
 
 
 def compute_rbf_interpolation_coeffs_edge(
@@ -483,4 +483,4 @@ def compute_rbf_interpolation_coeffs_vertex(
         array_ns=array_ns,
     )
     assert len(coeffs) == 2
-    return coeffs
+    return tuple(coeffs)


model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -346,7 +346,7 @@ def index_offset(f):
             rbf_vec_coeff_np[j][i + horizontal_start, valid_neighbors] = sla.cho_solve(
                 z_diag_np, rhs_np[j][i, valid_neighbors]
             )
-    rbf_vec_coeff = tuple([array_ns.asarray(x) for x in rbf_vec_coeff_np])
+    rbf_vec_coeff = [array_ns.asarray(x) for x in rbf_vec_coeff_np]
```
egparedes, line 349: This other way would have kept the original `tuple` type annotation, and since it seems the list is going to be converted to a tuple anyway, probably makes more sense.
```suggestion
    rbf_vec_coeff = tuple(array_ns.asarray(x) for x in rbf_vec_coeff_np)
```

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -227,7 +227,7 @@ def _compute_rbf_interpolation_coeffs(
     scale_factor: ta.wpfloat,
     horizontal_start: gtx.int32,
     array_ns: ModuleType = np,
-) -> tuple[data_alloc.NDArray, data_alloc.NDArray]:
+) -> list[data_alloc.NDArray]:
```
egparedes, line 230: Just for your information, based on your comments and a quick look at the code, it seems that the number of output items depends on the number of elements in the `uv` input argument (either 1 or 2), is that right? In that case, it is actually possible to define this relationship, by using `tuple` instead of list in the `uv` annotation, together with two `overload`s: one of the 1-element tuple case (`uv: tuple[tuple[....]]`) and another for the 2-elements case (`uv: tuple[tuple[....], tuple[....]]`).  More info: https://typing.python.org/en/latest/spec/overload.html

egparedes (commented):
A couple of comments for discussion.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: > Just for your information, based on your comments and a quick look at the code, it seems that the number of output items depends on the number of elements in the uv input argument (either 1 or 2), is that right?

That's right.

> In that case, it is actually possible to define this relationship, by using tuple instead of list in the uv annotation, together with two overloads: one of the 1-element tuple case (uv: tuple[tuple[....]]) and another for the 2-elements case (uv: tuple[tuple[....], tuple[....]]). More info: https://typing.python.org/en/latest/spec/overload.html

Interesting, thanks for the link. On one hand I like the idea since it accurately describes what I'm looking for. On the other I don't like the idea of having the function signature triplicated (except for the one parameter that has a different type). It feels like a lot of noise for not being able to express the type constraints in one function definition. Will ponder on this.

havogt: Welcome to Python. ;)

msimberg: @egparedes FYI assuming I understand the use of typing.overload correctly, this is what it would look like with that change: https://github.com/C2SM/icon4py/commit/4aee35f1dc0804374fa177e5e9ba28aa7539bc0c.

I think even now I still prefer the "dumb list" version, but won't object if you prefer the better typed version. What do you think?

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: > are you aware that tuple[NDArray, ...] also express arbitrary-sized tuples?

I was not! Thanks. In that case I like the option here (without `@overload`, with tuples). Let me try that as well.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
egparedes: I agree it makes more sense (at least now) to use the version without the overloads.

PR #889, opened 2025-10-01
Upgrade dace version

edopao:
The new dace version addresses the following issues:
- Fix in vertical map fusion, see https://github.com/spcl/dace/pull/2157
- Fix error in microphysics, see https://github.com/spcl/dace/pull/2158


edopao:
cscs-ci run default

PR #890, opened 2025-10-01
Write GT4Py timers in dycore_wrapper solve_nh_run

havogt:
This is a quick fix for writing the gt4py timers from the blueline. For a clean implementation we would create separate bindings for this call and write it once at the end of the simulation.


tools/src/icon4py/tools/py2fgen/wrappers/dycore_wrapper.py
```diff
@@ -412,4 +413,8 @@ def solve_nh_run(
         at_last_substep=idyn_timestep == (ndyn_substeps_var - 1),
     )
 
+    # dump gt4py timers
```
havogt, line 416:
```suggestion
    # TODO(havogt): create separate bindings for writing the timers
```

PR #896, opened 2025-10-08
Combine compute_perturbed_quantities_and_interpolation further

muellch:
Further in-lining of stencils 1 to 13 into fewer `field_operator` calls within the program `compute_perturbed_quantities_and_interpolation`.

Along the way:

- `interpolate_cell_field_to_half_levels_vp` and `..._wp` are merged into a single module `interpolate_cell_field_to_half_levels`. It also gains `..._with_surface_value_vp` and `..._with_surface_value_wp`, which combine the interior interpolation with a caller supplied surface value.
- `interpolate_to_surface` is renamed to `extrapolate_quadratically_to_surface`, which is what it does.
- `exner_at_cells_on_half_levels` and `perturbed_theta_v_at_cells_on_half_levels` become program local. Neither is read outside the programs that write them, so they no longer need to be allocated in `SolveNonhydro`. ICON keeps the latter as a block local, thread private array as well.
- The outputs get individual domains. `rho_at_cells_on_half_levels`, `perturbed_theta_v_at_cells_on_half_levels` and `nonhydro_buoy_at_cells_on_half_levels` are computed on owned cells only. `theta_v_at_cells_on_half_levels` keeps the halo because it is read through `E2C` by the hydrostatic correction term, and `perturbed_exner_at_cells_on_model_levels` keeps it because it is read back at the next time step.
- The vertical derivatives of the temporally extrapolated exner function are clamped to the first half level, mirroring `MAX(2, nflatlev)` in `mo_solve_nonhydro.f90`. Computing the top half level needs an upward extrapolation that icon4py does not implement, see #1418.
- `update_mass_flux_weighted` is restricted to the range used in ICON.

The comparisons against the serialized reference are narrowed to owned cells where the corresponding field is no longer computed on the halo. Note that the Fortran verification compares `rho_ic` over a range that includes the halo lines, so a blueline verify run may report differences there until that range is narrowed as well.


nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

muellch (commented):
Looks good aside from one problem.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/extrapolate_temporally_exner_pressure.py
```diff
@@ -24,8 +24,8 @@ def _extrapolate_temporally_exner_pressure(
 
     z_exner_ex_pr_wp = (wpfloat("1.0") + exner_exfac_wp) * (
         exner - exner_ref_mc_wp
-    ) - exner_exfac_wp * exner_pr
-    exner_pr_wp = exner - exner_ref_mc_wp
+    ) - exner_exfac * exner_pr
+    exner_pr_wp = exner - exner_ref_mc
```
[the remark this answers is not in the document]
edopao: I have a PR that will fix this issue: https://github.com/GridTools/gt4py/pull/2396

nfarabullini:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -382,29 +447,29 @@ def compute_perturbed_quantities_and_interpolation(
...
+        perturbed_theta_v_at_cells_on_model_levels=perturbed_theta_v_at_cells_on_model_levels,
         igradp_method=igradp_method,
+        surface_level=surface_level,
         out=(
             temporal_extrapolation_of_perturbed_exner,
             exner_at_cells_on_half_levels,
+            temporal_extrapolation_of_perturbed_exner,
+            perturbed_exner_at_cells_on_model_levels,
+            perturbed_rho_at_cells_on_model_levels,
+            perturbed_theta_v_at_cells_on_model_levels,
```
havogt: `surface_level == nlev+1`, right?
Then these field are out of bounds I guess?

havogt (changes requested):
Can you remind me of what was the goal of this PR? Did we want to combine more such that intermediate values on half or full levels disappear? Maybe that's still to do?

edopao:
The last commit removed a write-write pattern in `compute_perturbed_quantities_and_interpolation`. The optimized SDFG now contains a single access node to `exner_at_cells_on_half_levels`:
Screenshot 2025-12-23 at 11 47 00

This change has two advantages:
1. The new SDFG leads to better performance for `compute_perturbed_quantities_and_interpolation`, see the ~5% additional speedup in the attached plot (the blue bar is always the OpenACC baseline, then 2 bars for icon4py main and the last 2 bars for this PR).
2. By removing the write-write pattern, we can workaround an issue of DaCe state fusion.
bench_blueline_stencil_compute

edopao (approved):
The performance is in pair with baseline, and additional speedup will be provided by #951. This PR is needed to remove a write-write pattern which is not supported by dace state fusion.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -466,16 +433,24 @@ def compute_perturbed_quantities_and_interpolation(
...
         igradp_method=igradp_method,
         nflatlev=nflatlev,
         nflat_gradp=nflat_gradp,
+        surface_level=surface_level,
         out=(
             ddz_of_temporal_extrapolation_of_perturbed_exner_on_model_levels,
             d2dz2_of_temporal_extrapolation_of_perturbed_exner_on_model_levels,
+            perturbed_theta_v_at_cells_on_half_levels,
+            theta_v_at_cells_on_half_levels,
+            exner_at_cells_on_half_levels,
```
havogt: Are these fields actually needed outside of this computation? If not the k range can be changed back to `surface_level - 1` and we can save writes to memory which I believe was the idea of this refactoring.

edopao (dismissed):

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -374,41 +387,11 @@ def compute_perturbed_quantities_and_interpolation(
...
-            temporal_extrapolation_of_perturbed_exner,
-            exner_at_cells_on_half_levels,
-        ),
-        domain={
-            dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
-            dims.KDim: (surface_level - 1, surface_level),
-        },
-    )
 
     _compute_perturbed_quantities_and_interpolation(
```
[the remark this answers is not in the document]
nfarabullini: thanks Edo! I'll do that

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -86,27 +83,41 @@ def _compute_perturbed_quantities_and_interpolation(
...
+        perturbed_exner_at_cells_on_model_levels,
+    ) = _extrapolate_temporally_exner_pressure(
+        exner_exfac=time_extrapolation_parameter_for_exner,
+        exner=current_exner,
+        exner_ref_mc=reference_exner_at_cells_on_model_levels,
+        exner_pr=perturbed_exner_at_cells_on_model_levels,
+    )
+
     exner_at_cells_on_half_levels = (
         concat_where(
```
edopao: This `concat_where` and the ones below are not semantically equivalent, I wonder if there is a motivation. I was expecting something like:

```
 exner_at_cells_on_half_levels = (
    concat_where(
        dims.KDim < surface_level - 1,
        concat_where(
            maximum(1, nflatlev) <= dims.KDim,
            _interpolate_cell_field_to_half_levels_vp(
                wgtfac_c=wgtfac_c, interpolant=temporal_extrapolation_of_perturbed_exner
            ),
            exner_at_cells_on_half_levels,
        ),
        _interpolate_to_surface(
            wgtfacq_c=wgtfacq_c, interpolant=temporal_extrapolation_of_perturbed_exner
        ),
    )
)
```

nfarabullini:
cscs-ci run default

edopao:
I now see a clear speedup:
bench_blueline_stencil_compute

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -116,43 +137,63 @@ def _compute_perturbed_quantities_and_interpolation(
...
+        (
+            _compute_first_vertical_derivative_at_cells(
+                exner_at_cells_on_half_levels, inv_ddqz_z_full
+            )
+        )
+        if igradp_method == horzpres_discr_type.TAYLOR_HYDRO
+        else ddz_of_temporal_extrapolation_of_perturbed_exner_on_model_levels
+    )
+
+    d2dz2_of_temporal_extrapolation_of_perturbed_exner_on_model_levels = (
```
havogt: Note to myself: where is this in Fortran?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -116,43 +137,63 @@ def _compute_perturbed_quantities_and_interpolation(
...
+        -vpfloat("0.5")
+        * (
+            (
+                perturbed_theta_v_at_cells_on_half_levels
+                - perturbed_theta_v_at_cells_on_half_levels(Koff[1])
+            )
+            * d2dexdz2_fac1_mc
+            + perturbed_theta_v_at_cells_on_model_levels * d2dexdz2_fac2_mc
+        )
```
havogt: Should this be a generic d2d2z operator? It seems it could go directly from model_levels to model_levels.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -164,25 +205,35 @@ def _compute_perturbed_quantities_and_interpolation(
...
-    return exner_at_cells_on_half_levels
+    exner_at_cells_on_half_levels = _interpolate_to_surface(
+        wgtfacq_c=wgtfacq_c, interpolant=temporal_extrapolation_of_perturbed_exner
+    )
+
+    return (
+        perturbed_theta_v_at_cells_on_half_levels,
+        astype(theta_v_at_cells_on_half_levels, wpfloat),
+        exner_at_cells_on_half_levels,
+    )
```
havogt: Not used?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -116,43 +137,63 @@ def _compute_perturbed_quantities_and_interpolation(
         reference_theta_at_cells_on_model_levels,
     )
 
-    rho_at_cells_on_half_levels = concat_where(
-        dims.KDim >= 1,
-        _interpolate_cell_field_to_half_levels_wp(wgtfac_c, current_rho),
-        rho_at_cells_on_half_levels,
-    )
+    rho_at_cells_on_half_levels = _interpolate_cell_field_to_half_levels_wp(wgtfac_c, current_rho)
```
havogt: maybe also here the same operator with boundary conditions could be used (although it's boundary value seems never be used, but optimizations should remove it).

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -412,49 +420,56 @@ def compute_perturbed_quantities_and_interpolation(
...
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (1, surface_level - 1),
```
havogt:
```suggestion
                dims.KDim: (model_top + 1, surface_level - 1),
```
I am not sure if I like the namings for the vertical ranges in general. Maybe `model_top` should be a global constant (FrozenNamespace)? Why is `surface_level` referring to nlev+1? It's actually `half_level_surface` and `surface_level - 1` might be just `surface_level`?

havogt (commented):
Mainly questions on how to improve readability, now that boundary conditions close to the interior computations.
Additionally, we can now also combine the last field_operator into this one since we are already using multiple output domains.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -77,30 +80,48 @@ def _compute_perturbed_quantities_and_interpolation(
...
-            (maximum(1, nflatlev) <= dims.KDim),
+            dims.KDim < surface_level - 1,
             _interpolate_cell_field_to_half_levels_vp(
                 wgtfac_c=wgtfac_c, interpolant=temporal_extrapolation_of_perturbed_exner
             ),
-            exner_at_cells_on_half_levels,
+            _interpolate_to_surface(
+                wgtfacq_c=wgtfacq_c, interpolant=temporal_extrapolation_of_perturbed_exner
+            ),
         )
```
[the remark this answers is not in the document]
nfarabullini: we shouldn't change this name

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
nfarabullini: the `concat_where` condition was specifically `dims.KDim >= 1`, hence my assumption that boundary conditions start from 1

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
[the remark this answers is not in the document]
nfarabullini: apologies, let me rephrase: it is my personal opinion that this `concat_where` should stay as is.

nfarabullini:
cscs-ci run default

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
havogt: The `1` refers to 1 above `model_top`. I didn't check who introduced `model_top`, but it would have been more consistent to also replace the `1`s with `model_top+1` when this refactoring was done.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -116,150 +128,94 @@ def _compute_perturbed_quantities_and_interpolation(
...
-        reference_theta_at_cells_on_half_levels + perturbed_theta_v_at_cells_on_half_levels
-    )
-
-    exner_at_cells_on_half_levels = _interpolate_to_surface(
-        wgtfacq_c=wgtfacq_c, interpolant=temporal_extrapolation_of_perturbed_exner
+    (
+        perturbed_rho_at_cells_on_model_levels,
+        perturbed_theta_v_at_cells_on_model_levels,
+    ) = concat_where(
+        (dims.CellDim >= start_cell_halo_level_2) & (dims.CellDim < end_cell_halo_level_2),
```
havogt: We should either
- just drop the concat_where here, or 
- move the bounds into the output domain.

Probably the second is the better choice.

Reason: Assuming both `current_rho` and `current_theta_v` as well as the reference fields are correct in `(start_cell_lateral_boundary_level_3, end_cell_halo)`, then it should be correct to compute it everywhere. Since the now inlined operator `_compute_perturbation_of_rho_and_theta` computed on a smaller domain, I assume the result is only needed in that domain.

havogt: On a second look it seems we are already computing both fields above and there is no change in between. It seems we should delete this call and probably shrink the output domain to level2. (Or keep it the bigger domain as this might be better for fusion in dace, let's check with @edopao.)

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -116,43 +137,63 @@ def _compute_perturbed_quantities_and_interpolation(
...
 
     theta_v_at_cells_on_half_levels = concat_where(
-        dims.KDim >= 1,
+        dims.KDim < surface_level - 1,
         _interpolate_cell_field_to_half_levels_wp(
             wgtfac_c=wgtfac_c_wp, interpolant=current_theta_v
         ),
-        theta_v_at_cells_on_half_levels,
+        reference_theta_at_cells_on_half_levels + perturbed_theta_v_at_cells_on_half_levels,
     )
```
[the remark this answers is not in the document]
havogt: @OngChia we should check again if this could actually be expressed with the same operator including boundary condition (but on wp instead of vp) or not.

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
OngChia: I actually introduced this more "readable" vertical indices to replace `vertical_start` and `vertical_end` because I could not immediately tell what to expect and why for the vertical bounds directly from the old indices , especially when extra slicing or concat_where is applied in the vertical direction inside the field operator. Unless I go to `solve_nonhydro.py` or `velocity_advection.py` and look at the bounds.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -392,80 +277,69 @@ def compute_perturbed_quantities_and_interpolation(
...
-        rho=current_rho,
-        rho_ref_mc=reference_rho_at_cells_on_model_levels,
-        theta_v=current_theta_v,
-        theta_ref_mc=reference_theta_at_cells_on_model_levels,
-        out=(
-            perturbed_rho_at_cells_on_model_levels,
-            perturbed_theta_v_at_cells_on_model_levels,
+        domain=(
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo_level_2),
```
OngChia, line 301: Is it possible to put the bounds as slicing next to those output variables (`perturbed_rho_at_cells_on_model_levels[start_cell_lateral_boundary_level_3: end_cell_halo_level_2, model_top:surface_level-1]`) with the current GT4Py? It seems to look better in that way

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -392,80 +277,69 @@ def compute_perturbed_quantities_and_interpolation(
...
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo_level_2),
```
OngChia, line 305: halo_level_2 bound is needed here because this variable is used for computation of rho and theta at edges with the second order Miura scheme. We will not see these halo bounds when we combine the two programs following this program.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -392,80 +277,69 @@ def compute_perturbed_quantities_and_interpolation(
...
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (model_top + 1, surface_level),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (model_top + 1, surface_level),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
```
OngChia: I think all the `end_cell_halo` up until here can be replaced with `end_cell_local`, possibly the bounds used in the assert in the data test for this stencil may need to be modified because the bounds for these variables will be different from icon. The following three output variables below have to be computed until first halo level because they are involved in the computation of the horizontal derivative of pressure at edges. But they can also be removed if we combine the next two programs.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -112,160 +103,81 @@ def _compute_perturbed_quantities_and_interpolation(
...
-        dims.KDim >= 1,
-        _interpolate_cell_field_to_half_levels_wp(
-            wgtfac_c=wgtfac_c_wp, interpolant=current_theta_v
-        ),
-        theta_v_at_cells_on_half_levels,
+    perturbed_theta_v_at_cells_on_half_levels = (
+        _interpolate_cell_field_to_half_levels_with_surface_value_vp(
+            wgtfac_c=wgtfac_c,
+            interpolant=perturbed_theta_v_at_cells_on_model_levels,
+            surface_value=_extrapolate_quadratically_to_surface(
```
OngChia, line 111: this is a nice refactoring!

OngChia (commented):
I am still reading the changes until half way.

havogt:
Thanks — I went through all four points. Summary of what changed and what didn't.

## `end_cell_halo` → `end_cell_local`

Applied to **three** of the outputs, not all of them:

| output | range | reason |
|---|---|---|
| `rho_at_cells_on_half_levels` | → `end_cell_local` | ✅ |
| `perturbed_theta_v_at_cells_on_half_levels` | → `end_cell_local` | ✅ |
| `nonhydro_buoy_at_cells_on_half_levels` | → `end_cell_local` | ✅ |
| `theta_v_at_cells_on_half_levels` | stays `end_cell_halo` | read through `E2C` |
| `perturbed_exner_at_cells_on_model_levels` | stays `end_cell_halo` | read back next time step |

Your criterion — *"involved in the computation of the horizontal derivative of pressure at edges"* — was the right one; two fields fall under it that weren't in the list of three:

**`theta_v_at_cells_on_half_levels`.** It is read via `E2C` by `compute_hydrostatic_correction_term` (stencil 21), which is part of the horizontal pressure gradient — the correction over steep slopes:

```
theta_v_ic → compute_hydrostatic_correction_term  (theta_v_ic(E2C[0]), theta_v_ic(E2C[1]))
           → hydrostatic_correction_on_lowest_level
           → compute_rho_theta_pgrad_and_update_vn → next_vn
```

That stencil runs over `start_edge_nudging_level_2 .. end_edge_local`, and local edges *do* reach halo cells: in `decomposition/halo.py`, `edge_on_cutting_line` is the set of edges touching both an owned and a halo cell, and `_update_owner_mask_by_max_rank_convention` assigns them to the **higher rank** (matching `mo_decomposition_tools.f90`). So on the winning rank the edge is owned while one `E2C` neighbour is a halo cell.

**`perturbed_exner_at_cells_on_model_levels`.** Different mechanism — it is both an *input* and an *output* of this program. `_extrapolate_temporally_exner_pressure` consumes the previous time step's value and returns the new one, and its sibling output `temporal_extrapolation_of_perturbed_exner` is computed out to `end_cell_halo`. So the halo band of `exner_pr` has to be valid, otherwise the extrapolation on the halo is built from stale data on the next step.

This matches ICON, `mo_solve_nonhydro.f90`:

```fortran
      rl_start = 3
      IF (istep == 1) THEN
        rl_end = min_rlcell_int - 1
      ELSE ! halo points are not needed in step 2
        rl_end = min_rlcell_int
      ENDIF
```

with the read at `z_exner_ex_pr(jc,jk,jb) = ... - exner_exfac(jc,jk,jb) * p_nh%diag%exner_pr(jc,jk,jb)` happening one halo line deep, before `exner_pr` is overwritten for the next step. ICON also explicitly halo-exchanges `exner_pr` after physics (`mo_nh_interface_nwp.f90`, `mo_interface_iconam_aes.f90`).

For `rho_ic` the same check comes out the other way: every production read in ICON is at `rl_end = min_rlcell_int`, there is no neighbour-indexed access to it anywhere, and there is no production halo exchange — the only `sync_patch_array` calls for it are under `itime_scheme == TRACER_ONLY`, which never coexists with `solve_nh`.

## ⚠️ Please run a blueline verify

`rho_at_cells_on_half_levels` is the one to watch, and it is the only one of the three that is not
icon4py-private (it is in `DiagnosticStateNonHydro` and aliased to a Fortran array in granule mode).

ICON *does* write `rho_ic` into the first halo ring (`mo_solve_nonhydro.f90`, `istep==1` branch at
`rl_end = min_rlcell_int - 1`) even though it never reads it back there. And `rho_ic` is in the
VERIFY field set, compared over the **halo-inclusive** range:

```fortran
! mo_icon4py_interfaces.f90:1107-1114
rl_start = 1
rl_end   = min_rlcell          ! -8, includes all halo lines
CALL get_indices_c(p_patch, 1, i_startblk, i_endblk, ic_start, ic_end, rl_start, rl_end)

! :1229-1234
CALL verify_field_wp( &
   field_name='rho_ic', &
   field_ref=p_nh%diag%rho_ic(:,:,1), &
   field=rho_ic_before, &
   i_start=ic_start, &
   i_end=ic_end)
```

Since icon4py writes into `rho_ic_before` and no longer touches the halo band, that band keeps its
pre-call values while `p_nh%diag%rho_ic` gets ICON's freshly computed halo values — a mismatch that
is not a real error. (`theta_v_ic` and `exner_pr` are unaffected: both are still computed on the
halo, which is exactly why they were left at `end_cell_halo` above.)

The harness verifies block 1 (`rho_ic(:,:,1)`, `get_indices_c(p_patch, 1, ...)`), and since icon4py
requires `nblks_c == 1` the whole patch *is* block 1 — so the halo lines are inside the verified
range and this will fire.

The fix is one line — `i_end=ic_end_without_halo` (defined at `:1129` from `rl_end = min_rlcell_int`),
exactly as `ddt_w_adv_nnow`/`ddt_w_adv_nnew` already do two entries above at `:1215-1222`. Only
`i_end` changes: `ic_start_without_halo` would also move the start to `grf_bdywidth_c+1` and drop the
nest-boundary band, which both ICON (`rl_start = 3` in the predictor) and icon4py do compute.

That one-line ICON change is prepared on a local branch and still needs to be tested and turned into
an ICON MR.

## Test ranges adjusted

- `test_solve_nonhydro.py::test_compute_perturbed_quantities_and_interpolation` — `rho_ic` and `z_theta_v_pr_ic` now compared over `[lb:end_cell_local]`; the reference still holds ICON's halo values.
- `test_compute_perturbed_quantities_and_interpolation.py` — the numpy reference had three `np.where` blocks that each wrote a *mix* of now-local and still-halo outputs, so those had to be split per output.
- `test_parallel_solve_nonhydro.py` — `rho_ic` comparison sliced to `end_cell_local` (`theta_v_ic` left full-array, since it is still computed on the halo). **I could not run this one** — it needs MPI plus multi-rank datatest data.

Worth being explicit about test coverage: on serial grids `end_cell_local == end_cell_halo == num_cells`
(checked: 20896 for MCH_CH_R04B09, 20480 for EXCLAIM_APE, and 18 on the simple grid). So **no serial
test can detect this change** — they pass either way. The only meaningful coverage is
`test_parallel_solve_nonhydro.py` and a blueline verify run.

## Re: bounds as slicing next to the output variables

Nice idea, but GT4Py doesn't allow it today. `out=field[start:end, ...]` requires **compile-time constant** slice bounds — `past.Slice.lower/upper` is typed `Optional[Constant]`, and `past_to_itir._visit_slice_bound` branches on the sign of the value at trace time to decide whether the bound is relative to the start or the end of the field:

```python
if slice_bound.value < 0:
    lowered_bound = im.plus(stop_idx, ...)   # relative to end
else:
    lowered_bound = im.plus(start_idx, ...)  # relative to start
```

Our bounds are runtime `gtx.int32` program parameters, so the sign isn't known and it fails with `TypeError: 'Slice.lower' must be ... Constant`. That's why the existing example (`calculate_diagnostics_for_turbulence.py`, `out=(div_ic[:, 1:], hdef_ic[:, 1:])`) uses literals. Also worth knowing: slice bounds are relative to the field's domain start, whereas `domain=` bounds are absolute — so the two spellings aren't interchangeable in general.

Lifting this in GT4Py looks contained (relax the type, and either require non-negative or give relative-to-end its own spelling), but that's a separate GT4Py change.

## Re: the `halo_level_2` bound and the vertical index naming

The `halo_level_2` bound on `perturbed_rho/theta_v_at_cells_on_model_levels` is unchanged — agreed on the Miura reasoning.

The `model_top` / `surface_level` naming discussion is still open from January; leaving that as is for now.

havogt:
cscs-ci run default

havogt (changes requested):
Just blocking the merge until blueline is verified: need to change verification range for rho_ic.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -392,80 +278,73 @@ def compute_perturbed_quantities_and_interpolation(
...
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
+                dims.KDim: (nflatlev, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo),
```
OngChia, line 342: From readability perspective, can we directly perform slicing, for example  `temporal_extrapolation_of_perturbed_exner[start_cell_lateral_boundary_level_3:end_cell_halo, nflat_gradp, surface_level]`,  instead of passing a separate dictionary?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -392,80 +278,73 @@ def compute_perturbed_quantities_and_interpolation(
...
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo_level_2),
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            {
+                dims.CellDim: (start_cell_lateral_boundary_level_3, end_cell_halo_level_2),
+                dims.KDim: (model_top, surface_level - 1),
+            },
+            # `perturbed_exner_at_cells_on_model_levels` is read back at the next time step to
+            # compute `temporal_extrapolation_of_perturbed_exner`, which is needed on the halo.
```
OngChia, line 309: I am wondering whether we need a comment when we see a variable is updated until the halo region.
I think I would prefer to remove it, but without strong objection because it can remind me of the reason. If keeping the comment, I will also add comments to those output variables below whose dimension extends to the halo region for consistency.

OngChia (approved):
LGTM. I think it is a good improvement with the new feature in terms of readability. Hopefully we can combine more and the code will look more similar to what we see in the equation (@havogt Did we discuss something about a bound, right? But now I can't remember exactly the issue..). I left two non-blocking questions.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
```diff
@@ -112,160 +103,81 @@ def _compute_perturbed_quantities_and_interpolation(
...
+            + perturbed_theta_v_at_cells_on_model_levels * d2dexdz2_fac2_mc
+        )
+
     return (
         perturbed_rho_at_cells_on_model_levels,
         perturbed_theta_v_at_cells_on_model_levels,
         perturbed_exner_at_cells_on_model_levels,
         rho_at_cells_on_half_levels,
-        exner_at_cells_on_half_levels,
         perturbed_theta_v_at_cells_on_half_levels,
```
OngChia: Wait. Maybe I am wrong. It looks like `perturbed_theta_v_at_cells_on_half_levels` is only a temporary variable. We can remove it from the outputs and in `solve_nonhydro.py`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
havogt: not, at this point. Certainly we need to improve the syntax here. I'll check how much it would be to do that in gt4py.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
havogt: I added them for now. I usually don't like this kind of comments, but maybe they help short term in further cleanups.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_cell_diagnostics_for_dycore.py
havogt: I removed it.

havogt:
cscs-ci run default

havogt (approved):

PR #903, opened 2025-10-14
Remove nlev from decomposition info

halungge:
(Refactoring) remove `klevels` from `DecompositionInfo` it does not seem to be used.

based on this [PR-901](https://github.com/C2SM/icon4py/pull/901) which is a prerequisit to run the tests.

### DIFF
--- a/model/common/tests/common/decomposition/unit_tests/test_definitions.py
+++ b/model/common/tests/common/decomposition/unit_tests/test_definitions.py
@@ -27,7 +27,6 @@ from icon4py.model.testing.fixtures.datatest import (  # import fixtures form te
 @pytest.mark.datatest
 def test_create_single_node_runtime_without_mpi(icon_grid, processor_props):
     decomposition_info = DecompositionInfo(
-        klevels=10,
         num_cells=icon_grid.num_cells,
         num_edges=icon_grid.num_edges,
         num_vertices=icon_grid.num_vertices,


model/atmosphere/dycore/tests/dycore/mpi_tests/test_parallel_solve_nonhydro.py
```diff
@@ -6,57 +6,67 @@
...
+    processor_props: definitions.ProcessProperties,
+    decomposition_info: definitions.DecompositionInfo,  # : F811 fixture
+    backend: gtx_typing.Backend | None,
+) -> None:
     parallel_helpers.check_comm_size(processor_props)
     print(
         f"rank={processor_props.rank}/{processor_props.comm_size}: inializing dycore for experiment 'mch_ch_r04_b09_dsl"
     )
     print(
-        f"rank={processor_props.rank}/{processor_props.comm_size}: decomposition info : klevels = {decomposition_info.klevels} "
```
msimberg: Same here, keep rank/comm_size?

model/atmosphere/dycore/tests/dycore/mpi_tests/test_parallel_solve_nonhydro.py
```diff
@@ -204,11 +178,6 @@ def test_run_solve_nonhydro_single_step(
         prognostic_states.next.rho.asnumpy(),
     )
 
-    assert test_utils.dallclose(
-        savepoint_nonhydro_exit.rho_ic().asnumpy(),
-        diagnostic_state_nh.rho_ic.asnumpy(),
-    )
-
```
msimberg: I'm missing some context here: why does this not need to be checked?

model/common/tests/common/grid/mpi_tests/test_parallel_icon.py
```diff
@@ -124,22 +144,30 @@ def test_distributed_local(
...
     rank = processor_props.rank
     print(
-        f"rank {rank}/{processor_props.comm_size} dim = {dim}  {marker} : ({start_index}, {end_index})"
+        f"rank {rank}/{processor_props.comm_size} dim = {dim}  {zone} : ({start_index}, {end_index})"
     )
-
-    assert start_index == HALO_IDX[processor_props.comm_size][dim][rank][num - 1]
-    assert end_index == HALO_IDX[processor_props.comm_size][dim][rank][num]
+    expected = HALO_IDX[processor_props.comm_size][dim][rank][level - 1]
+    assert start_index == expected, f"expected start index {expected}, but was {start_index}"
```
msimberg: Just curious about this change: does the `assert` not print the values (including which one is `start_index` and `expected`) on failure?

msimberg (changes requested):
If I understand correctly the motivation to remove it is simply to clean things up I think that's nice. It seems to semantically not belong into the decomposition (since there's no vertical decomposition).

Type hints are welcome (though a bit unrelated, but still very good).

I see some of the tests were expanded. Just for my understanding, these are also not strictly related, but increase coverage (I don't necessarily mean in the technical lines-of-code sense, just in the generic more-test-cases sense)?

model/atmosphere/diffusion/tests/diffusion/mpi_tests/test_parallel_diffusion.py
```diff
@@ -53,10 +66,9 @@ def test_parallel_diffusion(
         f"rank={processor_props.rank}/{processor_props.comm_size}: initializing diffusion for experiment '{definitions.Experiments.MCH_CH_R04B09}'"
     )
     print(
-        f"rank={processor_props.rank}/{processor_props.comm_size}: decomposition info : klevels = {decomposition_info.klevels}, "
```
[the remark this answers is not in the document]
halungge: Sorry, I don't understand this comment.

nfarabullini (approved):

model/atmosphere/dycore/tests/dycore/mpi_tests/test_parallel_solve_nonhydro.py
halungge: I removed the check because it has been removed from the single node test. Then I assumed it that it is not computed in the dycore anymore, but I didn't check wether this is true.

model/common/tests/common/grid/mpi_tests/test_parallel_icon.py
halungge: not as such, only if you have `assert condition, error_msg`. The error_msg is optional.

halungge:
> If I understand correctly the motivation to remove it is simply to clean things up I think that's nice. It seems to semantically not belong into the decomposition (since there's no vertical decomposition).
> 
> Type hints are welcome (though a bit unrelated, but still very good).
> 
> I see some of the tests were expanded. Just for my understanding, these are also not strictly related, but increase coverage (I don't necessarily mean in the technical lines-of-code sense, just in the generic more-test-cases sense)?

Yes, there is no decomposition along `k` so if it is not needed for GHEX, it should not be there. Also it think it will allow us to a make the entire grid_manager independent of `k` which is nice, as it has nothing to do with the vertical... There is generally no relation between vertical and horizontal grid in ICON. 

The test expansion that you mention is something I did in [PR-901] (https://github.com/C2SM/icon4py/pull/901) where this one is based on. It seems that Nikki merged that one already...

model/atmosphere/diffusion/tests/diffusion/mpi_tests/test_parallel_diffusion.py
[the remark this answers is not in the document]
msimberg: The line used to print
- comm size
- comm rank
- klevels

I understand that klevels isn't printed anymore (because it's not in  decomposition info anymore). But the whole line is removed, so I was wondering if the comm size/rank should be kept, but...

I see now my confusion: the comm size/rank is added as a prefix on some of the output and is printed elsewhere so it makes sense to remove the whole line.

Sorry for the noise.

msimberg (approved):
Thanks @halungge for the explanations! The diff looks a lot more sane with #901 merged.

I left one more comment on the readme changes, but approving as it looks good to me otherwise.

PR #904, opened 2025-10-15
Edit plot script for blueline performance

edopao:


scripts/compare_icon_icon4py.py
iomaganaris:
```suggestion
    "interpolate_rho_theta_v_to_half_levels_and_compute_pressure_buoyancy_acceleration": None,
    "update_mass_flux_weighted": None,
```
By adding an `n` at the end of `"interpolate_rho_theta_v_to_half_levels_and_compute_pressure_buoyancy_acceleratio"` in `"bencher=exp.mch_icon-ch1_medium_stencils=0.373574=ACC.json"` we can also compare this stencil. Not sure how I should update the `json` file with the correct name in the cloud though. I don't know how you use it locally

PR #905, opened 2025-10-15
solve_nh: skip call to init_test_fields

edopao:


edopao:
cscs-ci run default

muellch:
@OngChia @halungge In what circumstance is this `init_test_fields` used? Can we delete it or just activate it for certain tests?

PR #907, opened 2025-10-15
Halo exchanges for factories

nfarabullini:
Addition of `exchange: decomposition.ExchangeRuntime` in fields factories and geometry to allow halo exchange for the static field computations. 

Subsequent implementation of factories parallel tests

### DIFF
--- a/model/common/tests/common/metrics/mpi_tests/test_parallel_metrics.py
+++ b/model/common/tests/common/metrics/mpi_tests/test_parallel_metrics.py
@@ -92,7 +92,7 @@ def test_distributed_metrics_attrs(
         (attrs.EXNER_REF_MC, "exner_ref_mc"),
         (attrs.RHO_REF_ME, "rho_ref_me"),
         (attrs.THETA_REF_ME, "theta_ref_me"),
-        (attrs.D2DEXDZ2_FAC1_MC, "d2dexdz2_fac1_mc"), #1-e-12
+        (attrs.D2DEXDZ2_FAC1_MC, "d2dexdz2_fac1_mc"),
         (attrs.D2DEXDZ2_FAC2_MC, "d2dexdz2_fac2_mc"),
         (attrs.DDXN_Z_FULL, "ddxn_z_full"),
         (attrs.DDXT_Z_FULL, "ddxt_z_full"),


nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -45,6 +45,7 @@ def __init__(
         geometry_source: geometry.GridGeometry,
         backend: gtx_typing.Backend | None,
         metadata: dict[str, model.FieldMetaData],
+        exchange: decomposition.ExchangeRuntime = decomposition.SingleNodeExchange(),  # noqa: B008
```
halungge: same here for the fixing `B008`.

model/common/tests/common/grid/mpi_tests/test_parallel_geometry.py
```diff
@@ -0,0 +1,80 @@
...
+@pytest.mark.mpi
+@pytest.mark.parametrize("processor_props", [True], indirect=True)
+def test_distributed_dual_normal_cell(
+    backend: gtx_typing.Backend,
+    grid_savepoint: sb.IconGridSavepoint,
+    experiment: test_defs.Experiment,
+    processor_props: decomposition.ProcessProperties,
+) -> None:
+    parallel_helpers.check_comm_size(processor_props)
+    grid_geometry = grid_utils.get_grid_geometry(backend, experiment)
```
halungge: You need to setup the geometry here passing an explicit `ExchangeRuntime` argument, as you dont want to use the default that you defined int the constructor of the `Geometry`. There are two implementations of the the `ExchangeRuntime` one for single node and one for multinode, you need the multinode version here, which is defined in `decomposition.mpi_decomposition.py`.

you can create this multi node aware exchange object with `create_exchange(props: ProcessProperties, decomp_info: DecompositionInfo)` (see test_parallel_diffusion.py` for an example.

nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -126,6 +127,7 @@ class FieldSource(GridProvider, Protocol):
     """
 
     _providers: MutableMapping[str, FieldProvider] = {}  # noqa:  RUF012 instance variable
+    _exchange: decomposition.ExchangeRuntime = decomposition.SingleNodeExchange()
```
halungge: you could use the predefined singleton here as well.

halungge (changes requested):
For me, still some tests do not validate. I have not explored further which ones these are. It could be that they need some bounds because the some local boundaries where neighbors are unknown.
What did you do for the 3d fields that we encountered? can you check how many there are and whether they need to be exchanged (maybe you did that already)

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/exceptions.py
```diff
@@ -22,3 +24,7 @@ def __init__(self, field_name):
 
 class IconGridError(RuntimeError):
     pass
+
+
+def dummy_exchange(dim: gtx.Dimension, field: gtx.Field) -> None:
```
halungge: that does not belong to `exceptions.py`: if you want to centralize it then put it somewhere in the `testing` for example `test_utils.py`

model/common/tests/common/interpolation/mpi_tests/test_parallel_interpolation.py
```diff
@@ -0,0 +1,207 @@
...
+@pytest.mark.mpi
+@pytest.mark.parametrize("processor_props", [True], indirect=True)
+@pytest.mark.parametrize(
+    "attrs_name, intrp_name, index",
+    [
+        (attrs.GEOFAC_GRG_X, "geofac_grg", 0),
+        (attrs.GEOFAC_GRG_Y, "geofac_grg", 1),
+    ],
+)
+def test_distributed_interpolation_grg(
```
halungge: If you separate out the test then I would not split the x, y components. They are computed together anyway, you can simply assert them both in the same test case, that saves one computation because the factories are setup from scratch in each test function (as the fixtures have default scope)

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -339,6 +339,12 @@ def index_offset(f):
     z_rbfmat_np = data_alloc.as_numpy(z_rbfmat)
     rhs_np = [data_alloc.as_numpy(x) for x in rhs]
     for i in range(num_elements):
+        # convert repeated indices back into -1
```
halungge: I don't know whether we should do that here, as it fixes a problem from the test input data (the connectivities from the serialization) or rather do it in the test input prior to the tests. It for sure needs some more context as why we do it here. Maybe @msimberg  has an opionion on this.

halungge (commented):
Some small things, and the tests where you restrict to one experiment, should be for the other really. I changed the configuration for some few fields.

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -251,6 +255,32 @@ def exchange_and_wait(self, dim: gtx.Dimension, *fields: gtx.Field) -> None:
...
+        assert buffers[0].ndim in (1, 2), "Buffers must be 1d or 2d"
+        assert (
+            field_dims[0] in dims.MAIN_HORIZONTAL_DIMENSIONS.values()
+        ), f"first dimension must be one of ({dims.MAIN_HORIZONTAL_DIMENSIONS.values()})"
+        shape = buffers[0].shape
+        for b in buffers:
+            assert b.shape == shape, "All buffers must have the same shape"
+        domain = {d: (0, shape[i]) for i, d in enumerate(field_dims)}
+        # uses gt4py internal API
+        wrapped = (gtx_common._field(b, domain=domain) for b in buffers)
```
havogt: I don't think we need to do this if we implement this function here, see `_get_applied_pattern` where we extract the buffer from the field. My perspective:
- either consider exchanging numpy/cupy buffers a workaround, then I would recommend a free function that doesn't pollute the interface of this class, or
- make it a first class feature and implement this function directly on buffers.
Or do I miss something?

nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
havogt: Yes, I understand that part, but I don't see the need to wrap it in a field here, because we can directly work on the buffers.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
halungge: "correct" is not really a term that I would use in that context without definition. Either `ICON` is correct, then the serialized data is correct, or the gridfile is `correct`: then  are the two gridfile generators even doing the same thing? ... In any case: _I_ would say: we should keep the invalid indices and deal with them explicitly as for the order I think we should not assume that they are contiguous, because there are some ordering conventions for local neighborhoods in ICON like "vertices in cells being order counterclockwise", these are not very transparent (to me) and I don't know whether ICON keeps them consistently by exactly doing things like moving missing indices to the end.

halungge (approved):

halungge:
cscs-ci run default

PR #908, opened 2025-10-15
Cleanups in factories

halungge:
Cleanups in the factories:
- delete registration of `vert_out` field in `MetricsFactory` as it is never used
- delete the computation of `pos_on_tplane_e` for the edge vertices which was not returned and not tested and seems to be nowhere used even in ICON.
- add new `experiment` fixture in `test_interpolation_fields.py` where missing


halungge:
cscs-ci run default

PR #914, opened 2025-10-21
Remove slicing of fields in exchange

havogt:
TODO: add caching of the constructed patterns


havogt:
Not sure anymore if this is a good strategy as it would disallow exchanging the non-overallocated fields (which we don't have now, but could have in the future).

PR #915, opened 2025-10-27
Cache applied ghex patterns

havogt:


model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -210,33 +212,31 @@ def _slice_field_based_on_dim(self, field: gtx.Field, dim: gtx.Dimension) -> dat
         else:
             raise ValueError(f"Unknown dimension {dim}")
 
+    def _get_applied_pattern(self, dim: gtx.Dimension, f: gtx.Field):
+        # TODO(havogt): the cache is never cleared, consider using functools.lru_cache in a bigger refactoring.
+        key = f.__gt_buffer_info__.hash_key
```
egparedes: Shouldn't the key also contain the dimension?
```suggestion
        key = (dim, f.__gt_buffer_info__.hash_key)
```

havogt: It looks like, but actually no.
`dim` is kind of only a safety check. I tried to remove it, but then realized I need to touch more things. Basically the current interface only allows to exchange fields on the same dim in a call to exchange. Therefore the interface forces you to provide that dimension. But we could just get the dimension from the field...
See also the assert a few lines below.

egparedes (approved):
LGTM

PR #916, opened 2025-10-27
Rearrange small programs in dycore

msimberg:
This moves two small programs (`scale_k` and `_compute_rayleigh_damping_factor`) a bit closer to where they are actually needed. The kernels for these programs are very quick to run, but the program setup takes a non-negligible amount of time. The two programs currently run first in diffusion and the predictor step (unless first substep) which means that the program setup latency isn't hidden at all.

This is a bandaid.

- The programs could potentially be moved to other places, I've chosen one place where the latency is hidden.
- @havogt mentioned that @muellch might remove `scale_k`. Should we do that instead? Remove it after this PR?
- Even better would be to merge programs into bigger programs; this sort of manual scheduling shouldn't be required.


muellch:
I am currently working on removing both calls.
If you want you can merge this PR, but you can also close and discard it.

PR #936, opened 2025-11-07
Configuration 

halungge:
Adds Configuration for ICON4Py in two parts:
- `common` provides a `ConfigurationHandler` for initialization of configuration for a `@dataclass` or a yaml file. This one can be used in each model component. There are two examples for this in `model.atmosphere.diffusion` and `model.atmosphere.dycore`
- `ConfigurationManger` in `model.driver` which is supposed to be used in a model run. It reads a yaml file and initializes a `ConfigurationHandler` for each model registered in the model config and makes the configuration available (as dictionary) to the model.


model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/config.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+
+    NO_DIFFUSION = -1  #: no diffusion
+    LINEAR_2ND_ORDER = 2  #: 2nd order linear diffusion on all vertical levels
+    SMAGORINSKY_NO_BACKGROUND = 3  #: Smagorinsky diffusion without background diffusion
+    LINEAR_4TH_ORDER = 4  #: 4th order linear diffusion on all vertical levels
+    SMAGORINSKY_4TH_ORDER = 5  #: Smagorinsky diffusion with fourth-order background diffusion
+
+
+class TurbulenceShearForcingType(int, enum.Enum):
```
egparedes:
```suggestion
class TurbulenceShearForcingType(enum.IntEnum):
```

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/config.py
```diff
@@ -0,0 +1,177 @@
...
+class NonHydrostaticConfig:
+    """
+    Contains necessary parameter to configure a nonhydro run.
+
+    Encapsulates namelist parameters and derived parameters.
+    Default values are taken from the defaults in the corresponding ICON Fortran namelist files.
+    """
+
+    # number of dynamics substeps
+    ndyn_substeps: int = dataclasses.field(
```
OngChia: In contrast to `DiffusionConfig`, we do not need `ndyn_substeps` for `NonHydrostaticConfig` because stencils in `SolveNonhydro` need the runtime changeable `ndyn_substeps_var`.

model/driver/src/icon4py/model/driver/config/configuration_manager.py
```diff
@@ -0,0 +1,147 @@
...
+                log.warning(
+                    f"no config module in package {package}, instantiating general dict-config instead"
+                )
+                reader = common_config.init_config()
+                if module_config is None:
+                    log.warning(
+                        f"No default configuration and no user-configuration for component {name} defined in file {self._config_file}"
+                    )
+                else:
+                    reader.update(module_config)
```
OngChia: It may be true in the if statement above that users have developed a component without any config parameters. 
However, I am wondering that should we just throw an error here in this case where users put a config for a particular component whereas there is in `config.py`.

OngChia (commented):
Thanks a lot for the configuration infrastructure!
It looks awesome. Currently, I have a few questions on the configuration manager.

DropD:
cscs-ci run default

msimberg:
@DropD let's close this? I guess it has (and still will) served as inspiration, but we're definitely not merging this specific PR.

PR #938, opened 2025-11-10
Use experimental GHEX async scheduling

havogt:
Note: This version of GHEX avoids the sync in icon4py that is needed with GHEX v0.4.1, giving a huge boost in performance. The minimum version of ghex is set to 0.5, which is not currently released, but might become the version with the async scheduling feature.

PR #941, opened 2025-11-12
GHEX update

halungge:
change GHEX to latest release version on PyPI, 
use ghex github master…in uv.sources


egparedes (dismissed):
I agree with @havogt about merging #938 instead, but I would add there the change in the common `pyproject.toml` to bump the minimal required version of ghex to the latest version available in PyPi: "ghex>=0.4.1"

egparedes (commented):
I approved earlier by mistake, we should merge #938 instead

PR #942, opened 2025-11-13
Update to GT4Py v1.1.1 and fix bool(Field) bugs

havogt:


havogt:
cscs-ci run default

havogt:
cscs-ci run default

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -563,7 +563,7 @@ def __call__(
         backend: gtx_typing.Backend | None,
         grid: factory.GridProvider,
     ) -> state_utils.GTXFieldType | None:
-        if not self._fields.get(field_name):
+        if self._fields.get(field_name, None) is None:
```
egparedes: `None` is already the default fallback balue, it doesn't need to be passed.
```suggestion
        if self._fields.get(field_name) is None:
```

PR #944, opened 2025-11-13
Fix e flx average or e2c2c

halungge:
(FIX): the computation of e_flx_avg from the factory did not verify and was not tested for the factory computation, the simple function test against serialized data verfied. 
We could track it down to being somehow dependent on the local ordering of the diamond connectivities being exactly in the order that icon uses.

For simplicity we fixed the ordering of e2c2e in the grid manager.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/grid_manager.py
+++ b/model/common/src/icon4py/model/common/grid/grid_manager.py
@@ -518,8 +518,8 @@ def _construct_diamond_edges(
     flattened = expanded.reshape(n_edges, n_e2c * n_c2e)
 
     centers = array_ns.arange(n_edges, dtype=gtx.int32)[:, None]
-    me_cell1 = _determine_center_position(centers, expanded[:, 0, :])
-    me_cell2 = _determine_center_position(centers, expanded[:, 1, :])
+    me_cell1 = _determine_center_position(centers, expanded[:, 0, :], array_ns=array_ns)
+    me_cell2 = _determine_center_position(centers, expanded[:, 1, :], array_ns=array_ns)
     ordered_local_index = array_ns.hstack(
         (icon_edge_order[me_cell1], icon_edge_order[me_cell2] + n_c2e)
     )


halungge:
cscs-ci run extra

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```suggestion
    e_flx_avg = array_ns.zeros([e2c.shape[0], c2e.shape[1]+e2c.shape[1]])
```
I would edit this 5 to a more generic notation, in case we will use completely different grids later on

```suggestion
    inv_neighbor_id = -array_ns.ones([c2e.shape[0] - llb, c2e.shape[1]], dtype=int)
```

halungge:
cscs-ci run default

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -760,24 +762,20 @@ def compute_e_flx_avg(
             )
 
     iie = -array_ns.ones([e2c.shape[0], 4], dtype=int)
```
```suggestion
    iie = -array_ns.ones(e2c2e.shape, dtype=int)
```

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -489,23 +489,30 @@ def _construct_diamond_edges(
...
     """
+    # used to make sure that the local neighborhood is ordered in the same way as in ICON. At least
+    # the compute_e_flx_avg function depends on that.
+    icon_edge_order = array_ns.asarray([[1, 2], [2, 0], [0, 1]])
+
+    def _determine_center_position(
+        centers: data_alloc.NDArray, neighbors: data_alloc.NDArray
+    ) -> data_alloc.NDArray:
+        center_idx = array_ns.where(neighbors == centers)
+        me_cell = array_ns.zeros(centers.shape[0], dtype=gtx.int32)
```
OngChia: Does the initial value of e2c2e matter along the boundary edges where there are no neighboring edges on one side of the diamond shape? Should we initialize it with `INVALID_INDEX`?
`me_cell = array_ns.full(centers.shape[0], fill_value=gridfile.GridFile.INVALID_INDEX, dtype=gtx.int32)`

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -489,23 +489,30 @@ def _construct_diamond_edges(
     Returns: ndarray containing the connectivity table for central edge-to- boundary edges
              on the diamond
     """
+    # used to make sure that the local neighborhood is ordered in the same way as in ICON. At least
+    # the compute_e_flx_avg function depends on that.
+    icon_edge_order = array_ns.asarray([[1, 2], [2, 0], [0, 1]])
```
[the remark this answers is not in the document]
halungge: Agree, understanding the pattern would maybe also help fixing the computation `for e_flx_avg` where they have a similar set of magic indices, that we failed to understand. I am sure they are correlated...

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -694,7 +698,9 @@ def compute_e_flx_avg(
     array_ns: ModuleType = np,
 ) -> data_alloc.NDArray:
     """
-    Compute edge flux average
+    Compute edge flux average.
```
OngChia: I always have a hard time understanding the logic from the code. And there seems to be no documentation about these interpolation fields in the Fortran ICON... (answer from Anurag). Not now, but I will add equations and docstring to explain these computations so that users or other developers in the future can understand the logic behind these computations better and facilitate debugging. I have already derived some of them, I will take time to finish all.

halungge: The more we have.... so yes please. Feel free to delete any docstring that there is and that is either wrong or not helpful. Also are these attributes (like `interpolation_attributes.py`  which contain metadata for these fields. If you understand what the field actually is you could fix the (CF like) `long_name`  and `standard_name` there... (We can also  make that a little project for you (and @jcanton )...)

halungge:
cscs-ci run default

PR #948, opened 2025-11-17
Use semantically correct bool comparison for some metrics field tests

msimberg:
Just extracts a minor inconsistency from #853. Also fixes dtype for mask_hdiff.

### DIFF
--- a/model/common/src/icon4py/model/common/metrics/metrics_attributes.py
+++ b/model/common/src/icon4py/model/common/metrics/metrics_attributes.py
@@ -429,7 +429,7 @@ attrs: dict[str, model.FieldMetaData] = {
         units="",
         dims=(dims.CellDim, dims.KDim),
         icon_var_name="mask_hdiff",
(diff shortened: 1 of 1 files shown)


model/common/tests/common/metrics/unit_tests/test_metrics_factory.py
```diff
@@ -490,7 +490,7 @@ def test_factory_mask_bdy_prog_halo_c(
     )
     field_1 = factory.get(attrs.MASK_PROG_HALO_C)
     field_2 = factory.get(attrs.BDY_HALO_C)
-    assert test_helpers.dallclose(field_ref_1.asnumpy(), field_1.asnumpy())
+    assert (field_ref_1.asnumpy() == field_1.asnumpy()).all()
```
nfarabullini: what about:
```
np.all(np.logical_and(field_ref_1.asnumpy(), field_1.asnumpy()))
```

PR #951, opened 2025-11-19
Combine divergence damping coefficient calculation with other stencils

muellch:


edopao:
I see a speedup in `compute_advection_in_vertical_momentum_equation` using the dace backend, it could be related to the watermark of the memory pool or some other side-effect of not launching the `divdamp` kernel:
bench_blueline_stencil_compute

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
```diff
@@ -298,7 +297,23 @@ def _apply_divergence_damping_and_update_vn(
     limited_area: bool,
     apply_2nd_order_divergence_damping: bool,
     apply_4th_order_divergence_damping: bool,
+    interpolated_fourth_order_divdamp_factor: fa.KField[ta.wpfloat],
```
edopao: We should probably move this argument close to the other fields.

edopao:
Please check if you agree with this change, from an application-code point of view. If you agree, the next step is to update the serialized data to fix the failing test.

muellch (commented):
Looks good!
I cannot approve since I am the original PR author.

edopao:
cscs-ci run default

PR #953, opened 2025-11-19
Add torus support to GridGeometry

msimberg:
This PR continues the series of PRs to add torus grid support. This is extracted from #853.

This PR:
- Adds two subclasses, `IcosahedronGridGeometry` and `TorusGridGeometry`, of `GridGeometry` that handle initializing geometry fields for the two grid types.
- Adds a factory method to `GridGeometry`, `with_geometry_type` (bikeshedding on name and placement welcome) which initializes `GridGeometry` based on `geometry_type`. With a closed set of geometry types, this could possibly also be nicely solved with branching inside a single `GridGeometry` class, but I think the current setup isn't bad either.
- Adds tests for the geometry fields.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/geometry.py
+++ b/model/common/src/icon4py/model/common/grid/geometry.py
@@ -667,8 +667,6 @@ class TorusGridGeometry(GridGeometry):
         inverse_dual_edge_length = self._inverse_field_provider(attrs.DUAL_EDGE_LENGTH)
         self.register_provider(inverse_dual_edge_length)
 
-        # TODO(msimberg): Is this needed for torus? Even the serialized data
-        # doesn't seem to make sense. Earth radius is used even for the torus.
         vertex_vertex_distance = factory.ProgramFieldProvider(
             func=stencils.compute_distance_of_far_edges_in_diamond_torus,
             domain={


model/common/src/icon4py/model/common/grid/geometry_stencils.py
```diff
@@ -93,6 +138,30 @@ def cartesian_coordinates_of_edge_normal(
...
+
+    Args:
+        edge_tangent_x: x coordinate of the tangent
+        edge_tangent_y: y coordinate of the tangent
+    Returns:
+        edge_normal_x: x coordinate of the normal
+        edge_normal_y: y coordinate of the normal
+        edge_normal_z: y coordinate of the normal
+    """
+    z = 0.0 * edge_tangent_x  # TODO(msimberg): zeros
```
msimberg: To do: same as in https://github.com/C2SM/icon4py/pull/953/files#r2542084148.

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -63,7 +63,10 @@ def __call__(self, array: data_alloc.NDArray):
         )
 
 
-CoordinateDict: TypeAlias = dict[gtx.Dimension, dict[Literal["lat", "lon"], gtx.Field]]
+# TODO(msimberg): x, y, z added temporarily for torus grids, but should they be in a separate dict?
+CoordinateDict: TypeAlias = dict[
+    gtx.Dimension, dict[Literal["lat", "lon", "x", "y", "z"], gtx.Field]
+]
```
msimberg: I haven't tried it yet, but I think the places where `CoordinateDict` is used could be changed to a `GeometryDict`. We possibly lose a bit of type safety (any `GeometryName` could be used instead of the closed set of naems currently in `CoordinateDict`).

@halungge do you have thoughts on what you'd prefer to see here (I see the TODO below, so you may already have thought about this)?

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -501,50 +578,253 @@ def _register_computed_fields(self) -> None:
...
+        log.info(
+            f"initialized geometry for backend = '{self._backend_name()}' and grid = '{self._grid}'"
+        )
+
+        coordinates_ = {
+            attrs.CELL_LAT: coordinates[dims.CellDim]["lat"],
+            attrs.CELL_LON: coordinates[dims.CellDim]["lon"],
+            attrs.CELL_CENTER_X: coordinates[dims.CellDim]["x"],
+            attrs.CELL_CENTER_Y: coordinates[dims.CellDim]["y"],
+            attrs.CELL_CENTER_Z: coordinates[dims.CellDim]["z"],
```
jcanton: We can remove all `z` fields, they are present in the grid file but ICON never uses them (just to double check, right @OngChia ?)

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -174,9 +249,11 @@ def __init__(
             }
         )
         self.register_provider(input_fields_provider)
-        self._register_computed_fields()
+        self._register_computed_fields(input_fields_provider)
 
-    def _register_computed_fields(self) -> None:
+    def _register_computed_fields(
+        self, input_fields_provider: factory.PrecomputedFieldProvider
```
jcanton:
```suggestion
        self
```
this input is unused?

model/common/src/icon4py/model/common/grid/geometry.py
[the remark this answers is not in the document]
msimberg: Diff is unclear, but are you referring to `latitude_of_edge_cell_neighbor_0` and friends? To me they also look unused, but @halungge can hopefully confirm if there was another planned use for them.

PR #954, opened 2025-11-20
Add torus support for interpolation fields

msimberg:
Best viewed as diff to #953, on which this is based: https://github.com/msimberg/icon4py/compare/better-torus-support-geometry...msimberg:icon4py:better-torus-support-interpolation.

### DIFF
--- a/model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
+++ b/model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
@@ -15,6 +15,7 @@ import icon4py.model.common.interpolation.stencils.compute_nudgecoeffs as nudgec
 from icon4py.model.common import constants, dimension as dims
 from icon4py.model.common.decomposition import definitions
 from icon4py.model.common.grid import (
+    base,
     geometry,
     geometry_attributes as geometry_attrs,
     grid_refinement as refinement,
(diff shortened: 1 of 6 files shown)


model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -182,31 +185,59 @@ def _register_computed_fields(self) -> None:
...
+                        ),
+                        "horizontal_start_level_3": self.grid.start_index(
+                            cell_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_3)
+                        ),
+                        "divavg_cntrwgt": self._config["divavg_cntrwgt"],
+                    },
+                )
+                self.register_provider(cell_average_weight)
+            case _:
+                raise ValueError("TODO")
```
msimberg: Probably `assert_never` is better. Here and elsewhere.

model/common/src/icon4py/model/common/interpolation/interpolation_factory.py
```diff
@@ -359,10 +426,17 @@ def _register_computed_fields(self) -> None:
...
             params={
                 "rbf_kernel": self._config["rbf_kernel_cell"].value,
+                "geometry_type": self._grid.global_properties.geometry_type.value,
                 "scale_factor": self._config["rbf_scale_cell"],
                 "horizontal_start": self.grid.start_index(
                     cell_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_2)
                 ),
+                "domain_length": self._grid.global_properties.domain_length
+                if self._grid.global_properties.domain_length
+                else -1.0,
```
msimberg: Ugly, but have to pass numeric values to factory.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -647,6 +674,31 @@ def compute_mass_conserving_bilinear_cell_average_weight(
...
+    # with the the following. Is it needed?
+    return _force_mass_conservation_to_c_bln_avg(
+        c2e2c0,
+        c_bln_avg,
+        cell_areas,
+        cell_owner_mask,
+        divavg_cntrwgt,
+        horizontal_start_level_3,
+        array_ns,
+    )
```
msimberg: I can't really judge if there are cases for torus where `force_mass_conservation` is actually required for some grids.

If not, we could leave it out, but the inaccuracy that it introduces is also minimal, so may be safer to leave it in.

PR #956, opened 2025-11-21
blueline: move vertical grid to grid_wrapper, add flat_height

havogt:
- Move the VerticalGrid initialization to grid_wrapper (instead of repeating in each granule).
- Adds `flat_height` to the interface. Fixes the problem with the exclaim_uncoupled experiments added in https://github.com/C2SM/icon-exclaim/pull/383, as they are using non-default flat_height.

In a follow-up PR we should require to pass all options from Fortran.


muellch (approved):
Looks good!

PR #962, opened 2025-11-25
Update DaCe version

edopao:
Update the dace version to ensure that the correct version of `networkx` is installed: DaCe is currently not compatible with `networkx == 3.6`.


edopao:
cscs-ci run default

PR #966, opened 2025-12-02
Global min reduction

nfarabullini:
Addition of `global_min` reduction for `nflat_gradp`:

```python
    def global_min(self, buffer: data_alloc.NDArray) -> state_utils.ScalarType:
        min_buffer_arr = self._xp.array([self._xp.min(buffer)])
        recv_buffer = self._xp.empty(1, dtype=buffer.dtype)
        self.props.comm.Allreduce(min_buffer_arr, recv_buffer, mpi4py.MPI.MIN)
        return recv_buffer.item()
```

### DIFF
--- a/model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
+++ b/model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
@@ -194,6 +194,9 @@ def test_domain_descriptor_id_are_globally_unique(
 @pytest.mark.mpi
 @pytest.mark.datatest
 @pytest.mark.parametrize("processor_props", [True], indirect=True)
+@pytest.mark.parametrize(
+    "experiment", [test_defs.Experiments.MCH_CH_R04B09, test_defs.Experiments.EXCLAIM_APE]
+)
 def test_decomposition_info_matches_gridsize(
     caplog: Any,
     decomposition_info: definitions.DecompositionInfo,
@@ -223,6 +226,9 @@ def test_decomposition_info_matches_gridsize(
 
 @pytest.mark.mpi
 @pytest.mark.parametrize("processor_props", [True], indirect=True)
+@pytest.mark.parametrize(
+    "experiment", [test_defs.Experiments.MCH_CH_R04B09, test_defs.Experiments.EXCLAIM_APE]
+)
 def test_create_multi_node_runtime_with_mpi(
     decomposition_info: definitions.DecompositionInfo,
     processor_props: definitions.ProcessProperties,
@@ -248,6 +254,9 @@ def test_create_single_node_runtime_without_mpi(
 @pytest.mark.mpi
 @pytest.mark.parametrize("processor_props", [True], indirect=True)
 @pytest.mark.parametrize("dimension", (dims.CellDim, dims.VertexDim, dims.EdgeDim))
+@pytest.mark.parametrize(
+    "experiment", [test_defs.Experiments.MCH_CH_R04B09, test_defs.Experiments.EXCLAIM_APE]
+)
 def test_exchange_on_dummy_data(
     processor_props: definitions.ProcessProperties,
     decomposition_info: definitions.DecompositionInfo,
@@ -290,6 +299,9 @@ def test_exchange_on_dummy_data(
 @pytest.mark.mpi
 @pytest.mark.datatest
 @pytest.mark.parametrize("processor_props", [False], indirect=True)
+@pytest.mark.parametrize(
+    "experiment", [test_defs.Experiments.MCH_CH_R04B09, test_defs.Experiments.EXCLAIM_APE]
+)
 def test_halo_exchange_for_sparse_field(
     interpolation_savepoint: serialbox.InterpolationSavepoint,
     experiment: test_defs.Experiment,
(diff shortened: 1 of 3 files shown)


model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -437,3 +439,18 @@ def create_multinode_node_exchange(
...
     else:
         return SingleNodeExchange()
+
+
+class GlobalReductions(Reductions):
+    def __init__(self, array_ns: ModuleType = np):
+        self._xp = array_ns
+
+    def global_min(self, buffer: data_alloc.NDArray) -> state_utils.ScalarType:
+        if mpi4py is None:
```
halungge: remove this if and add an implementation of the protocol without the `mpi4py` in `definitions.py`. 
Add a `singledispatch` function to the the return the version base on the `ProcessProperties` implementation. Pretty much the same pattern as for the `definitions.create_exchange` function.

If you run single node (and have no mpi4py installed at all) this file here should not get imported anywhere.

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -365,4 +379,20 @@ def create_single_node_exchange(
...
+    Create an Exchange depending on the runtime size.
+
+    Depending on the number of processor a SingleNode version is returned or a GHEX context created and a Multinode returned.
+    """
+    raise NotImplementedError(f"Unknown ProcessorProperties type ({type(props)})")
+
+
+@create_global_reduction.register(SingleNodeProcessProperties)
+def create_global_reduction_exchange(props: SingleNodeProcessProperties) -> Reductions:
+    return SingleNodeReductions()
```
halungge: you need to handle the `array_ns` here somehow.... or you remove it from the constructor and pass it to the `min` function to post pone the decision here, where we don not really know. Wherever we call the `min` there probably is a `backend` around to determine the device.

model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
```diff
@@ -329,3 +333,20 @@ def test_halo_exchange_for_sparse_field(
...
+    processor_props: definitions.ProcessProperties,
+) -> None:
+    my_rank = processor_props.rank
+    print(f"rank={my_rank}")
+
+    global_reduc = definitions.create_global_reduction(processor_props)
+    arr = np.arange(3) - my_rank
+
+    min_val = global_reduc.min(arr)
+    expected_val = -my_rank
```
halungge: That will fail: The expected value is a constant, the same on all nodes because you do a reduction over all ranks.  It is  `- (processor_props.comm_size-1)` which is `np.min(np.arange(3) - "maximal rank number") `

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -423,3 +425,20 @@ def create_multinode_node_exchange(
...
+
+class GlobalReductions(Reductions):
+    def __init__(self, array_ns: ModuleType = np):
+        self._xp = array_ns
+
+    def min(self, buffer: data_alloc.NDArray) -> state_utils.ScalarType:
+        props = get_multinode_properties(definitions.get_runtype(with_mpi=True))
+        min_buffer_arr = self._xp.array([self._xp.min(buffer)])
+        recv_buffer = self._xp.empty(1, dtype=buffer.dtype)
+        props.comm.Allreduce(min_buffer_arr, recv_buffer, mpi4py.MPI.MIN)
```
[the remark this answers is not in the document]
nfarabullini: I can't test that at the moment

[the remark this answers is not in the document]
nfarabullini: However, I can accept your edit and if you can test on your local machine, that's fine by me

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run default

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -423,3 +425,19 @@ def create_multinode_node_exchange(
...
+
+
+class GlobalReductions(Reductions):
+    def __init__(self, props: definitions.ProcessProperties):
+        self._props = props
+
+    def min(self, buffer: data_alloc.NDArray, array_ns: ModuleType = np) -> state_utils.ScalarType:
+        local_min = array_ns.min(buffer)
+        recv_buffer = array_ns.empty(1, dtype=buffer.dtype)
+        self._props.comm.Allreduce(local_min, recv_buffer, mpi4py.MPI.MIN)
```
halungge: We should probably consider the stream synchronisation issue here for cupy arrays. See the example in the tutorial: https://mpi4py.readthedocs.io/en/stable/tutorial.html#gpu-aware-mpi-python-gpu-arrays 
I did not test it on GPU, did you? I would at least leave an TODO here that this might become a problem. Also ask the HPC - MPI pros like @msimberg ...

nfarabullini:
cscs-ci run default

PR #978, opened 2025-12-18
Update dace version to tag 2025_12_18

edopao:
Update dace version to tag `__gt4py-next-integration_2025_12_18` (includes fix for write-write hazard in state fusion, see https://github.com/spcl/dace/pull/2256).


edopao:
cscs-ci run default

PR #979, opened 2025-12-18
Cleanup surface computations for cell diagnostics

havogt:
Might fix performance regression after DaCe fix handling multiple writes to same output buffer from a program.

Might be a temporary solution before #896 is ready.


havogt:
cscs-ci run default

PR #987, opened 2026-01-08
Combined stencils with better name

OngChia:
It was not always a good idea to explicitly name those combined stencils according to their functionality because they were combined more or less for performance reasons. Output of variables on multiple domains was also not allowed when we combined them. Therefore, some stencils became excessively long. For example, `compute_contravariant_correction_and_advection_in_vertical_momentum_equation` is not a good name, These two operations are grouped together just because of performance issue. `compute_derived_horizontal_winds_and_ke_and_contravariant_correction` is also not nice. They are inside one stencil simply because they are all edge k fields.
As a work around before we can freely combine stencils with more scientific meaning, I rename those stencils with excessively long names using more concise, case-by-case naming, while retaining the sufficient information of what these stencils compute so that domain scientists or users still can easily understand the computational flow of dycore. Here is a list of the old and new name of the combined stencils:
```
1. compute_advection_in_horizontal_momentum_equation -> compute_advection_in_horizontal_momentum
2. compute_advection_in_vertical_momentum_equation -> compute_advection_in_corrector_vertical_momentum
3. compute_contravariant_correction_and_advection_in_vertical_momentum_equation -> compute_advection_in_predictor_vertical_momentum
4. compute_derived_horizontal_winds_and_ke_and_contravariant_correction -> compute_diagnostics_from_normal_wind
5. compute_averaged_vn_and_fluxes_and_prepare_tracer_advection -> compute_averaged_vn_and_fluxes
6. compute_theta_rho_face_values_and_pressure_gradient_and_update_vn -> compute_rho_theta_pgrad_and_update_vn
7. interpolate_rho_theta_v_to_half_levels_and_compute_pressure_buoyancy_acceleration -> compute_interpolation_and_nonhydro_buoy
```

### DIFF
--- a/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity_advection.py
+++ b/model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity_advection.py
@@ -374,7 +374,7 @@ class VelocityAdvection:
         # Note, if we compute `apply_extra_diffusion_on_vn = max_vertical_cfl > cfl_w_limit * dtime` here,
         # we would have to synchronize with the device already here to get the value of `max_vertical_cfl`.
         apply_extra_diffusion_on_vn = True
-        self._compute_advection_in_horizontal_momentum_equation(
+        self._compute_advection_in_horizontal_momentum(
             normal_wind_advective_tendency=diagnostic_state.normal_wind_advective_tendency.corrector,
             vn=prognostic_state.vn,
             horizontal_kinetic_energy_at_edges_on_model_levels=horizontal_kinetic_energy_at_edges_on_model_levels,


OngChia:
cscs-ci run default

havogt (commented):
I like the names getting shorter.

One note on `at_xxx_on_yyy_levels`. Thanks to the requirement of having type annotations which already encode this info (at least after having support for staggering in gt4py, which might be soon -> "KHalfDim"), I would consider shortening this. E.g. in places where there is the field only on one dimension type, I would consider entirely removing the suffix. Let me know if you want to discuss further what I mean here.

Other thinks to check:
- Just make sure that the changes are synced with icon-exclaim names of the timers (if that applies to these changes).
- In case timer names need to change, also maybe double-check with Will if these changes are already enough for the symbol length problem that he reported for the long timer timers.

OngChia:
> I like the names getting shorter.
> 
> One note on `at_xxx_on_yyy_levels`. Thanks to the requirement of having type annotations which already encode this info (at least after having support for staggering in gt4py, which might be soon -> "KHalfDim"), I would consider shortening this. E.g. in places where there is the field only on one dimension type, I would consider entirely removing the suffix. Let me know if you want to discuss further what I mean here.
> 
> Other thinks to check:
> 
> * Just make sure that the changes are synced with icon-exclaim names of the timers (if that applies to these changes).
> * In case timer names need to change, also maybe double-check with Will if these changes are already enough for the symbol length problem that he reported for the long timer timers.

I definitely it's worth discussing the naming of variables again while we refactor the dycore stencils.

vectorflux:
I have a branch `icon-exclaim:shorter_kernel_names` which mirrors the new kernel names in the timers. I think it is time to merge your PR and follow up with a PR in icon-exclaim.

vectorflux:
Thanks, Chia Rui!

PR #988, opened 2026-01-09
Global ops reduction

nfarabullini:
Implementation of global `max` and `sum` for MPI operations in the same fashion as `min`:
Addition of `global_min` reduction for `nflat_gradp`:

```python
    def global_min(self, buffer: data_alloc.NDArray) -> state_utils.ScalarType:
        min_buffer_arr = self._xp.array([self._xp.min(buffer)])
        recv_buffer = self._xp.empty(1, dtype=buffer.dtype)
        self.props.comm.Allreduce(min_buffer_arr, recv_buffer, mpi4py.MPI.MIN)
        return recv_buffer.item()
```
Note: mpi4py.MPI does not have a `mean` operator

### DIFF
--- a/model/common/src/icon4py/model/common/decomposition/definitions.py
+++ b/model/common/src/icon4py/model/common/decomposition/definitions.py
@@ -400,7 +400,7 @@ def create_single_node_exchange(
 
 
 @functools.singledispatch
-def create_global_reduction(props: ProcessProperties) -> Reductions:
+def create_reduction(props: ProcessProperties) -> Reductions:
     """
     Create a Global Reduction depending on the runtime size.
 
@@ -409,8 +409,8 @@ def create_global_reduction(props: ProcessProperties) -> Reductions:
     raise NotImplementedError(f"Unknown ProcessorProperties type ({type(props)})")
 
 
-@create_global_reduction.register(SingleNodeProcessProperties)
-def create_global_reduction_exchange(props: SingleNodeProcessProperties) -> Reductions:
+@create_reduction.register(SingleNodeProcessProperties)
+def create_single_reduction_exchange(props: SingleNodeProcessProperties) -> Reductions:
     return SingleNodeReductions()
 
 
--- a/model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
+++ b/model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
@@ -501,6 +501,6 @@ class GlobalReductions(Reductions):
         )
 
 
-@definitions.create_global_reduction.register(MPICommProcessProperties)
-def create_global_reduction_exchange(props: MPICommProcessProperties) -> Reductions:
+@definitions.create_reduction.register(MPICommProcessProperties)
+def create_global_reduction(props: MPICommProcessProperties) -> Reductions:
     return GlobalReductions(props)
--- a/model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
+++ b/model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
@@ -343,7 +343,7 @@ def test_global_reductions_min(
     chunks = np.array_split(global_list, comm_size)
     local_data = xp.array(chunks[my_rank])
 
-    global_reduc = definitions.create_global_reduction(processor_props)
+    global_reduc = definitions.create_reduction(processor_props)
 
     if len(global_list) > 0:
         min_val = global_reduc.min(local_data, array_ns=xp)
@@ -368,7 +368,7 @@ def test_global_reductions_max(
     chunks = np.array_split(global_list, comm_size)
     local_data = xp.array(chunks[my_rank])
 
-    global_reduc = definitions.create_global_reduction(processor_props)
+    global_reduc = definitions.create_reduction(processor_props)
 
     if len(global_list) > 0:
         max_val = global_reduc.max(local_data, array_ns=xp)
@@ -393,7 +393,7 @@ def test_global_reductions_sum(
     chunks = np.array_split(global_list, comm_size)
     local_data = xp.array(chunks[my_rank])
 
-    global_reduc = definitions.create_global_reduction(processor_props)
+    global_reduc = definitions.create_reduction(processor_props)
 
     if len(global_list) > 0:
         sum_val = global_reduc.sum(local_data, array_ns=xp)
@@ -417,7 +417,7 @@ def test_global_reductions_mean(
     comm_size = processor_props.comm_size
     chunks = np.array_split(global_list, comm_size)
     local_data = xp.array(chunks[my_rank])
-    global_reduc = definitions.create_global_reduction(processor_props)
+    global_reduc = definitions.create_reduction(processor_props)
 
     if len(global_list) > 0:
         mean_val = global_reduc.mean(local_data, array_ns=xp)
--- a/model/common/tests/common/fixtures.py
+++ b/model/common/tests/common/fixtures.py
@@ -135,7 +135,7 @@ def metrics_factory_from_savepoint(
     interpolation_factory_from_savepoint: interpolation_factory.InterpolationFieldsFactory,
 ) -> Generator[metrics_factory.MetricsFieldsFactory]:
     exchange = decomposition.create_exchange(processor_props, decomposition_info)
-    global_reductions = decomposition.create_global_reduction(processor_props)
+    global_reductions = decomposition.create_reduction(processor_props)
     geometry_source = geometry_from_savepoint
     interpolation_field_source = interpolation_factory_from_savepoint
     topography = topography_savepoint.topo_c()


model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -423,3 +425,46 @@ def create_multinode_node_exchange(
...
+
+    def min(self, buffer: data_alloc.NDArray, array_ns: ModuleType = np) -> state_utils.ScalarType:
+        local_min = array_ns.min(buffer)
+        recv_buffer = array_ns.empty(1, dtype=buffer.dtype)
+        if hasattr(
+            array_ns, "cuda"
+        ):  #  https://mpi4py.readthedocs.io/en/stable/tutorial.html#gpu-aware-mpi-python-gpu-arrays
+            array_ns.cuda.runtime.deviceSynchronize()
+        self._props.comm.Allreduce(local_min, recv_buffer, mpi4py.MPI.MIN)
+        return recv_buffer.item()
```
msimberg: I would refactor this to use a helper function for the common functionality. Something like:

```suggestion

    def _reduce(self, buffer: data_alloc.NDArray, local_reduction: ???, distributed_reduction: ???, array_ns: ModuleType = np) -> state_utils.ScalarType:
        local_min = local_reduction(buffer)
        recv_buffer = array_ns.empty(1, dtype=buffer.dtype)
        if hasattr(
            array_ns, "cuda"
        ):  #  https://mpi4py.readthedocs.io/en/stable/tutorial.html#gpu-aware-mpi-python-gpu-arrays
            array_ns.cuda.runtime.deviceSynchronize()
        self._props.comm.Allreduce(local_min, recv_buffer, global_reduction)
        return recv_buffer.item()        
    
    def min(self, buffer: data_alloc.NDArray, array_ns: ModuleType = np) -> state_utils.ScalarType:
        return _reduce(buffer, array_ns.min, mpi4py.MPI.MIN, array_ns)
```
etc. for the other operations.

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -607,8 +611,8 @@ def compute_nflat_gradp(
         flat_idx_max,
         nlev,
     )
-    nflat_gradp = array_ns.min(mask_array)
-    return nflat_gradp.item()
+    nflat_gradp = min_reduction(mask_array, array_ns=array_ns)
+    return nflat_gradp
```
msimberg: Just checking, does this change belong in this PR? If no, would you mind separating it out into a different PR?

nfarabullini:
cscs-ci run default

model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
```diff
@@ -320,7 +333,87 @@ def test_halo_exchange_for_sparse_field(
...
+    arr = np.arange(3) - my_rank
+
+    global_reduc = definitions.create_global_reduction(processor_props)
+
+    mean_val = global_reduc.mean(arr, xp)
+    expected_val = (processor_props.comm_size + 1) / arr.size
+    print(
+        f"rank={my_rank}/{processor_props.comm_size}: input data [{arr}] - reduction result [{mean_val}] "
+    )
+    assert expected_val == mean_val
```
msimberg: I think these could do with some more tests. I would consider parameterizing the tests over a list of global lists of scalars. You could then do a "decomposition" to get the local data using something like `array_split` (https://numpy.org/doc/stable/reference/generated/numpy.array_split.html). The global MPI reduction can then easily be compared to a local reduction over the full list. If the input values are relatively small lists this should still be quick to test.

Additionally, I would add tests with empty local buffers. You should get that for free if you add some input lists with fewer elements than ranks in the communicator. Typically this should of course not happen, but it doesn't hurt to guard against it.

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -365,4 +399,20 @@ def create_single_node_exchange(
...
 
 
+@functools.singledispatch
+def create_global_reduction(props: ProcessProperties) -> Reductions:
+    """
+    Create an Exchange depending on the runtime size.
+
+    Depending on the number of processor a SingleNode version is returned or a GHEX context created and a Multinode returned.
+    """
+    raise NotImplementedError(f"Unknown ProcessorProperties type ({type(props)})")
```
msimberg: This is still to do, right?

model/common/src/icon4py/model/common/metrics/metrics_factory.py
```diff
@@ -69,6 +69,7 @@ def __init__(
         exner_expol: float,
         vwind_offctr: float,
         exchange: decomposition.ExchangeRuntime = decomposition.single_node_default,
+        global_reductions: decomposition.Reductions = decomposition.single_node_reductions,
```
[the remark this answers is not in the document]
nfarabullini: global_reductions are defined like this: 
```
global_reductions = decomposition_defs.create_global_reduction(parallel_props)
```
whereas, exchange: 
```
exchange = decomposition_defs.create_exchange(parallel_props, decomposition_info)
```
They come from 2 different functions

model/common/src/icon4py/model/common/decomposition/definitions.py
msimberg: I see now where the other implementation is. Sorry for the noise.

model/common/src/icon4py/model/common/grid/icon.py
```diff
@@ -97,19 +98,26 @@ def from_fields(
...
-        mean_edge_length = init_mean(mean_edge_length, edge_lengths)
-        mean_dual_edge_length = init_mean(mean_dual_edge_length, dual_edge_lengths)
-        mean_cell_area = init_mean(mean_cell_area, cell_areas)
-        mean_dual_cell_area = init_mean(mean_dual_cell_area, dual_cell_areas)
+        mean_edge_length = init_mean(mean_edge_length, edge_lengths, array_ns=array_ns)
+        mean_dual_edge_length = init_mean(
+            mean_dual_edge_length, dual_edge_lengths, array_ns=array_ns
+        )
+        mean_cell_area = init_mean(mean_cell_area, cell_areas, array_ns=array_ns)
+        mean_dual_cell_area = init_mean(mean_dual_cell_area, dual_cell_areas, array_ns=array_ns)
```
msimberg: I don't think you need to pass `array_ns` to each invocation, you can take it from the outer scope:
```suggestion
        def init_mean(
            value: float | None, data: data_alloc.NDArray | None) -> float | None:
if value is not None:
return value
if data is not None:
                return mean_reduction(data, array_ns=array_ns)
return None

        mean_edge_length = init_mean(mean_edge_length, edge_lengths)
        mean_dual_edge_length = init_mean(mean_dual_edge_length, dual_edge_lengths)
        mean_cell_area = init_mean(mean_cell_area, cell_areas)
        mean_dual_cell_area = init_mean(mean_dual_cell_area, dual_cell_areas)
```

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -423,3 +426,70 @@ def create_multinode_node_exchange(
...
+
+    def _calc_buffer_size(
+        self,
+        buffer: data_alloc.NDArray,
+        array_ns: ModuleType = np,
+    ) -> state_utils.ScalarType:
+        return self._reduce(array_ns.asarray(buffer.size), array_ns.sum, mpi4py.MPI.SUM, array_ns)
+
+    def min(self, buffer: data_alloc.NDArray, array_ns: ModuleType = np) -> state_utils.ScalarType:
```
msimberg: If we handle the empty list case with returning `None`, then the return type should be updated to reflect that. It seems like `numpy.mean` returns nan. `numpy.min/max` raise a `ValueError`. I think doing the latter for all cases is fair, but don't have a strong opinion.

msimberg (changes requested):
Two minor requests, sorry for not spotting them earlier. Let's get this in after that.

model/common/tests/common/decomposition/mpi_tests/test_mpi_decomposition.py
```diff
@@ -320,7 +321,112 @@ def test_halo_exchange_for_sparse_field(
...
+    global_reduc = definitions.create_global_reduction(processor_props)
+
+    if len(global_list) > 0:
+        min_val = global_reduc.min(local_data, array_ns=xp)
+        expected_val = np.min(global_list)
+        assert expected_val == min_val
+    else:
+        with pytest.raises(ValueError, match="global_min requires a non-empty buffer") as excinfo:
+            global_reduc.min(local_data, array_ns=xp)
+        assert excinfo.type is ValueError
```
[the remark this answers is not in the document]
nfarabullini: good point

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -423,3 +426,81 @@ def create_multinode_node_exchange(
...
+            array_ns.min,
+            mpi4py.MPI.MIN,
+            array_ns,
+        )
+
+    def max(self, buffer: data_alloc.NDArray, array_ns: ModuleType = np) -> state_utils.ScalarType:
+        if self._calc_buffer_size(buffer, array_ns) == 0:
+            raise ValueError("global_max requires a non-empty buffer")
+        return self._reduce(
+            buffer if buffer.size != 0 else array_ns.asarray([-np.inf]),
```
jcanton:
```suggestion
            buffer if buffer.size != 0 else array_ns.asarray([-array_ns.inf]),
```

jcanton (approved):
LGTM

PR #995, opened 2026-01-19
muphys: Only write the surface level of the fluxes

havogt:


edopao:
cscs-ci run default

edopao:
cscs-ci run dace

PR #999, opened 2026-01-21
Muphys: allow re-use of fields in graupeloutput

havogt:


model/atmosphere/subgrid_scale_physics/muphys/tests/muphys/integration_tests/test_graupel_only.py
```diff
@@ -66,6 +66,14 @@ def test_graupel_only(
     out = common.GraupelOutput.allocate(
```
havogt:
```suggestion
    # We are passing the same buffers for `Q` as input and output. This is not best GT4Py practice,
    # but save in this case as we are not reading the input with an offset.
    out = common.GraupelOutput.allocate(
```

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/driver/run_graupel_only.py
```diff
@@ -69,8 +69,19 @@ def main():
     allocator = model_backends.get_allocator(backend)
 
     inp = common.GraupelInput.load(filename=pathlib.Path(args.input_file), allocator=allocator)
+    # We are passing the same buffers for `Q` as input and output. This is not best GT4Py practice,
```
vectorflux: Thanks.

PR #1000, opened 2026-01-21
muphys: add masking switch

havogt:
- [ ] Needs to be applied to full_muphys


model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/driver/run_graupel_only.py
vectorflux: Nothing mysterious here.

PR #1012, opened 2026-01-28
Add gtfn_gpu to distributed CI pipeline

msimberg:
Adds gtfn_gpu backend to the distributed CI pipeline. dace_gpu is still left out because compilation takes too long.

The base image is upgraded because it's possible, but not strictly necessary. The CPU-only version of the pipeline needed 25.04 (24.04 and 25.10 did not work for various reasons). However, since OpenMPI and libfabric are now built manually in the container the base image version is less of a constraint. 24.04 doesn't have matching GCC/CUDA versions and 26.04 doesn't exist yet, but the pipeline should eventually use 26.04.

OpenMPI and libfabric are built manually for slingshot support because getting the ubuntu repository packages to work with GPU support did not seem possible/easy. The installation is based on https://github.com/eth-cscs/cray-network-stack.

GHEX needs an upgrade, because there's a bug in how strides are calculated for GPU buffers. @philip-paul-mueller has already fixed this in https://github.com/ghex-org/GHEX/pull/190 but we should wait for that to be merged (and probably test in icon-exclaim first).

This also fixes a few cupy/numpy incompatibilities. `revert_repeated_index_to_invalid` was updated to only deal with numpy for now as the connectivities are always numpy arrays. `test_halo_exchange_for_sparse_field` is marked `embedded_only`. The non-MPI test was already marked embedded-only.

This does not try to unify the default and distributed CI pipeline definitions. That should, however, be done done sooner or later as well.


msimberg:
cscs-ci run distributed

pyproject.toml
```diff
@@ -361,7 +361,7 @@ url = 'https://gridtools.github.io/pypi/'
 
 [tool.uv.sources]
 dace = {index = "gridtools"}
-ghex = {git = "https://github.com/msimberg/GHEX.git", branch = "async-mpi"}
+ghex = {git = "https://github.com/philip-paul-mueller/GHEX.git", branch = "phimuell__async-mpi-2"}
```
msimberg: This is updated because https://github.com/ghex-org/GHEX/pull/190 contains a bugfix to how strides are computed for GPU buffers. Tests fail with `master` and `async-mpi`. We should get https://github.com/ghex-org/GHEX/pull/190 merged ASAP to be able to use GHEX master here.

model/common/src/icon4py/model/common/grid/utils.py
```diff
@@ -5,21 +5,20 @@
...
 # SPDX-License-Identifier: BSD-3-Clause
-from types import ModuleType
 
 import numpy as np
 
 from icon4py.model.common.grid import gridfile
 
 
-def revert_repeated_index_to_invalid(offset: np.ndarray, array_ns: ModuleType):
+def revert_repeated_index_to_invalid(offset: np.ndarray):
```
nfarabullini:
```suggestion
from icon4py.model.common.utils import data_allocation as data_alloc
def revert_repeated_index_to_invalid(offset: data_alloc.NDArray):
```

jcanton:
sorry ;-)

jcanton:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

nfarabullini:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

PR #1020, opened 2026-01-29
Fix data download paths

msimberg:
Attempting to fix data downloads after #1002.

With my current understanding of how the downloads work, the download and extraction should happen in the "root" directory (`_ranked_data_path`), but the tarball actually contains subdirectories which match the experiment-specific subdirectory (which is what we check for to see if the data already exists).

I've tested re-downloading one grid and that worked. Please test it on your local setups to see I haven't messed something up.

### DIFF
--- a/model/testing/src/icon4py/model/testing/locking.py
+++ b/model/testing/src/icon4py/model/testing/locking.py
@@ -25,5 +25,5 @@ def lock(
     if not directory.is_dir():
         raise ValueError(f"Expected a directory, got: {directory}")
 
-    path = directory / f"{lockfile}"
+    path = directory / lockfile
     return filelock.FileLock(str(path))


havogt (commented):
Thanks for the fix! With the suggested change, my use-case would be covered as well. I tested a few cases and seems good.

model/testing/src/icon4py/model/testing/data_handling.py
```diff
@@ -36,13 +36,14 @@ def download_and_extract(uri: str, dst: pathlib.Path, data_file: str = "download
...
 
-def download_test_data(dst: pathlib.Path, uri: str) -> None:
+def download_test_data(dst_root: pathlib.Path, dst_subdir: pathlib.Path, uri: str) -> None:
+    dst = dst_root.joinpath(dst_subdir)
     if config.ENABLE_TESTDATA_DOWNLOAD:
         # We create and lock the *parent* directory as we later check for existence of `dst`.
-        dst.parent.mkdir(parents=True, exist_ok=True)
-        with locking.lock(dst.parent):
+        dst_root.mkdir(parents=True, exist_ok=True)
+        with locking.lock(dst_root):
```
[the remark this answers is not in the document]
jcanton: I don't think so

model/testing/src/icon4py/model/testing/data_handling.py
```diff
@@ -36,13 +36,14 @@ def download_and_extract(uri: str, dst: pathlib.Path, data_file: str = "download
...
             if not dst.exists():
```
havogt: check if dst is empty

model/testing/src/icon4py/model/testing/data_handling.py
```diff
@@ -36,13 +36,14 @@ def download_and_extract(uri: str, dst: pathlib.Path, data_file: str = "download
     pathlib.Path(data_file).unlink(missing_ok=True)
 
 
-def download_test_data(dst: pathlib.Path, uri: str) -> None:
+def download_test_data(dst_root: pathlib.Path, dst_subdir: pathlib.Path, uri: str) -> None:
```
havogt:
```suggestion
def download_test_data(dst_root: pathlib.Path, dst_subdir: pathlib.Path, uri: str) -> None:
    # TODO(havogt): remove `dst_subdir` after icon serialization data doesn't contain subdir with the `dst_subdir` name.
```

model/testing/src/icon4py/model/testing/data_handling.py
msimberg: I now explicitly set the lockfile name, so that even if we change the internals the lockfile will not be a hidden file.

In theory the current change means that the lockfile could be created outside the destination directory and this should still work. Slightly overkill, but maybe still ok.

model/testing/src/icon4py/model/testing/locking.py
```diff
@@ -17,11 +17,13 @@
...
+def lock(
+    directory: pathlib.Path | str, lockfile: str = "filelock.lock"
+) -> contextlib.AbstractContextManager:
     """Create a lock for the given path."""
     directory = pathlib.Path(directory)
     if not directory.is_dir():
         raise ValueError(f"Expected a directory, got: {directory}")
 
-    path = directory / f"filelock{suffix}"
+    path = directory / f"{lockfile}"
```
havogt:
```suggestion
    path = directory / lockfile
```

PR #1022, opened 2026-01-30
Muphys: edit config for dace CI

edopao:


havogt:
I would do it globally for all tests as this is the setup we care for production.

PR #1025, opened 2026-01-30
Update distributed CI testdata path

msimberg:
Use the cwci02 path instead of the old d126 path. #692 was created before the path was changed, and I didn't think to update it to cwci02 while working on it.


edopao:
cscs-ci run default

edopao:
cscs-ci run distributed

edopao:
I noticed that I had to copy from `d126/icon4py/ci/testdata_004`, and not `testdata_003`, because that was the latest data.

PR #1038, opened 2026-02-04
Run extra pipeline in CSCS-CI

msimberg:
Just testing.

### DIFF
--- a/ci/extra.yml
+++ b/ci/extra.yml
@@ -1,3 +1,4 @@
+# TODO: Remove this comment. This is a dummy change.
 include:
   - local: 'ci/base.yml'
 


msimberg:
cscs-ci run extra

msimberg:
The following tests fail (not on all backends, but most):

- tests/common/metrics/unit_tests/test_compute_diffusion_metrics.py::test_compute_diffusion_intcoef_and_vertoffset[False-gauss3d_torus]: wrong mask
- tests/common/metrics/unit_tests/test_compute_diffusion_metrics.py::test_compute_diffusion_mask_and_coeff[False-gauss3d_torus]: wrong mask
- tests/common/interpolation/unit_tests/test_rbf_interpolation.py::test_rbf_interpolation_coeffs_cell[False-experiment1-0.03]: possibly very slightly different tolerance for different backends or gpu, needs checking

The mask failures are already xfail'ed for factory (integration) tests, but the unit test was not marked with anything.

I'll start looking into these.

PR #1049, opened 2026-02-05
Deduplicate RBF test tolerances, use common experiment fixture

msimberg:
This fixes an RBF test "failure", that was caused by the tolerances being slighly out of sync between the factory tests and the regular unit tests (with the latter not run in the default CI pipeline). I've moved the tolerances to a dictionary that is used by all the RBF tests so that they are in sync. At the same time I've updated the RBF tests to use the common experiment fixture so that if we add new experiments for testing the RBF tests will be run with those experiments without changes. Tolerances still have to be added manually per-experiment.

### DIFF
--- a/model/common/tests/common/interpolation/mpi_tests/test_parallel_interpolation.py
+++ b/model/common/tests/common/interpolation/mpi_tests/test_parallel_interpolation.py
@@ -12,6 +12,7 @@ from typing import TYPE_CHECKING
 
 import pytest
 
+from icon4py.model.common import dimension as dims
 from icon4py.model.common.decomposition import definitions as decomposition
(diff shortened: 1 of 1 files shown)


msimberg:
> Nice cleanup! Should this be applied also to the mpi tests?

Good point, I forgot about those. Let me add that as well. The tolerances also seem to be unnecessarily large there.

msimberg:
@nfarabullini I added the common tolerances to `test_parallel_interpolation.py` as well. The rbf tests themselves are still disabled there though, fixing them is part of https://hackmd.io/O4Fymu1dTxqTZSC8rdiVVw?both.

PR #1056, opened 2026-02-11
Replace `mask_hdiff` by multiplying with 0.0

jcanton:
When naively translating a Fortran list to a full field, we constructed a mask instead of using the fact that we are actually initializing a weight with 0.0 in the masked-out cases.

Additionally: Remove program and a trivial test.

same as https://github.com/C2SM/icon4py/pull/1053 but for `mask_hdiff`

### DIFF
--- a/model/atmosphere/diffusion/tests/diffusion/stencil_tests/test_apply_diffusion_to_theta_and_exner.py
+++ b/model/atmosphere/diffusion/tests/diffusion/stencil_tests/test_apply_diffusion_to_theta_and_exner.py
@@ -87,7 +87,9 @@ class TestApplyDiffusionToThetaAndExner(StencilTest):
         inv_dual_edge_length = data_alloc.random_field(grid, dims.EdgeDim)
         theta_v_in = data_alloc.random_field(grid, dims.CellDim, dims.KDim)
         geofac_div = data_alloc.random_field(grid, dims.CellDim, dims.C2EDim)
-        zd_vertoffset = data_alloc.zero_field(grid, dims.CellDim, dims.C2E2CDim, dims.KDim, dtype=gtx.int32)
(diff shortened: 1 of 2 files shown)


model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/apply_diffusion_to_theta_and_exner.py
havogt:
```suggestion
    if apply_zdiffusion_t:
        # Only where `zd_diffcoef` is not 0 to avoid loading the other fields.
        # Eventually, the toolchain could consider extracting a mask `zd_diffcoef != 0` if we tell gt4py that `zd_diffcoef` is static.
        z_temp = (
            where(zd_diffcoef != 0, _truly_horizontal_diffusion_nabla_of_theta_over_steep_points(
                zd_vertoffset,
                zd_diffcoef,
                geofac_n2s_c,
                geofac_n2s_nbh,
                vcoef,
                theta_v,
                z_temp,
            ), z_temp)
    )
```
maybe this?

havogt: I think the suggestion doesn't work properly because it spans to many lines that where not touched.

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run distributed

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/apply_diffusion_to_theta_and_exner.py
jcanton: ah ok, I had misunderstood :-)

PR #1054, opened 2026-02-11
CI: disable cuda mps

havogt:
After CUDA MPS was properly enabled in the CI infrastructure the CI started to fail, probably due to our CI using too many parallel jobs for the default MPS settings. Switching it off as we probably don't profit from MPS anyway.

PR #1057, opened 2026-02-12
Reenable MPS in CI

msimberg:
From local tests it seems to help quite a bit, when concurrency is limited so tests don't run out of memory.

My unverified hypothesis for the out of memory issues we saw when MPS was enabled is that it actually enabled more real concurrency, thus increasing the peak memory usage. Without MPS different pytest workers would be serialized (possibly interleaved, but not running at the same time) which would lead to lower peak memory usage.

I've also increased the default `NUM_PROCESSES` for CPU backends to 32 (I think it could go even higher, though at some point worker startup time starts to dominate; I very approximately observed about 0.5s per worker for the startup time). I changed `NUM_PROCESSES` to 8 for all GPU backends (previously even the CPU backends for common/integration were using one pytest worker).

I've also enabled the persistent gt4py cache so that pytest workers can share the cache. From local testing at least some dace test runs benefited from this.

### DIFF
--- a/ci/default.yml
+++ b/ci/default.yml
@@ -49,7 +49,7 @@ test_tools_datatests_aarch64:
       variables:
         NUM_PROCESSES: 8
         SLURM_TIMELIMIT: '00:30:00'
-    - if: $COMPONENT == 'driver' || $COMPONENT == 'dycore'
+    - if: $COMPONENT == 'driver' || $COMPONENT == 'dycore' || $COMPONENT == 'muphys'
       variables:
         SLURM_TIMELIMIT: '00:30:00'
     - if: $BACKEND == 'embedded'


msimberg:
cscs-ci run default

msimberg:
@edopao with the new settings I get for example for common/integration:
- 25 minutes (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504671/-/jobs/13088519233) vs 37 minutes (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504671/-/jobs/13081609148 from #1054) with dace_gpu
- 4 minutes (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504671/-/jobs/13088519232) vs 8 minutes (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504671/-/jobs/13081609145 from #1054) with dace_cpu

Some of the other jobs don't see any improvements, or even slight degradation. I guess in the worst case the shared cache might actually be a bottleneck, but I think it's still worth it to keep it enabled for the average case.

ci/dace.yml
```diff
@@ -32,17 +36,12 @@ test_model_stencils_aarch64:
...
         LEVEL: [integration]
   rules:
-    - if: $COMPONENT == 'common'
-      variables:
-        NUM_PROCESSES: 2
-        SLURM_TIMELIMIT: '01:30:00'
     - if: $BACKEND == 'dace_gpu'
       variables:
         NUM_PROCESSES: 8
         SLURM_TIMELIMIT: '01:00:00'
```
edopao: This timeout could probably be lowered to 45 minutes.

ci/default.yml
```diff
@@ -39,11 +43,7 @@ test_tools_datatests_aarch64:
...
       when: never  # run only in daily CI, to save compute resources
-    - if: $COMPONENT == 'common' && $LEVEL == 'integration'
-      variables:
-        NUM_PROCESSES: 1
-        SLURM_TIMELIMIT: '00:45:00'
-    - if: $BACKEND == 'dace_gpu'
+    - if: $BACKEND == 'dace_gpu' || $BACKEND == 'gtfn_gpu'
       variables:
         NUM_PROCESSES: 8
         SLURM_TIMELIMIT: '01:00:00'
```
edopao: Same here.

msimberg:
Waiting for CI to be healthy again so I can run the default pipeline. Updating timelimits after that.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

ci/base.yml
```diff
@@ -89,6 +95,7 @@ build_image_aarch64:
     # Grace-Hopper gpu architecture is not enabled by default in CUDA build
     CUDAARCHS: "90"
     # Limit test parallelism to avoid "OSError: too many open files" in the gt4py build stage.
```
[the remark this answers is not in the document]
msimberg: I'll go ahead and merge with the current change, but feel free to comment afterwards if you prefer a different comment.

PR #1062, opened 2026-02-12
Remove xfails in distributed metrics tests

msimberg:
`COEFF_GRADEKIN` has fixed itself so removing the xfail. I also tried it locally with #1013 and the GAUSS3D test case also passes for with `COEFF_GRADEKIN`.

I've also reduced the embedded backend xfail to the one field that actually fails.


msimberg:
cscs-ci run default

PR #1067, opened 2026-02-18
Remove unused vct_b

havogt:


model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -311,19 +300,17 @@ def _read_vct_a_and_vct_b_from_file(
         file_path: Path to the vertical grid file
         num_levels: number of cell levels
         allocator: GT4Py field allocator
-    Returns:  one dimensional vct_a and vct_b arrays.
+    Returns:  one dimensional vct_a arrays.
```
jcanton: maybe explain here why `vct_b` is ignored (even though we keep the file format including vct_b for interoperability with fortran)

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -334,16 +321,14 @@ def _read_vct_a_and_vct_b_from_file(
...
+def _compute_vct_a(  # noqa: PLR0912 [too-many-branches]
     vertical_config: VerticalGridConfig, allocator: gtx_typing.FieldBufferAllocationUtil
-) -> tuple[fa.KField, fa.KField]:
+) -> fa.KField:
     """
-    Compute vct_a and vct_b.
+    Compute vct_a.
 
     When the thickness of lowest level grid cells is larger than 0.01:
         vct_a[k] = H (2/pi arccos((k/N)^s) )^d      N = num_levels      s = stretch_factor in vertical configuration
```
jcanton: and add here a line, something like `if needed vct_b can be computed as vct_b = np.exp(-vct_a / 5000.0)` (so we don't have to look it up somewhere else)

havogt:
Conclusion was: leave it icon4py, because potentially the vct_b might be custom in the grid file (or an alternative computation might be used), but we remove it from the Fortran bindings as we pass the derived quantities.

PR #1068, opened 2026-02-19
Muphys: configure `t` as inout field

edopao:
It is possible to use the same `t` field for input and output of the graupel program. This allows us to use the same interface as other implementations in the muphys-ppp project.


edopao:
cscs-ci run distributed

PR #1069, opened 2026-02-19
Muphys: Use view node in self-copy removal SDFG transformation

edopao:


edopao:
cscs-ci run dace

edopao:
The speedup is very small, on latest baseline. We decided to exclude this change, for now.

PR #1078, opened 2026-02-24
Add missing asserts to dallclose calls

msimberg:


nfarabullini (approved):
Some of these were actually my fault so thank you for fixing them

PR #1079, opened 2026-02-25
Reject namelist parameters not supported

OngChia:
`rayleigh_type`, `lsmag_3d` (`compute_3d_smag_coeff`), `itype_vn_diffu` (`type_vn_diffu`), `itype_t_diffu` (`type_t_diffu`), and `ldiffu_smag_w` (`apply_smag_diff_to_vertical_wind`), where the variable in the bracket represents its corresponding new name in icon4py, are not correctly checked in the initialization of `solve_nonhydro` and `diffusion` granules.
In this PR, we correctly check these namelist parameters in solve_nonhydro and diffusion granules and reject options we do not support based on this table:
https://hackmd.io/YIWGHQ1ERmWW8r7VmF05Lg


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -312,6 +312,11 @@ def _validate(self):
                 "`DivergenceDampingType.TWO_DIMENSIONAL` (2) is not yet implemented"
             )
 
+        if self.rayleigh_type != constants.RayleighType.KLEMP:
+            raise NotImplementedError(
+                "Onle Klemp type of the Rayleigh damping (nudging vertical wind towards zero) is implemented."
+            )
+
```
OngChia: Do we need to check `lextra_diffu` and `ldeepatmo`?

muellch: Hmm we probably hard coded it for performance reasons, although it could be changed to a namelist switch and then the pre-compilation will make sure the performance is good.

For now we can assume `lextra_diffu==True` and reject `lextra_diffu==False`, since no one uses it.  Could you also leave a TODO where we set `apply_extra_diffusion_on_vn=True` that it either should be a namelist parameter or it should completely disappear and we just always do the diffusion?

OngChia:
cscs-ci run default

OngChia:
cscs-ci run distributed

OngChia:
cscs-ci run distributed

OngChia:
cscs-ci run distributed

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/velocity_advection.py
```diff
@@ -372,6 +373,7 @@ def run_corrector_step(
 
         # Note, if we compute `apply_extra_diffusion_on_vn = max_vertical_cfl > cfl_w_limit * dtime` here,
         # we would have to synchronize with the device already here to get the value of `max_vertical_cfl`.
+        # TODO (Chia Rui): to decide whether make apply_extra_diffusion_on_vn a config parameter or remove it or always turn on extra diffusion
```
muellch: Nice, thank you!

PR #1080, opened 2026-02-26
Try combining test jobs in CI

msimberg:
Just testing to see what the effect is on jobs, job runtimes etc.


msimberg:
I might get back to this someday because right now we generate even more test jobs, but closing this for now.

PR #1084, opened 2026-03-02
Test WEISMAN_KLEMP_TORUS experiment

msimberg:
This is not necessarily for merging. Mostly just to check that it also passes torus tests. We have to discuss if it's worth always testing this experiment.


msimberg:
Closing for now.

PR #1092, opened 2026-03-03
Turn data download off for distributed CI

jcanton:
turn back off data download in distributed CI


jcanton:
cscs-ci run distributed

PR #1093, opened 2026-03-05
Move the mask of _q_t_update outside in graupel

iomaganaris:
I think the mask makes more sense to be outside of the `_q_t_update` function since the output of the functions are updated only if the mask is enabled. Not sure yet if that's necessary for improving performance or can get away with DaCe transformations atm

Needs:
- [x] https://github.com/GridTools/gt4py/pull/2511

### DIFF
--- a/model/atmosphere/advection/pyproject.toml
+++ b/model/atmosphere/advection/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
   # workspace members
   "icon4py-common>=0.0.6",
   # external dependencies
-  "gt4py==1.1.6",
+  "gt4py==1.1.7",
   'packaging>=20.0'
 ]
 description = "ICON advection."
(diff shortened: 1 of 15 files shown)


edopao:
cscs-ci run dace

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/implementations/graupel.py
```diff
@@ -338,12 +338,8 @@ def _q_t_update(  # noqa: PLR0915
     fa.CellKField[ta.wpfloat],
 ]:
     if enable_masking:
```
havogt: I think we should either remove the enable_masking or copy it over to the new location

PR #1065, opened 2026-02-17
Further adv refac

nfarabullini:
Further refactoring of àdvection` such that it's up-to-date with coding standards. Refactoring of lsp coeffs into factories

PR #1099, opened 2026-03-09
Add single-rank lsq pseudoinv factory test

msimberg:
Related to https://github.com/C2SM/icon4py/pull/1065#discussion_r2865941754. This uses @nfarabullini's fixes for doing the LSQ_PSEUDOINV computation with a GPU backend from #1012 to have them merged separately.

This also adds LSQ_PSEUDOINV to the parallel test in test_parallel_grid_manager.py. This required changing the computation to avoid doing the SVD on halo points.

Custom dimensions are added for LSQ_PSEUDOINV to correctly declare the dimensions in the factory. Previously the dimension was equivalne to a scalar, which lead to the factory returning the plain numpy array. With this change the mypy type ignores can be removed from test_parallel_interpolation.py.


model/common/src/icon4py/model/common/math/projection.py
```diff
@@ -38,11 +39,16 @@ def gnomonic_proj(
     TODO:
         replace this with a suitable library call
```
nfarabullini: do we still need this TODO?

PR #540, opened 2024-09-06
Halo construction

halungge:

PR #1100, opened 2026-03-10
Update mpi4py and pymetis groups to make them optional

msimberg:
#540 added mpi4py as required dependency to the typing group. It is in principle required, but this makes mpi4py be installed by default with a regular `uv sync` which is suboptimal. Removing it for now from the typing group. mpi4py is still a required dependency in the distributed group.

### DIFF
--- a/model/common/pyproject.toml
+++ b/model/common/pyproject.toml
@@ -43,10 +43,7 @@ version = "0.0.6"
 all = ["icon4py-common[distributed,io]"]
 cuda11 = ['cupy-cuda11x>=13.0', 'gt4py[cuda11]']
 cuda12 = ['cupy-cuda12x>=13.0', 'gt4py[cuda12]']
-distributed = [
-  "ghex>=0.5.0",
-  "mpi4py>=3.1.5"
-]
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -20,6 +20,14 @@ dev = [
...
   {include-group = "typing"}
 ]
+dev-distributed = [
+  {include-group = "build"},
+  {include-group = "docs"},
+  {include-group = "lint"},
+  {include-group = "test"},
+  {include-group = "typing"},
+  {include-group = "typing-distributed"}
+]
```
msimberg: This doesn't feel right... are there other ways to express this? Alternatively consider making sure @nfarabullini can actually install mpi4py which is what triggered this PR.

edopao:
We need this PR for integration in muphys-ppp, where we do not install MPI:
```
      IndexError: list index out of range

      hint: This usually indicates a problem with the package or the build environment.
  help: `mpi4py` (v4.0.1) was included because `icon4py:dev` (v0.0.6) depends on `mpi4py`
```

havogt:
Can we also try to move pymetis to the distributed? Why is it in the `common[io]` extra? Maybe in a separate PR.

PR #1104, opened 2026-03-11
Update GPU configuration for graupel

iomaganaris:
Set the `gpu_maxnreg` to a lower value. This allows increased occupancy of the GPU (tested on GH200). Even if there is excessive register spilling that's not a problem for graupel because I expect that the register spilling will only happen when the mask is enabled. When the mask is disabled (most of the time) we get the benefit of higher occupancy

### DIFF
--- a/model/atmosphere/advection/pyproject.toml
+++ b/model/atmosphere/advection/pyproject.toml
@@ -26,7 +26,7 @@ dependencies = [
   # workspace members
   "icon4py-common>=0.0.6",
   # external dependencies
-  "gt4py==1.1.6",
+  "gt4py==1.1.7",
   'packaging>=20.0'
 ]
 description = "ICON advection."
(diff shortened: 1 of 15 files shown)


iomaganaris:
`R02B06 main`
```
For 100 iterations it took 0.6509003639221191 seconds!
For 100 iterations it took 0.6452944278717041 seconds!
For 100 iterations it took 0.6502292156219482 seconds!
For 100 iterations it took 0.6481423377990723 seconds!
```
`R02B08 main`
```
For 100 iterations it took 8.613034963607788 seconds!
For 100 iterations it took 8.910228967666626 seconds!
For 100 iterations it took 8.931719303131104 seconds!
For 100 iterations it took 8.839605808258057 seconds!
```
vs
`R02B06 graupel_gpu_opt`
```
For 100 iterations it took 0.6054491996765137 seconds!
For 100 iterations it took 0.6052908897399902 seconds!
For 100 iterations it took 0.5998942852020264 seconds!
For 100 iterations it took 0.5863747596740723 seconds!
```
`R02B08 graupel_gpu_opt`
```
For 100 iterations it took 8.290613174438477 seconds!
For 100 iterations it took 8.158649444580078 seconds!
For 100 iterations it took 8.326894760131836 seconds!
For 100 iterations it took 8.067954063415527 seconds!
```
That's an ~8% increase in performance on average for `R02B06` and ~7% for `R02B08`

PR #1110, opened 2026-03-13
py2fgen: don't recompile if unchanged

havogt:
py2fgen will avoid recompilation (and write of the f90 and py file) if there are no changes. This allows for better integration with tools that detect changes by timestamp.

`--regenerate` will force regeneration/compilation.


havogt:
cscs-ci run default

PR #1052, opened 2026-02-10
Cleanup _dsl fields

jcanton:
WIP for blueline merging.
sync with https://github.com/C2SM/icon-exclaim/pull/412

PR #1112, opened 2026-03-16
Fortran bindings: fix numpy allocation and cleanups

havogt:
Allocate dummy fields for the correct device as they are still passed via the program interface, even if they are not used.


havogt:
> good that you found the issue! in my defense these were already allocated without allocator before my changes in https://github.com/C2SM/icon4py/pull/1052/changes#diff-251b19acad18ff205cb24df810762f3534de9485d95656b8b255f9c87e43b531 😜

yes, I know...

PR #1116, opened 2026-03-18
Enforce StencilTest conventions programmatically

egparedes:
The `StencilTest` conventions were implicit: `reference` and `input_data` had to be written a particular way, but nothing stated it and nothing checked it. A suite that got it subtly wrong — above all, one allocating its input fields without the backend's allocator — did not fail, it quietly did the wrong thing. The framework compensated with workarounds: fixture fields were re-allocated afterwards to correct the device, and connectivities were reached through a shim that misused `Dimension` as an identifier.

This makes the conventions explicit and checks them when a suite class is created, so a malformed suite fails as soon as its module is imported. `reference` and `input_data` get dedicated decorators and neither takes `self`: `reference` receives the `Grid`, with connectivities exposed as an ordinary `Mapping` keyed by `FieldOffset`, and `input_data` receives `data_alloc`, a wrapper with the grid and the backend's allocator already bound. Both are plain fixture arguments, so a suite asks for `grid` only where it uses it. Keeping the wrapper out of `self` also sidesteps a pytest deprecation: class-scoped fixtures defined as instance methods are removed in pytest 10.

Allocation being correct at the source, both workarounds go away. The ~200 existing suites are migrated to match, the module gains docstrings and unit tests, and `data_allocation` is tidied along the way: its constructors now forward to the gt4py ones instead of building arrays by hand, and their signatures agree.

## Note for reviewers of the earlier version

Rebuilt on top of `main` rather than merged, because #1270 rewrote every `reference` signature this PR also rewrites; its keyword-only markers and keyword-style helper calls are preserved. Three parts of the original are gone, having been solved on `main` meanwhile: the `grid/base.py` change and the `@pytest.mark.uses_as_offset` markers (both obsoleted by the GT4Py 1.1.11 update), and the `TODO(havogt)` about keyword-passing connectivities.

## Testing

Full suite on `embedded` and `roundtrip`, the benchmark and GT4Py metrics paths, and pytest 9.1 to confirm no deprecation warning. `gtfn_gpu` and `dace_gpu` stencils pass on CSCS CI.

Device correctness is the point of the change, and dropping the re-allocation exposed eight latent defects in seven fixtures: five wrote a NumPy index pattern into what is now device memory, one zeroed a level through the copy `asnumpy()` returns (so the sweep ran on uninitialised values), one built its field with a gt4py constructor and no allocator, and one passed the same field as both an input and an `out` entry. The host-allocation class is now rejected at import time, so a CPU run catches it too.

---

Includes review suggestions from @edopao in #1111 ([comment](https://github.com/C2SM/icon4py/pull/1111/changes/BASE..fe21507847b5ef0fd2029fe7615ef2c420bb02e3#r2941397453)) and ideas from @iomaganaris in #968.


edopao (commented):
LGTM

iomaganaris (commented):
LGTM in general. Just a few suggestions

model/testing/src/icon4py/model/testing/stencil_tests.py
```diff
@@ -6,175 +6,435 @@
...
+) -> None:
+    """Check that `func` is named `name` and its signature starts with `leading_params`."""
+    if not isinstance(func, allowed_types):
+        raise TypeError(f"The '{name}' method must be {allowed_description} but got {type(func)}.")
+    if func.__name__ != name:
+        raise ValueError(f"The '{name}' method must be named '{name}' but got '{func.__name__}'.")
+    params = tuple(inspect.signature(func).parameters)
+    if params[: len(leading_params)] != leading_params:
+        raise ValueError(
+            f"The '{name}' method signature must be '{name}({', '.join(leading_params)}, ...)'"
```
edopao, line 99: Just a formatting comment: both `leading_params` and `params` are tuples of strings. Then, I assume that the formatting string should be the same. Currently `({', '.join(leading_params)}, ...)` vs `{params}`

egparedes: Good catch, fixed in d38469d15. Both are rendered the same way now:

```
The 'input_data' method signature must be 'input_data(data_alloc, ...)' but got 'input_data(self, grid)'.
```

previously the second half came out as a tuple repr (`input_data('self', 'grid')`).

egparedes:
Addressed the review comments in d38469d15. Three of the four suppressed Copilot comments were real; I reproduced each before fixing it.

**Directly imported constructors slipped through the allocation check.** The check compared against the `data_allocation` module, so `from ...data_allocation import zero_field` bound the function itself and passed validation. It now also matches functions defined in that module.

**Positional-only leading parameters passed the import-time check.** `reference` is called as `reference(grid=...)` and pytest injects `input_data`'s arguments by keyword, so `def reference(grid, /, ...)` validated fine and then failed once the test ran. Now rejected up front.

**`connectivities_asnumpy` violated the `Mapping` contract.** `__getitem__` let the grid's `MissingConnectivityError` (a `ValueError`) escape, so `view.get("missing", default)` raised instead of returning the default, and `in` raised too. It now raises `KeyError`, including for a key that exists but is not a neighbor table, since `__iter__` already omits those.

Each has a regression test, and each was checked against the pre-fix code to confirm it fails there.

**Not fixed: the `Mapping[FieldOffset, ...]` cast in `connectivities_asnumpy`.** That one is a deliberate, documented compromise rather than an oversight. The underlying mapping is keyed by name, and `Mapping` is invariant in its key type, so the honest `Mapping[str | FieldOffset, ...]` is not accepted where the reference helpers ask for `Mapping[FieldOffset, ...]`. Making iteration yield real offsets is not possible today either: `grid.connectivities` is keyed by `str` with no reverse mapping to `FieldOffset`. The alternative is re-annotating the ~55 `*_numpy` helpers, which is worth doing but is a wider change than this PR. The docstring records why the cast is there. Happy to split it out into a follow-up if you would rather it not linger.

The older March comments are all resolved by the rebuild: @havogt's question about the reallocation workaround is answered by removing it (`input_data` fixtures now allocate correctly at the source through `data_alloc`), which also makes @iomaganaris's memory-saving concern and the `device_input_data` naming suggestion moot, since nothing is reallocated any more.

edopao (approved):
LGTM

iomaganaris (approved):
LGTM as well. Memory usage measured on my laptop seems to drop similarly to https://github.com/C2SM/icon4py/pull/968 so this old PR can be closed after this PR is merged

model/testing/src/icon4py/model/testing/stencil_tests.py
```diff
@@ -6,174 +6,479 @@
...
+_REQUIRED_MEMBERS: Final = (
+    ("reference", _STENCIL_REFERENCE_MARKER, "static_reference"),
+    ("input_data", _INPUT_DATA_FIXTURE_MARKER, "input_data_fixture"),
+)
+
+# TODO(iomaganaris, havogt, nfarabullini): tolerance was increased from 1e-7 to 1e-6 to
+# cover floating point discrepancies observed in CI tests. Failing CI can be found in
+# https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504673/-/pipelines/2184694383
+# from PR#861. Reason is probably derivatives of random data. Investigate and lower the
+# tolerance back to 1e-7 if possible.
```
egparedes, line 78: @iomaganaris @havogt @nfarabullini Is this comment still correct or is it an artifact caused by merge conflicts?

iomaganaris: Since this is still there in `main` I expect it is valid. I think at some point we were discussing with @havogt that we could have random data with smooth derivatives so we don't see these discrepancies and we could that way lower the tolerance possibly. Now with LLMs it's easier to try it out. I'll add a project in `icon4py-knowledge`: https://c2sm.github.io/icon4py-knowledge/personal/iomaganaris/smoother-random-data

OngChia: There are many stencil tests that use randomly chosen hardcoded values. For examples, some interpolation coeff are only meaningful when they are between 0 and 1. Scalar inputs are also randomly decided in the stencil tests.

OngChia (approved):
I can only understand how the stencil test works now. The new implementation looks good and cleaner than before to me. I can't judge more than that.

egparedes:
cscs-ci run default;BACKENDS=gtfn_cpu;MODEL_SUBSETS=stencils

edopao:
cscs-ci run default;BACKENDS=gtfn_cpu;MODEL_SUBSETS=stencils

egparedes:
cscs-ci run default;BACKENDS=gtfn_gpu;MODEL_SUBSETS=stencils

egparedes:
cscs-ci run default;BACKENDS=gtfn_gpu;MODEL_SUBSETS=stencils

egparedes:
cscs-ci run default;BACKENDS=gtfn_gpu;MODEL_SUBSETS=stencils

egparedes:
cscs-ci run default;BACKENDS=dace_gpu;MODEL_SUBSETS=stencils

PR #1117, opened 2026-03-19
Remove two slow test grids from `test_global_grid_params_from_grid_manager`

msimberg:
The two removed test grids run in 10 and 50 seconds each because of the size of the grids. All the other grids for that test run in at most a couple of seconds. Given that the test is checking basic properties of creating global grid parameters this feels like overkill. Since the other test cases cover toruses and global grids well I'm simply proposing to remove them. I don't believe we get important test coverage with the two removed grids.


msimberg:
cscs-ci run default

PR #819, opened 2025-07-30
Try node sharing

havogt:

PR #1121, opened 2026-03-23
Use only a single GPU/numa node for distributed tests

msimberg:
Since tests are currently anyway serialized, I think we don't benefit from running across the full node. This uses MPS to run all four ranks on a single GPU.  This is based on #819.


edopao (approved):
LGTM

PR #1151, opened 2026-03-26
Script to bump all package versions

havogt:
In the future we might want to release packages individually and won't need this script anymore.

Disclaimer: AI generated...


msimberg:
This is effectively the same as https://github.com/C2SM/icon4py/blob/main/scripts/python/bump_versions.py? Can we close this?

havogt:
Yes

PR #1186, opened 2026-03-31
remove the duplication of fixtures

yiluchen1066:
- Removed duplicate ```timeloop_diffusion_savepoint_exit (driver)``` and ```timeloop_diffusion_savepoint_exit_standalone (standalone_driver) ```
  fixtures that were identical to the shared ```savepoint_diffusion_exit``` in ```datatest.py```
 - Added a small ```linit``` fixture alias in both driver and standalone_driver to bridge the parametrized ```timeloop_diffusion_linit_exit``` name
   to the ```linit``` name expected by the shared fixture


yiluchen1066:
> I'm not sure but: there already is a `linit` fixture in `model/testing/datatest.py` why not use that directly instead of the alias?

the `linit` fixture in `datatest.py` always return False. The driver and the standalone driver tests need to override that value via `@pytest.mark.parameterize("timeloop_diffusion_linit_exit", ...)`. So I added this alias fixture to do this override. Withouth the alias, `savepoint_diffusion_exit` would use `linit=False` regardless of what `timeloop_diffusion_linit_exit` is set in the parametrize decorator.

yiluchen1066:
cscs-ci run default

yiluchen1066:
cscs-ci run distributed

jcanton:
> > I'm not sure but: there already is a `linit` fixture in `model/testing/datatest.py` why not use that directly instead of the alias?
> 
> the `linit` fixture in `datatest.py` always return False. The driver and the standalone driver tests need to override that value via `@pytest.mark.parameterize("timeloop_diffusion_linit_exit", ...)`. So I added this alias fixture to do this override. Withouth the alias, `savepoint_diffusion_exit` would use `linit=False` regardless of what `timeloop_diffusion_linit_exit` is set in the parametrize decorator.

and it's not possible to change the `linit` fixture in `datatest.py` to return True/False?

yiluchen1066:
> > > I'm not sure but: there already is a `linit` fixture in `model/testing/datatest.py` why not use that directly instead of the alias?
> > 
> > 
> > the `linit` fixture in `datatest.py` always return False. The driver and the standalone driver tests need to override that value via `@pytest.mark.parameterize("timeloop_diffusion_linit_exit", ...)`. So I added this alias fixture to do this override. Withouth the alias, `savepoint_diffusion_exit` would use `linit=False` regardless of what `timeloop_diffusion_linit_exit` is set in the parametrize decorator.
> 
> and it's not possible to change the `linit` fixture in `datatest.py` to return True/False?

technically it could be made parametrizable, but I think the issue is that `linit` in `datatest.py` serves as a default for all diffusion tests. Most diffusion tests expect `linit=False (a normal timestep)`, and only a few specific tests override it to True via @pytest.mark.parametrize("linit", [True]).  If we changed the linit fixture in datatest.py to return both True and False, it would affect every test that depends on it

yiluchen1066:
cscs-ci run default

jcanton (approved):
nice! thanks for the cleanup and clarifications!

PR #1192, opened 2026-04-13
Added smagorinsky factors and heights

muellch:


bindings/src/icon4py/bindings/diffusion_wrapper.py
vectorflux, line 1: All good, as agreed upon.

bindings/tests/bindings/references/diffusion/diffusion.f90
vectorflux, line 1: Consistent with diffusion_wrapper.py

bindings/tests/bindings/references/diffusion/diffusion.h
vectorflux, line 1: Consistent with diffusion_wrapper.py

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -209,38 +218,38 @@ def __init__(
         #: Called 'hdiff_w_efdt_ratio' in mo_diffusion_nml.f90.
         self.hdiff_w_efdt_ratio: float = hdiff_w_efdt_ratio
 
-        #: Scaling factor for Smagorinsky diffusion at height smagorinski_scaling_height and below
+        #: Smagorinsky factor for z <= smagorinski_scaling_height (constant base value)
```
vectorflux, line 222: Where are these updated comments coming from? Directly from ICON, or a clarification from you?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -927,17 +889,36 @@ def _do_diffusion_step(
...
+            if initial_run or self.config.iforcing not in (ForcingType.NWP, ForcingType.AES):
+                log.debug("communication of prognostic cell field: w - start")
+                self._exchange.exchange(
+                    dims.CellDim,
+                    prognostic_state.w,
+                    prognostic_state.theta_v,
+                    prognostic_state.exner,
+                    stream=decomposition.DEFAULT_STREAM,
+                )
+                log.debug("communication of prognostic cell field: w - done")
```
vectorflux: Taking the initial_run inside makes this more elegant, but I can't judge if the logic is correct.

vectorflux (approved):
This all looks OK to me. Another check of the logic for `initial_run` might be helpful.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
vectorflux: Copilot says:

`The new code will skip cell field communication during regular timesteps when using NWP or AES forcing, which changes model behavior compared to the old code that always performed this communication. This could affect the physical results of the simulation in those forcing modes.`

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Claude wrote them by looking at diffusion in ICON. I reviewed them.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Yes, this is a behavior change. But now we are doing the same as ICON Fortran.
This was discussed with Chia Rui and Jacopo.

To inspect the ICON Fortran code, see:
https://github.com/C2SM/icon-exclaim/blob/8b73908e1dfb5755565ce8f1b78917013d7eaaf8/src/atm_dyn_iconam/mo_nh_diffusion.f90#L1624
https://github.com/C2SM/icon-exclaim/blob/8b73908e1dfb5755565ce8f1b78917013d7eaaf8/src/atm_dyn_iconam/mo_nh_diffusion.f90#L1632

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -329,53 +387,26 @@ def __post_init__(self, config):
...
+        object.__setattr__(
+            self,
+            "smagorinski_factor",
+            (
+                config.smagorinski_scaling_factor,
+                config.smagorinski_scaling_factor2,
+                config.smagorinski_scaling_factor3,
+                config.smagorinski_scaling_factor4,
+            ),
+        )
```
havogt, line 400: something is wrong here, the dataclass says this field is a float, but we are assigning a tuple.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: @havogt Here I am putting a new halo exchange before a `halo_exchange_wait`. Is this ok or a bug? If it is a bug we need to move the wait up.

havogt (changes requested):

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
havogt: @msimberg forwarding to you. Here we are placing a full exchange between a schedule and wait of another exchange.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Done.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -149,14 +156,26 @@ def __init__(
...
+        smagorinski_scaling_factor2: float = 2e-6
+        * (1600.0 + 25000.0 + math.sqrt(1600.0 * (1600 + 50000.0))),
+        smagorinski_scaling_factor3: float = 0,
+        smagorinski_scaling_factor4: float = 1,
+        smagorinski_scaling_height: float = 32500.0,
+        smagorinski_scaling_height2: float = 1600.0
+        + 50000.0
+        + math.sqrt(1600.0 * (1600 + 50000.0)),
+        smagorinski_scaling_height3: float = 50000.0,
+        smagorinski_scaling_height4: float = 90000.0,
```
havogt, line 168: Why is it a tuple in DiffusionParams, but not here?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -329,53 +387,26 @@ def __post_init__(self, config):
...
+        object.__setattr__(
+            self,
+            "smagorinski_height",
+            (
+                config.smagorinski_scaling_height,
+                config.smagorinski_scaling_height2,
+                config.smagorinski_scaling_height3,
+                config.smagorinski_scaling_height4,
+            ),
+        )
```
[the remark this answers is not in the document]
muellch: This should be fixed already, can you check?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -310,6 +315,11 @@ def _validate(self):
                 "Only Klemp type of the Rayleigh damping (nudging vertical wind towards zero) is implemented."
             )
 
+        if not self.extra_diffu:
+            raise NotImplementedError(
+                "extra_diffu=False is not supported; only True is implemented"
+            )
```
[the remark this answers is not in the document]
muellch: This is intentional and we do this with many other parameters. We want to capture all namespace variables on the Fortran side but reject those we do not support on the Python side.
This should give good user error messages and allows us later to implement the missing features without touching the interface.

havogt, line 322: I'd actually consider a slightly different approach for options we will never support. Instead of rejecting here, we could do it in the Fortran bindings. What do you thing?

OngChia: Both hands up! I agree!

muellch: I think we might support it because it is very very easy to support?
I would like to deal with the whole namelist evaluation and rejection in one place, not two places.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Oh you want it to be a tuple already in the `__init__`?
I guess that is fine, I did not think that far.
We then should change the dycore as well though, with the divdamp factors and divdamp heights.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: @vectorflux You were right that something is strange with my halo exchanges, now should be good.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -913,17 +889,36 @@ def _do_diffusion_step(
...
+
+        if initial_run or self.config.iforcing not in (ForcingType.NWP, ForcingType.AES):
+            log.debug("communication of prognostic cell fields: theta and exner - start")
+            self._exchange.exchange(
+                dims.CellDim,
+                prognostic_state.theta_v,
+                prognostic_state.exner,
+                stream=decomposition.DEFAULT_STREAM,
+            )
+            log.debug("communication of prognostic cell fields: theta and exner - done")
```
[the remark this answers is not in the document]
muellch: I updated the halo exchange behavior, it should now be the same as in Fortran ICON.

To inspect the ICON Fortran code, see:
https://github.com/C2SM/icon-exclaim/blob/8b73908e1dfb5755565ce8f1b78917013d7eaaf8/src/atm_dyn_iconam/mo_nh_diffusion.f90#L1624
https://github.com/C2SM/icon-exclaim/blob/8b73908e1dfb5755565ce8f1b78917013d7eaaf8/src/atm_dyn_iconam/mo_nh_diffusion.f90#L1632

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -913,17 +889,36 @@ def _do_diffusion_step(
...
+            if initial_run or self.config.iforcing not in (ForcingType.NWP, ForcingType.AES):
+                log.debug("communication of prognostic cell field: w - start")
+                self._exchange.exchange(
+                    dims.CellDim,
+                    prognostic_state.w,
+                    prognostic_state.theta_v,
+                    prognostic_state.exner,
+                    stream=decomposition.DEFAULT_STREAM,
+                )
+                log.debug("communication of prognostic cell field: w - done")
```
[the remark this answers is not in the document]
muellch: I moved the wait up so they do no longer overlap, have a look please.

bindings/src/icon4py/bindings/dycore_wrapper.py
```diff
@@ -112,6 +112,7 @@ def solve_nh_init(
     l_vert_nested: bool,
     ldeepatmo: bool,
     iau_init: bool,
+    extra_diffu: bool,
```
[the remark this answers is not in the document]
muellch: I do not understand, I set the default of `extra_diffu` to `True` in `solve_nh_init. Can you check?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -149,14 +156,26 @@ def __init__(
...
+        smagorinski_scaling_height4: float = 90000.0,
         n_substeps: int = 5,
         zdiffu_t: bool = True,
         velocity_boundary_diffusion_denom: float = 200.0,
         temperature_boundary_diffusion_denom: float = 135.0,
         _nudge_max_coeff: float | None = None,  # default is set in __init__
         max_nudging_coefficient: float | None = None,  # default is set in __init__
         shear_type: TurbulenceShearForcingType = TurbulenceShearForcingType.VERTICAL_OF_HORIZONTAL_WIND,
-        ltkeshs: bool = True,
+        iforcing: ForcingType = ForcingType.NO_FORCING,
```
OngChia, line 176: Instead of putting `iforcing` here in the diffusion config, I am thinking it is better to instead have an option called `do_sync `(we can think of a better name), and convert `self.config.iforcing not in (ForcingType.NWP, ForcingType.AES)` to `do_sync `in the wrapper for better code quality in diffusion granule.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -695,96 +726,40 @@ def _get_start_index_for_w_diffusion() -> gtx.int32:
...
         self._horizontal_start_index_w_diffusion = _get_start_index_for_w_diffusion()
 
-    def initial_run(
+    def run(
         self,
         diagnostic_state: diffusion_states.DiffusionDiagnosticState,
         prognostic_state: prognostics.PrognosticState,
         dtime: float,
+        initial_run: bool = False,
     ):
```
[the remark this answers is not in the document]
muellch: There should not be any more calls to `initial_run` in the repo

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: We also need to mix `initial_run` into that boolean. :) 
I do not know if I like it or not since it hides the true condition.
I am undecided.

muellch:
cscs-ci run default

muellch:
cscs-ci run distributed

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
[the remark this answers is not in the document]
havogt: @copilot can you double-check?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -897,11 +874,16 @@ def _do_diffusion_step(
             "running stencils 07 08 09 10 (apply_diffusion_to_w_and_compute_horizontal_gradients_for_turbulence): end"
         )
 
+        self.halo_exchange_wait(
+            handle_edge_comm,
+            # stream=decomposition.DEFAULT_STREAM,  # noqa: ERA001  # See NOTE above.
+        )  # need to do this here, since we currently only use 1 communication object.
```
msimberg, line 881: Holy indirection, I think this could just be `handle_edge_comm.finish()`? Not insisting on a change though because this `halo_exchange_wait` already exists and I'm not 100% confident I'm following the layers of abstraction in the halo exchanges.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -312,8 +370,8 @@ class DiffusionParams:
     K4: Final[float] = dataclasses.field(init=False)
     K6: Final[float] = dataclasses.field(init=False)
     K4W: Final[float] = dataclasses.field(init=False)
-    smagorinski_factor: Final[float] = dataclasses.field(init=False)
-    smagorinski_height: Final[float] = dataclasses.field(init=False)
+    smagorinski_factor: Final[tuple[float, float, float, float]] = dataclasses.field(init=False)
+    smagorinski_height: Final[tuple[float, float, float, float]] = dataclasses.field(init=False)
```
OngChia, line 375: These actually should be wpfloat, right? 
It can be cleaned up by Claude with another PR.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
msimberg: Yeah, mixing a full exchange between start and end of an async exchange would be a bug (best case it happens to work, but it's not intended to work). If halo exchanges need to be overlapped like that we'll need to set up two communication objects, but from the latest changes it looks like we're fine just waiting for the first halo exchange to finish.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Yeah I just moved it so there is no more overlap, did not touch it.

muellch:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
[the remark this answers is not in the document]
muellch: @copilot Please refer to my message above, does this address your concern?

bindings/src/icon4py/bindings/dycore_wrapper.py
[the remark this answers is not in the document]
muellch: @copilot I do not understand, I set the default of extra_diffu to True in `solve_nh_init. Can you check?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
jcanton: I agree with @OngChia : cleaner to have it in the config where you can write a line of comment about it and call them `perform_w_echange_at_end` `perform_theta_v_exner_echange_at_end`
cleaner to pass these (or some rename of them) around instead of iforcing which makes you wonder why iforcing is ending up in diffusion 🤔

muellch:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: So you want to define a boolean for `perform_exchange_at_end` which is `self.config.iforcing not in (ForcingType.NWP, ForcingType.AES)`, and then the expression will be `initial_run or perform_exchange_at_end'?
I guess at the very least we need a better name.

muellch:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
jcanton: no, what I meant is to define a bool which is `wanderful_name_here = IF ( linit .OR. (iforcing /= inwp .AND. iforcing /= iaes) )`
which communicates what is happeninig (the w / theta_v_exner exchanges) and not why.
The why becomes clear where the bool is defined (in DiffusionConfig) with the if and with a line of comment all in one place (also adding that this is there only because in the case of running slow physics but not fast physics in a idealized simulation you can skip an exchange).
So actually I'm much more in favour of removing this and always performing the exchange, because the case where it can be skipped is so rare / useless and making the code more difficult to read just for that seems unnecessary.

muellch:
cscs-ci run distributed

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: `linit` is a runtime parameter given to the `run` function. It cannot be defined in the config? Am I missing something?

muellch:
cscs-ci run distributed

muellch:
cscs-ci run default

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: I will wait for Philipp to come back, we will resolve this when merging the single precision PR.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Had a conversation with Jacopo, since `initial_run` is a runtime parameter we settled for comments explaining why the halo exchange sometimes can be skipped.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
muellch: Added TODO, separate PR, then also for solve_nonhydro.

muellch:
cscs-ci run distributed

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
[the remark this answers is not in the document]
havogt: @copilot do the latest changes address your concern?

bindings/src/icon4py/bindings/dycore_wrapper.py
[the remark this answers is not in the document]
havogt: tests would catch this if there is a problem. I'll close

muellch:
cscs-ci run default

muellch:
cscs-ci run distributed

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
havogt: We should only add flags here that we want for greenline. Let's leave the check here for now, but if we don't implement the option we should move it to the bindings.

havogt (approved):
lgtm

PR #1198, opened 2026-04-15
Vectorize static field computations

OngChia:
Static field vectorization based on Claude's work in this branch https://github.com/C2SM/icon4py/compare/main...scientific_validation_distributed_driver_hannes

Example of speed up on the R2B7 grid (the number can fluctuation ~15%):
image


OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run distributed

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -689,6 +689,17 @@ def _create_inverse_neighbor_index(
...
+        assume a and b are dimensions
+        source_offset = [a, a2b]
+        inverse_offset = [b, b2a]
+        loop x:
+            inv_neighbor = inverse_offset[:, x] # shape (b,)
+            src_neighbor = source_offset[inv_neighbor, :] # shape (a, a2b)
+            src_neighbor == inv_neighbor[:, None] # shape bool(a, a2b), a2b == b
+
+
+    inv_neighbor =
```
msimberg: Does this belong here? If yes, it looks incomplete.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -699,13 +710,28 @@ def _create_inverse_neighbor_index(
 
     """
     inv_neighbor_idx = MISSING * array_ns.ones(inverse_offset.shape, dtype=gtx.int32)
```
msimberg: We do this in many places so doesn't have to be changed now, but this could be:
```suggestion
    inv_neighbor_idx = array_ns.full(MISSING, inverse_offset.shape, dtype=gtx.int32)
```
If missing is an enum, it may have be `MISSING.value` to work with cupy.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -448,14 +446,19 @@ def index_offset(f):
...
+        group_idx = np.where(n_valid == nv)[0]
+        valid_cols = np.arange(nv)
+        mat_batch = z_rbfmat_np[np.ix_(group_idx, valid_cols, valid_cols)]
         for j in range(num_zonal_meridional_components):
-            rbf_vec_coeff_np[j][i + horizontal_start, valid_neighbors] = sla.cho_solve(
-                z_diag_np, rhs_np[j][i, valid_neighbors]
-            )
+            rhs_batch = rhs_np[j][np.ix_(group_idx, valid_cols)]
+            sol = np.linalg.solve(mat_batch, rhs_batch[:, :, np.newaxis])[:, :, 0]
+            rbf_vec_coeff_np[j][group_idx + horizontal_start, :nv] = sol
```
msimberg: This is indeed clever, but may need a comment. I'm not sure I fully follow, but is `np.linalg.solve` used because it provides a batched interface unlike `scipy.cho_solve/cho_factor`? If I'm not completely mistaken the new version does a full linear solve and does not take into account that `z_rbfmat_np` (I may be forgetting details) is symmetric which is why we can do a cholesky decomposition and solve using that instead. In theory each individual solve should be faster because of the symmetry, plus the decomposed matrix can be used for `num_zonal_meridional_components` (most of the time 2, 1 for one of the dimensions). In practice, the batching probably easily makes this faster overall. But it is different from icon fortran, so I think it deserves a comment.

An alternative would be to somehow utilize one of the thread pools that numpy/something sets up and spread out the cho_factor/cho_solve across the pool.

msimberg (commented):
There's a lot of indexing smartness that I'm not following (the comments definitely help, thanks @OngChia!) and need some time to digest. I'd tend to generally trust this if we have decent test coverage for the changes. 

I commented for now on the RBF changes (probably good as suggested, but also benefits from comments).

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -617,12 +617,23 @@ def _construct_diamond_vertices(
...
+    # find the first and second neighboring index of e2c2v that belongs to far indices (v0 and v2)
+    far_index_1 = array_ns.argmax(ascending_far_indices == 1, axis=1)
+    far_index_2 = array_ns.argmax(ascending_far_indices == 2, axis=1)
+    e2v_far[:, 0] = array_ns.squeeze(
+        array_ns.take_along_axis(flat, far_index_1[:, array_ns.newaxis], axis=1)
+    )
+    e2v_far[:, 1] = array_ns.squeeze(
+        array_ns.take_along_axis(flat, far_index_2[:, array_ns.newaxis], axis=1)
+    )
+    return array_ns.hstack((e2v, e2v_far))
```
nfarabullini, line 628: why do we need all of this mumbo jumbo?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Yes. Thanks for noticing the incompleteness of the docstring. I tried to write down the numpy/cupy implementation detail because I find the way that we compute the inverse neighbor table is not intuitively straightforward. (I actually find that explicitly writing down the do loop as in the Fortran code is easier to read for this computation)

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Is `MISSING` going to be enum and is this clean-up issue already in the github issues? Maybe we can add a new issue there.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: I asked Claude, it said that `np.linalg.solve` is preferred because `scipy.cho_solve` does not work with batching. And using `np.linalg.sove `is the most efficient way. I will post it full answer later.

model/common/src/icon4py/model/common/grid/grid_manager.py
OngChia: Indeed you are right. I worked with Claude, and shortened the code a bit. However, the basic logic is still the same. I cannot find a better way to compute e2c2v.

nfarabullini: that's fine. We don't usually write long comments, can you restrict them to one lines or remove them?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
msimberg: I don't know about `MISSING` being an enum (it probably isn't, and probably doesn't need to be one now that I look at the definition). But good point on issue: I opened https://github.com/C2SM/icon4py/issues/1203. Probably quite straightforward for an agent to fix.

msimberg: And thanks for editing this one!

OngChia:
cscs-ci run distributed

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -606,6 +606,9 @@ def _construct_diamond_vertices(
 
     For example for this diamond: e0 -> (v0, v1, v2, v3)
     Ordering is the same as ICON uses.
+    
+    Implementation:
+
```
nfarabullini: some text is missing here

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
-    flat = expanded.reshape(sh[0], sh[1] * sh[2])
-    far_indices = array_ns.zeros_like(e2v)
-    # TODO(halungge): vectorize speed this up?
-    for i in range(sh[0]):
-        far_indices[i, :] = flat[i, ~array_ns.isin(flat[i, :], e2v[i, :])][:2]
-    return array_ns.hstack((e2v, far_indices))
+    # `flat` includes duplicated e2v vertices (v1, v3), shape (n_edges, 6).
+    flat = expanded.reshape(expanded.shape[0], -1)
+
+    # create a mask to find which vertices in flat are not v1 or v3, this will result in a mask of shape (n_edges, 6)
```
nfarabullini:
```suggestion
```
code is self-explanatory

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
+    far_indices_mask = (flat != e2v[:, 0, array_ns.newaxis]) & (flat != e2v[:, 1, array_ns.newaxis])
+
+    # create a 1d ascending array of length 6
```
nfarabullini:
```suggestion
```
Also self-explanatory

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
+    col_idx = array_ns.arange(flat.shape[1], dtype=gtx.int32)
+    # identify the position of v0 and v2 vertices with the mask, and the position of v1 and v3 vertices is replaced with 6
```
nfarabullini:
```suggestion
    # identify v0 and v2 positions with the mask, and replace v1 and v3 positions with 6
```

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
+    far_indices_pos = array_ns.where(far_indices_mask, col_idx, col_idx.shape[0])  # (n_edges, 6)
+    # sort the far_indices_pos along the second axis and take the first two numbers
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
+    far_indices_pos = array_ns.sort(far_indices_pos, axis=1)[:, :2]  # (n_edges, 2)
```
nfarabullini:
```suggestion
    far_indices_pos = array_ns.sort(far_indices_pos, axis=1)[:, :2]  
```

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
     """
     dummy_c2v = _patch_with_dummy_lastline(c2v, array_ns=array_ns)
     expanded = dummy_c2v[e2c, :]
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -616,13 +619,20 @@ def _construct_diamond_vertices(
...
-    sh = expanded.shape
-    flat = expanded.reshape(sh[0], sh[1] * sh[2])
-    far_indices = array_ns.zeros_like(e2v)
-    # TODO(halungge): vectorize speed this up?
-    for i in range(sh[0]):
-        far_indices[i, :] = flat[i, ~array_ns.isin(flat[i, :], e2v[i, :])][:2]
-    return array_ns.hstack((e2v, far_indices))
+    # `flat` includes duplicated e2v vertices (v1, v3), shape (n_edges, 6).
+    flat = expanded.reshape(expanded.shape[0], -1)
```
nfarabullini:
```suggestion
    flat = expanded.reshape(dummy_c2v[e2c, :].shape[0], -1)
```
This way we reduce number of lines

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -689,23 +689,50 @@ def _create_inverse_neighbor_index(
...
+        ]  # (n_valid, n_source_neighbors)
+        # find for each valid neighbor of an inverse_offset element, which neighbor of the source offset matches the index of the inverse neighbor.
+        inv_elem_indices = array_ns.arange(n_inv_elem)[valid_mask]
+        matches = array_ns.equal(
+            valid_source_offset, inv_elem_indices[:, array_ns.newaxis]
+        )  # (n_valid, n_source_neighbors)
+        # we need has_match here before argmax, because argmax will return 0 if there is no match, which is a valid index, but we want to ignore it in that case
+        has_match = array_ns.any(matches, axis=1)
+        first_match = array_ns.argmax(matches, axis=1)
+        inv_neighbor_idx[inv_elem_indices[has_match], i] = first_match[has_match].astype(gtx.int32)
```
nfarabullini, line 724: clean this code up of all of these comments as well

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -689,23 +689,50 @@ def _create_inverse_neighbor_index(
...
+        inverse_neighbor = inverse_offset[:, i]  # (n_inv_elem,)
+        valid_mask = inverse_neighbor >= 0
+        if not array_ns.any(valid_mask):
+            continue
+        # get all elements of inverse_offset that have a valid neighbor
+        valid_inverse_neighbor = inverse_neighbor[valid_mask]
+        # get the neighbors of those valid neighbors from the source offset
+        valid_source_offset = source_offset[
+            valid_inverse_neighbor, :
+        ]  # (n_valid, n_source_neighbors)
```
```suggestion
        valid_source_offset = source_offset[
            inverse_neighbor[valid_mask], :
        ] 
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
nfarabullini, line 724: do we need type cast to gtx.int32 here? Shouldn't values be integers by default since they're indexes?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -910,34 +937,56 @@ def compute_cells_aw_verts(
...
+        valid_v2c[:, jc] = (
+            v2c[:, jc] != MISSING
+            if jc == 0
+            else (v2c[:, jc] != MISSING) & (v2c[:, jc] != v2c[:, jc - 1])
+        )
+
+    for je in range(num_edges_per_vert):
+        # edges for all vertices in this slot
+        v2e_edge = v2e[horizontal_start:, je]  # (n_verts_active,)
+        vertex_range = array_ns.arange(horizontal_start, num_verts)
```
nfarabullini: this variable is the same irrespective of `je`, it can be moved outside

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1164,48 +1212,69 @@ def compute_lsq_pseudoinv(
...
-        for jjk in range(lsq_dim_unk):
-            for jc in range(start_idx, min_rlcell_int):
-                if cell_owner_mask[jc]:
-                    u, s, v_t = array_ns.linalg.svd(z_lsq_mat_c[jc, :, :])
-                    lsq_pseudoinv[jc, :lsq_dim_unk, jjb] = (
-                        lsq_pseudoinv[jc, :lsq_dim_unk, jjb]
-                        + v_t[jjk, :lsq_dim_unk] / s[jjk] * u[jjb, jjk] * lsq_weights_c[jc, jjb]
-                    )
+    cell_size = cell_owner_mask.shape[0]
+    cell_sequence = array_ns.arange(cell_size)
```
nfarabullini, line 1225:
```suggestion
    cell_sequence = array_ns.arange(cell_owner_mask.shape[0])
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1164,48 +1212,69 @@ def compute_lsq_pseudoinv(
...
     z_dist_g: data_alloc.NDArray,
-    jc: int,
+    start_idx: int,
+    min_rlcell_int: int,
     lsq_dim_unk: int,
     lsq_dim_c: int,
+    array_ns: ModuleType = np,
 ) -> data_alloc.NDArray:
+    cell_size = cell_owner_mask.shape[0]
+    cell_sequence = array_ns.arange(cell_size)
```
nfarabullini, line 1267:
```suggestion
    cell_sequence = array_ns.arange(cell_owner_mask.shape[0])
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1164,48 +1212,69 @@ def compute_lsq_pseudoinv(
...
-        z_lsq_mat_c[jc, js, :lsq_dim_unk] = lsq_weights_c[jc, js] * z_dist_g[js, :]
+    z_lsq_mat_c = array_ns.zeros((cell_size, lsq_dim_c, lsq_dim_c), dtype=ta.wpfloat)
+    z_lsq_mat_c[valid_cell_mask, :min_lsq_bound, :min_lsq_bound] = 1.0
 
-    return z_lsq_mat_c[jc, js, :lsq_dim_unk]
+    for c_idx in range(lsq_dim_c):
+        z_lsq_mat_c[valid_cell_mask_with_halo, c_idx, :lsq_dim_unk] = (
+            lsq_weights_c[valid_cell_mask_with_halo, c_idx, array_ns.newaxis]
+            * z_dist_g[valid_cell_mask_with_halo, c_idx, :]
+        )
```
```suggestion
    z_lsq_mat_c[valid_cell_mask_with_halo,:lsq_dim_c, :lsq_dim_unk] = (
        lsq_weights_c[valid_cell_mask_with_halo, :lsq_dim_c, array_ns.newaxis]
        * z_dist_g[valid_cell_mask_with_halo, :lsq_dim_c, :]
    )
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1228,10 +1297,8 @@ def compute_lsq_coeffs(
     exchange: Callable[[data_alloc.NDArray], None] = decomposition.single_node_exchange,
     array_ns: ModuleType = np,
 ) -> data_alloc.NDArray:
-    lsq_weights_c = array_ns.zeros((min_rlcell_int, lsq_dim_stencil))
-    lsq_pseudoinv = array_ns.zeros((min_rlcell_int, lsq_dim_unk, lsq_dim_c))
-    z_lsq_mat_c = array_ns.zeros((min_rlcell_int, lsq_dim_c, lsq_dim_c))
-    z_dist_g = array_ns.zeros((min_rlcell_int, lsq_dim_c, 2))
+    cell_size = cell_owner_mask.shape[0]
+    z_dist_g = array_ns.zeros((cell_size, lsq_dim_c, 2), dtype=ta.wpfloat)
```
```suggestion
    z_dist_g = array_ns.zeros((cell_owner_mask.shape[0], lsq_dim_c, 2), dtype=ta.wpfloat)
```

nfarabullini (changes requested):
Please remove the poems of comments that Claude wrote. After these edits are tackled, I'll continue with more review

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Let me look at the matrix properties and the common package's dependencies to give a precise answer.  

Read [](file:///capstor/scratch/cscs/cong/icon4py/model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py#395-395), lines 395 to 470

Read [](file:///capstor/scratch/cscs/cong/icon4py/model/common/pyproject.toml#1-1), lines 1 to 80

Here is the full picture across three dimensions: matrix structure, API availability, and practical performance.

---

## Matrix structure — is Cholesky applicable?

`z_rbfmat` is the **Hadamard (element-wise) product** of two matrices:

```
z_rbfmat = z_nxprod ⊙ kernel(z_dist)
```

- `z_nxprod[i,j,k] = n_j · n_k` — a **Gram matrix**, hence symmetric positive semi-definite.
- `kernel(z_dist)` — a Gaussian or inverse-multiquadratic RBF kernel matrix over distinct points, which is strictly positive definite (Mercer condition).

By the **Schur product theorem**, the Hadamard product of two PSD matrices is PSD. When both factors have full rank (distinct points, non-degenerate normals — the normal operating regime), the product is **symmetric positive definite**. This is why the original ICON Fortran code uses Cholesky for exactly this system.

---

## `scipy.linalg.cho_solve` — not a viable batch replacement

`scipy.linalg.cho_factor` / `cho_solve` **only accept 2D inputs**. They have no batched interface for stacked 3D arrays. Using them would require a Python loop over every element or every group, which destroys the batch efficiency the current code achieves via `numpy.linalg.solve(mat_3d, rhs_3d)`. The comment about this in the code (line 440) refers precisely to this limitation on the CuPy side; the same limitation applies to SciPy on the CPU side.

---

## The right Cholesky-based approach: `scipy.linalg.solve(..., assume_a='pos')`

`scipy.linalg.solve` supports **stacked/batched 3D inputs** and accepts `assume_a='pos'`, which routes internally to LAPACK's `dpotrf`/`dpotrs` (Cholesky) instead of `dgesv` (LU):

```python
# drop-in replacement for the inner loop:
sol = scipy.linalg.solve(mat_batch, rhs_batch, assume_a='pos')
```

This is structurally feasible — but has two blockers in this codebase:

1. **SciPy is not a runtime dependency of `icon4py-common`** — only `scipy-stubs` (typing only) is listed in the root pyproject.toml. Adding it would be a non-trivial dependency bump.
2. **The matrices are tiny** — 4×4 (edges), 6×6 (vertices), 9×9 (cells). Cholesky saves ~1/3 of the FLOPs over LU for the factorization of an n×n matrix (n³/3 vs 2n³/3), but at n=4–9 this is completely dominated by NumPy/Python call overhead. The batch solve over `~O(n_cells)` matrices is memory-throughput-bound, not compute-bound, so the algorithmic advantage does not materialise in practice.

---

## Summary

| Approach | Batched API | SPD exploitation | Practical speedup | Feasible today |
|---|---|---|---|---|
| `scipy.linalg.cho_factor` + `cho_solve` | No (2D only) | Yes | Negative (forces loop) | No |
| `scipy.linalg.solve(..., assume_a='pos')` | Yes (3D stacked) | Yes (Cholesky internally) | Negligible for 4–9×4–9 | Only if SciPy added as dependency |
| Current `numpy.linalg.solve` (batched) | Yes | No (LU) | — (baseline) | Yes |

**Bottom line**: `scipy.cho_solve` itself cannot match the batch efficiency. `scipy.linalg.solve(..., assume_a='pos')` would be the correct Cholesky-based equivalent and has the right batched interface, but for 4–9×4–9 matrices the theoretical ~2x Cholesky FLOPs advantage is not observable — the current grouped `numpy.linalg.solve` is already essentially as fast as anything you can do on CPU for this problem size.

OngChia: So it seems that we can also use `scipy.linalg.solve` with Cholesky internally for batched solving.

nfarabullini: no because scipy is numpy based, which doesn't work for gpu backends which use cupy

OngChia: Maybe I am wrong, @msimberg what is the reason these rbf coefficients are computed on CPU?

nfarabullini: ah if you only use numpy, then it's fine. Sorry, I took for granted you were running these on both gpu and cpu

OngChia: Is it because we want to make use of the symmetric property of the matrix and perform the Cholesky decomposition same as what ICON does?

OngChia:
> Please remove the poems of comments that Claude wrote. After these edits are tackled, I'll continue with more review

No, actually I wrote 90% of the comments except for the `get ***` and some comments with the array shape. I agree that I have overdone it, but some numpy operations are too difficult for me to understand without a comment. I will probably forget why it was written in that way in 3 months if I do not have a comment now. A good example is `compute_e_flx_avg` and `global_to_local`. They are "clean" and "concise", without any comments. But when something happens, it becomes a difficult task to debug (I guess only Claude can understand what the code does). Now, I will try to stick to this rule: I add docstring of mathematical notation/equations to explain the computations in a function. This should not change for most of the time, so I will expect the out-of-sync problem won't occur often. I removed those comments such as `get ** ` and some obvious array shape, I prefer to add comments in the code to explain why and how at places that I feel not straightforward to understand. How do you think?

model/common/src/icon4py/model/common/grid/grid_manager.py
OngChia: copy and paste error. Removed.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Thanks! I did not manage to think of that further vectorization.

model/common/src/icon4py/model/common/grid/grid_manager.py
OngChia: Renamed to `e2c_c2v` . I prefer to keep the comment as a reminder of why we have the "complicated" manipulation in this function to derive `e2c2v`
```
e2c_c2v = _patch_with_dummy_lastline(c2v, array_ns=array_ns)[e2c, :]
# `flat` includes duplicated e2v vertices (v1, v3), shape (n_edges, 6).
flat = e2c_c2v.reshape(e2c_c2v.shape[0], -1)
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: I removed some comments. But I prefer to especially keep, for example, `# we need has_match **` (this one is from Claude) because it explains why I need `has_match`, it is not obvious, it is there because of how numpy computes argmax, which we can't control and may change. I may forget the reason next time I read this line without the comment.

OngChia: My understanding is that, by default, numpy uses int64, but our offset table indices use int32.

model/common/src/icon4py/model/common/grid/grid_manager.py
nfarabullini: sure!

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
nfarabullini: it's not a big deal to keep it, but I'm wondering whether even is numpy defaults to int64, that causes problems later

OngChia:
cscs-ci run default

OngChia:
cscs-ci run distributed

OngChia:
cscs-ci run distributed

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run default

OngChia:
cscs-ci run distributed

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: The cell_size is also used below. So, I think keeping it?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Similarly, cell_size is used twice here.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Thanks! I will inline it!

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -910,34 +932,56 @@ def compute_cells_aw_verts(
...
+    vertex_range = array_ns.arange(horizontal_start, num_verts)
+
+    # Precompute valid v2c mask: skip MISSING and consecutive duplicates that may come from the pentagon points or the domain boundary
+    valid_v2c = array_ns.zeros(v2c.shape, dtype=bool)
+    for jc in range(num_cells_per_vert):
+        valid_v2c[:, jc] = (
+            v2c[:, jc] != MISSING
+            if jc == 0
+            else (v2c[:, jc] != MISSING) & (v2c[:, jc] != v2c[:, jc - 1])
+        )
```
nfarabullini, line 948: do you want to move this down to where the other jc for loop is since `valid_v2c` is only used there and do a basic refactoring there?

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -13,7 +13,6 @@
 
 import gt4py.next as gtx
 import numpy as np
```
nfarabullini, line 15:
```suggestion
```
This should go and if any other numpy function is used, it should be replaced with `array_ns`

model/common/src/icon4py/model/common/grid/grid_manager.py
```diff
@@ -604,23 +604,28 @@ def _construct_diamond_vertices(
 
     For example for this diamond: e0 -> (v0, v1, v2, v3)
     Ordering is the same as ICON uses.
-
+
```
msimberg: Nit:
```suggestion

```

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -436,27 +436,34 @@ def index_offset(f):
...
-    for i in range(num_elements):
-        valid_neighbors = np.where(rbf_offset_np[i, :] >= 0)[0]
-        rbfmat_np = np.squeeze(z_rbfmat_np[np.ix_([i], valid_neighbors, valid_neighbors)])
-        z_diag_np = sla.cho_factor(rbfmat_np)
+    # Batch solve by grouping elements with the same number of valid neighbors.
+    # In ICON grids, valid entries in connectivity tables are contiguous from the start.
+    n_valid = (rbf_offset >= 0).sum(axis=1)
+    for nv in array_ns.unique(n_valid):
+        if nv == 0:
+            continue
```
nfarabullini:
```suggestion
    for nv in array_ns.unique(n_valid[n_valid != 0]):
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -190,21 +152,69 @@ def compute_diffusion_intcoef_and_vertoffset(
...
+        0
+    ]
+    n_valid_cells = len(valid_cell_mask)
+
+    # Compute max vertical z-range for all cells, this is used later to construct a one-to-one mapping when searching for neighboring cell's level index that lies within the vertical range of the current cell
+    max_vertical_zrange = array_ns.maximum(z_mc.max() - z_mc.min(), 0)  # (n_valid_cells)
+
+    if n_valid_cells > 0:
+        masked_z_mc = z_mc[valid_cell_mask, :]  # (n_valid_cells, nlev)
+        masked_kstart = k_start[valid_cell_mask]  # (n_valid_cells)
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -190,21 +152,69 @@ def compute_diffusion_intcoef_and_vertoffset(
...
+        masked_kend = k_end[valid_cell_mask]  # (n_valid_cells)
+        # Level mask: True where level is in [kstart, kend) for each valid cell
+        default_level_idx = array_ns.arange(nlev)[array_ns.newaxis, :]
+        krange_mask = (default_level_idx >= masked_kstart[:, array_ns.newaxis]) & (
```
nfarabullini:
```suggestion
        krange_mask = (default_level_idx >= k_start[valid_cell_mask][:, array_ns.newaxis]) & (
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -133,25 +96,27 @@ def compute_diffusion_mask_and_coef(
...
-                    * array_ns.sqrt(
-                        array_ns.maximum(0.0, maxhgtd_avg[jc, k_range] - thhgtd_zdiffu)
-                    ),
-                ),
-            )
-            zd_diffcoef[jc, k_range] = array_ns.minimum(0.002, zd_diffcoef_var)
+    cell_sequence = array_ns.arange(n_cells)
+    valid_cell_mask = cell_index_mask & (cell_sequence >= cell_nudging)
+    masked_kend = k_end[valid_cell_mask]
+    masked_kstart = k_start[valid_cell_mask]
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -133,25 +96,27 @@ def compute_diffusion_mask_and_coef(
...
+    default_level_idx = array_ns.arange(nlev)[array_ns.newaxis, :]
+    krange_mask = (default_level_idx >= masked_kstart[:, array_ns.newaxis]) & (
+        default_level_idx < masked_kend[:, array_ns.newaxis]
+    )  # (n_valid_cells, nlev)
```
nfarabullini:
```suggestion
    krange_mask = (default_level_idx >= k_start[valid_cell_mask][:, array_ns.newaxis]) & (
        default_level_idx < k_end[valid_cell_mask][:, array_ns.newaxis]
    )  # (n_valid_cells, nlev)
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -190,21 +152,69 @@ def compute_diffusion_intcoef_and_vertoffset(
...
+    ]
+    n_valid_cells = len(valid_cell_mask)
+
+    # Compute max vertical z-range for all cells, this is used later to construct a one-to-one mapping when searching for neighboring cell's level index that lies within the vertical range of the current cell
+    max_vertical_zrange = array_ns.maximum(z_mc.max() - z_mc.min(), 0)  # (n_valid_cells)
+
+    if n_valid_cells > 0:
+        masked_z_mc = z_mc[valid_cell_mask, :]  # (n_valid_cells, nlev)
+        masked_kstart = k_start[valid_cell_mask]  # (n_valid_cells)
+        masked_kend = k_end[valid_cell_mask]  # (n_valid_cells)
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -190,21 +152,69 @@ def compute_diffusion_intcoef_and_vertoffset(
...
+        # Level mask: True where level is in [kstart, kend) for each valid cell
+        default_level_idx = array_ns.arange(nlev)[array_ns.newaxis, :]
+        krange_mask = (default_level_idx >= masked_kstart[:, array_ns.newaxis]) & (
+            default_level_idx < masked_kend[:, array_ns.newaxis]
+        )  # (n_valid_cells, nlev)
```
nfarabullini:
```suggestion
            default_level_idx < k_end[valid_cell_mask][:, array_ns.newaxis]
        )  # (n_valid_cells, nlev)
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1228,10 +1299,7 @@ def compute_lsq_coeffs(
     exchange: Callable[[data_alloc.NDArray], None] = decomposition.single_node_exchange,
     array_ns: ModuleType = np,
 ) -> data_alloc.NDArray:
-    lsq_weights_c = array_ns.zeros((min_rlcell_int, lsq_dim_stencil))
-    lsq_pseudoinv = array_ns.zeros((min_rlcell_int, lsq_dim_unk, lsq_dim_c))
-    z_lsq_mat_c = array_ns.zeros((min_rlcell_int, lsq_dim_c, lsq_dim_c))
-    z_dist_g = array_ns.zeros((min_rlcell_int, lsq_dim_c, 2))
+    z_dist_g = array_ns.zeros((cell_owner_mask.shape[0], lsq_dim_c, 2), dtype=ta.wpfloat)
```
nfarabullini: isn't the default for `array_ns.zeros`  `float` already?

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -242,7 +241,7 @@ def _compute_distance_vector_matrix(
         case base_grid.GeometryType.TORUS:
             # For pairs of points p1 and p2 compute:
             # norm(p1 - p2) noqa: ERA001
-            diff = np.abs(v1 - v2)
+            diff = array_ns.abs(v1 - v2)
```
msimberg, line 244: 🙈 good catch!

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -436,27 +436,34 @@ def index_offset(f):
...
     ]
-    rbf_offset_np = data_alloc.as_numpy(rbf_offset)
-    z_rbfmat_np = data_alloc.as_numpy(z_rbfmat)
-    rhs_np = [data_alloc.as_numpy(x) for x in rhs]
-    for i in range(num_elements):
-        valid_neighbors = np.where(rbf_offset_np[i, :] >= 0)[0]
-        rbfmat_np = np.squeeze(z_rbfmat_np[np.ix_([i], valid_neighbors, valid_neighbors)])
-        z_diag_np = sla.cho_factor(rbfmat_np)
+    # Batch solve by grouping elements with the same number of valid neighbors.
+    # In ICON grids, valid entries in connectivity tables are contiguous from the start.
```
msimberg: I'm 90% sure this is true, but always have a bit of doubt if these things are true only for the grids we test so far or universally true. I'm relatively sure the solve will fail sooner or later if this isn't true...

msimberg: Probably worth explicitly adding a comment that this assumes invalid neighbors are marked with negative indices. Repeated invalid neighbors is not supported (it wasn't supported before either, so not a new constraint) and I think an explicit comment doesn't hurt. The `n_valid` line sort of implies this, but not fully.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: I don't know if it looks cleaner to remove the zero upfront:

```suggestion
    for nv in (u := array_ns.unique(n_valid))[u != 0]:
```
? Completely optional as I'm not sure myself if it's better.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -436,27 +436,34 @@ def index_offset(f):
...
-    rbf_vec_coeff = tuple([array_ns.asarray(x) for x in rbf_vec_coeff_np])
+            rhs_batch = rhs[j][array_ns.ix_(group_idx, valid_cols)]
+            # array_ns.linalg.solve supports batched inputs: mat_batch (B, nv, nv),
+            # rhs_batch (B, nv, 1). The solution of mat_batch x = rhs_batch is sol.
+            # rhs_batch is expanded to 3D so both numpy and cupy treat it as a
+            # batched column vector (core dims (nv,1)) rather than a matrix
+            # (core dims (B, nv)), which would mismatch m=nv from the LHS.
+            # This problem is well explained in https://github.com/numpy/numpy/issues/26598
+            # The RBF matrix is symmetric but NOT necessarily positive definite
+            # (z_nxprod = n_i·n_j can be negative), so Cholesky would be wrong.
```
msimberg: Wait what is this about? Is this claude's comment? I'm just wondering if this is an indication that it's including elements that should not be in the RBF matrix... because when one does that cholesky does indeed fail.

If the above question is resolved and we stick with a general purpose solve here, I'd definitely add a comment explaining why this doesn't use cholesky like ICON (explaining that it's purely performance, not "better" solve or something like that, and the batching makes up for the less efficient individual solves).

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -133,25 +96,27 @@ def compute_diffusion_mask_and_coef(
...
+
+    zd_diffcoef_var = array_ns.maximum(
+        0.0,
+        array_ns.maximum(
+            array_ns.sqrt(array_ns.maximum(0.0, maxslp_avg[valid_cell_mask, :] - thslp_zdiffu))
+            / 250.0,
+            2.0e-4
+            * array_ns.sqrt(array_ns.maximum(0.0, maxhgtd_avg[valid_cell_mask, :] - thhgtd_zdiffu)),
+        ),
+    )
```
msimberg, line 106: This looks suspiciously like one or more of the `maximum(0.0, ...)`s could be removed, but I can't say if that's actually possible.

msimberg (changes requested):
A few more small comments. As before, I mostly trust the tests to make sure this is correct. For performance I noticed at least one place which seems a bit overengineered (too many `maximum`s, possibly; there might be others like this). I'll do a local review as well to see if there's anything more that could be simplified.

But even then, I think this is already mergeable and we shouldn't delay it much. There's always room for improvement and we can come back to these later.

The comment on the RBF interpolation about Cholesky being the wrong choice is my only blocking comment, that smells very suspicious.

msimberg:
And I think it'd be interesting (as a follow-up) to see if any of these would benefit from jax/numba, or if they're too complex for that. I opened https://github.com/C2SM/icon4py/issues/1228 to keep track of that.

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
```diff
@@ -119,9 +71,20 @@ def compute_diffusion_mask_and_coef(
     nlev: int,
     array_ns: ModuleType = np,
 ) -> tuple[data_alloc.NDArray, data_alloc.NDArray]:
+    """
```
msimberg, line 74: Outside the diff, but there's a hardcoded `np.argmax` up on line 54 of `_compute_k_start_end` that likely should be fixed.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1164,48 +1207,76 @@ def compute_lsq_pseudoinv(
...
+    valid_cell_mask = (
+        cell_owner_mask & (cell_sequence >= start_idx) & (cell_sequence < min_rlcell_int)
+    )
+    lsq_pseudoinv = array_ns.zeros((cell_size, lsq_dim_c, lsq_dim_c), dtype=ta.wpfloat)
+    u_matrix, s_matrix, v_t_matrix = array_ns.linalg.svd(z_lsq_mat_c[valid_cell_mask, :, :])
+    v_t_over_s = (
+        v_t_matrix[:, :lsq_dim_unk, :lsq_dim_unk] / s_matrix[:, :lsq_dim_unk, array_ns.newaxis]
+    )
+    pinv = array_ns.matmul(
+        array_ns.transpose(v_t_over_s, (0, 2, 1))[:, :lsq_dim_unk, :lsq_dim_unk],
```
msimberg: The slicing on `v_t_over_s` looks redundant since it was already sliced earlier. Since only dimension 1 and 2 are swapped (which are both `lsq_dim_unk`) I think you can just do the transpose without slicing?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1164,48 +1207,76 @@ def compute_lsq_pseudoinv(
...
+    z_lsq_mat_c has dimensions (c, k), lsq_pseudoinv has (k, c), and lsq_weights_c has (c)
+    singular value decomposition of the matrix z_lsq_mat_c gives u_matrix (c, k), s_matrix (k), v_t_matrix (k, k)
+    lsq_pseudoinv = v_t_matrix^T u_matrix^T / s * lsq_weights_c
+    """
+    cell_size = cell_owner_mask.shape[0]
+    cell_sequence = array_ns.arange(cell_size)
+    valid_cell_mask = (
+        cell_owner_mask & (cell_sequence >= start_idx) & (cell_sequence < min_rlcell_int)
+    )
+    lsq_pseudoinv = array_ns.zeros((cell_size, lsq_dim_c, lsq_dim_c), dtype=ta.wpfloat)
```
msimberg: Unsure if this is a typo or intentional "scratchpad":

```suggestion
    lsq_pseudoinv = array_ns.zeros((cell_size, lsq_dim_unk, lsq_dim_c), dtype=ta.wpfloat)
```
?

msimberg (commented):
A few more comments from a friendly neighborhood LLM.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: I did some local testing and as long as we do RBF on the interior only then we don't get invalid indices other than at the end. For vertices we still get invalid indices on the pentagon points, but they seem to currently be at the end of the neighbor list (again, not 100% sure if this is convention or guaranteed). If we compute all the way into the halos we do get invalid indices in the middle of the array. We shouldn't do this, because we can fill in the halos with halo exchanges. Also, lateral boundaries should still be fine because we start the computation at LATERAL_BOUNDARY_LEVEL_2, so I think we avoid the problem there as well (and this should mirror what ICON fortran does).

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Yes, that is from claude's comment. I blindly believed what claude said because I saw that `z_nxprod = n_i·n_j can  indeed be negative..

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: But it is used as the default value for `array_ns` when `array_ns` is passed in as an argument, for example in `_dot_product`. Should I remove the default as well?

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Thanks for the nice suggestion!

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Actually I am not sure, but the test does not complain without `ta.wpfloat`. I will remove it.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: I found that if I move `valid_v2c` mask to the `jc` loop below, the mask will be unnecessarily constructed 6 times per neighboring cell instead of one per neighboring cell because of the outer `je` loop. So, I would prefer to keep it outside.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Thanks for the explanation! I have added assumptions at the beginning of the batched solving of rbf coefficients
```
    # ASSUMPTIONS FOR MAKING THE FOLLOWING BATCH SOLVE POSSIBLE:
    #   (1) In ICON grids, valid entries in connectivity tables are contiguous from the start.
    #       In other words, those invalid neighbors must be located at the end of the neighbor list.
    #       Therefore, we cannot compute rbf all the way into halo cells or the first boundary layer
    #       because the invalid neighbors may appear in the middle the the neighbor list.
    #   (2) Invalid indices must be a negative.
```

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Good suggestion! Thanks!

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
OngChia: Indeed it can be removed.

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
OngChia: Thanks!

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: I will keep it for now to be consistent with the rest of icon4py where `np` is used as the default.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: You are right. v_t_over_s already has dimension `(:, lsq_dim_unk, lsq_dim_unk)`. There is no need to slice further.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: You are right. `lsq_pseudoinv` has dimensions `(cell_size, lsq_dim_unk, lsq_dim_c)`. Although `lsq_dim_c` is always larger than `lsq_dim_unk` because the number of cells used for least squares interpolation should always be larger than the number of unknowns, there is no need to "pad" the matrix.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
OngChia: Updated comment:
```
            # array_ns.linalg.solve supports batched inputs: mat_batch (B, nv, nv),
            # rhs_batch (B, nv, 1). The solution of mat_batch x = rhs_batch is sol.
            # rhs_batch is expanded to 3D so both numpy and cupy treat it as a
            # batched column vector (core dims (nv,1)) rather than a matrix
            # (core dims (B, nv)), which would mismatch m=nv from the LHS.
            # This problem is well explained in https://github.com/numpy/numpy/issues/26598
            # The RBF matrix is symmetric and positive definite. However,
            # the Cholesky method is not chosen, as in ICON, simply because scipy
            # does not support batched solving of the linear equation,
            # necessitating a Python loop and resulting in poor performance.
```

OngChia:
cscs-ci run distributed

OngChia:
cscs-ci run default

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
msimberg: This is resolved but the diff doesn't show as outdated. Did you forget to push?

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: #1178 will/should clean this up, so in my opinion it's ok if you leave it as it was.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: We discussed this many times in person already, but just to resolve this I'll answer here as well.

The reason they were computed on CPU (at the time) was indeed to use the cholesky decomposition (like ICON) and the cupy equivalent of scipy did not have a cholesky decomposition back then (also in the comment in the code).

msimberg (approved):
Tentative approval, with a request to check if the nested `maximum` change was actually applied (or if it doesn't work, comment that it doesn't work). Edit: this was just me not seeing the latest changes, all is good.

Santis is still coming back to life after last week, so we'll have to retry a few times with CI.

model/common/src/icon4py/model/common/metrics/compute_diffusion_metrics.py
msimberg: GitHub's UI is not my friend today. I see this was fixed now in https://github.com/C2SM/icon4py/pull/1198/changes/c9ef4a015a6857f9d8e6a90fbaa5556808383652, I was seeing the old version of this.

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

PR #1199, opened 2026-04-16
Amd profiling

dganellari:


model/common/src/icon4py/model/common/model_options.py
```diff
@@ -58,6 +60,13 @@ def get_dace_options(
...
         optimization_args["gpu_memory_pool"] = False
         optimization_args["make_persistent"] = True
+        # AMD MI300A: ~7% speedup by fusing DaCe tasklet ops within map scopes.
+        optimization_args.setdefault("fuse_tasklets", True)
+        # AMD MI300A: (256,1,1) for all maps — 14.3% faster than default (32,8,1).
+        # All threads on Cell dimension maximizes coalescing on MI300A.
+        optimization_args.setdefault("gpu_block_size", (256, 1, 1))
+        optimization_args.setdefault("gpu_block_size_1d", (256, 1, 1))
+        optimization_args.setdefault("gpu_block_size_2d", (256, 1, 1))
     if program_name == "graupel_run":
```
iomaganaris, line 64: Can we just quantify what change these 2 options do in the `GT4Py timers` separately?

dganellari: gpu_block_size_2d=(256,1,1) gives -21% (only _2d matters)

_1d and fuse_tasklets are neutral on this gt4py version. GH200 gap down to 1.09x.

Details in PROFILING_RESULTS.md.

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -58,6 +58,9 @@ def get_dace_options(
     if backend_descriptor["device"] == model_backends.DeviceType.ROCM:
         optimization_args["gpu_memory_pool"] = False
         optimization_args["make_persistent"] = True
+    # AMD MI300A: (256,1,1) for 2D maps gives ~21% speedup on the solver.
```
iomaganaris: Out of curiosity, did you do a sweep for different thread block sizes? i.e. (128,1), (128,2) or (256, 2), etc.
I think that similar to GH200 having in the Y dim something > 1 might be beneficial but we have to figure out the XCD assignment first

dganellari: Just did the rest of the sweep on the solver (GT4Py Timer, 1000 runs each, MI300A):

(32, 8, 1): 753 μs (+25% vs best)
(128, 1, 1): 719 μs (+19%)
(128, 2, 1): 642 μs (+6.3%)
(256, 1, 1): 604 μs (best)
(256, 2, 1): 598-615 μs (within ~3% noise)
X-axis size dominates. Y > 1 only helps when X is small; at X=256 it's within noise.

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -58,6 +58,9 @@ def get_dace_options(
     if backend_descriptor["device"] == model_backends.DeviceType.ROCM:
         optimization_args["gpu_memory_pool"] = False
         optimization_args["make_persistent"] = True
+    # AMD MI300A: (256,1,1) for 2D maps gives ~20% speedup on the solver.
+    # All threads on Cell dimension maximizes coalescing on MI300A.
+    optimization_args.setdefault("gpu_block_size_2d", (256, 1, 1))
```
edopao:
```suggestion
        # AMD MI300A: (256,1,1) for 2D maps gives ~20% speedup on the solver.
        # All threads on Cell dimension maximizes coalescing on MI300A.
        optimization_args.setdefault("gpu_block_size_2d", (256, 1, 1))
```

dganellari: @iomaganaris Should I set this only for MI300A or also for GH200 for which the improvement only amounts to ~5%?

iomaganaris: I think for both GH200 and MI300A we have to run the whole dycore to set these values globally. I'm a bit more relaxed with setting this only for MI300A for the moment because I think it will improve all stencils of icon4py without checking 🙈

dganellari: Agreed. Also because even though it is 5% faster, the cache hit rate for going to 256 for GH200 drops.

iomaganaris: Also we have to check for different grid sizes

amd_scripts/set_waves_per_eu.py
```diff
@@ -0,0 +1,100 @@
...
+import shutil
+import sys
+from pathlib import Path
+
+
+def patch_kernels(cpp_file: Path, min_waves: int, max_waves: int, dry_run: bool = False) -> int:
+    """Add amdgpu_waves_per_eu attribute to all __global__ kernel declarations."""
+    content = cpp_file.read_text()
+
+    attr = f"__attribute__((amdgpu_waves_per_eu({min_waves}, {max_waves})))"
```
iomaganaris, line 32: These changes should go into DaCe for sure if necessary. Did you measure any difference in performance with that?

model/common/src/icon4py/model/common/model_options.py
dganellari: done and in the profiling results.

amd_scripts/set_waves_per_eu.py
dganellari: Yes but in this case it was slower since the occupancy is already good on MI300. For more details:

https://github.com/dganellari/icon4py/blob/amd_profiling/amd_scripts/docs/ATTEMPTED_OPTIMIZATIONS.md#occupancy-experiment

PR #1200, opened 2026-04-16
Improve script infrastructure 

egparedes:
This PR refactors the development scripts folder to support different kind of standalone scripts while still offering an (optional) single-entry point to all of them. The design and instructions are in the README file.

Discussion points:
- should we move the CI scripts to this  folder or not? I've tentatively moved (and adapted) python script there (`validation_grid_utils.py`) to this folder as an experiment, but haven't really changed the CI action to use it yet. What's your view on this   @msimberg ?
- @jcanton, are you ok with the refactoring of the serialization scripts ?


msimberg:
>     * should we move the CI scripts to this  folder or not? I've tentatively moved (and adapted) python script there (`validation_grid_utils.py`) to this folder as an experiment, but haven't really changed the CI action to use it yet. What's your view on this   @msimberg ?

I have very little opinion on the implementation details of this as long as it works. Yeah, I don't have any objection to moving the scripts from `ci/scripts`! It feels intuitively extremely overengineered for calling a bash script, but I understand the setup isn't there purely for bash scripts :P and still, as long as it works, I don't care too much.

So that would practically mean instead of `ci/scripts/ci-mpi-wrapper.sh` I get something like `scripts/run ci-mpi-wrapper`? Note that `start-cuda-mps.sh` is meant to be sourced, not run, so I'm not sure you can move that? Or is that also supported? `eval $(scripts/run start-cuda-mps)`?

egparedes:
> I have very little opinion on the implementation details of this as long as it works. Yeah, I don't have any objection to moving the scripts from `ci/scripts`! It feels intuitively extremely overengineered for calling a bash script, but I understand the setup isn't there purely for bash scripts :P and still, as long as it works, I don't care too much.
> 
> So that would practically mean instead of `ci/scripts/ci-mpi-wrapper.sh` I get something like `scripts/run ci-mpi-wrapper`? Note that `start-cuda-mps.sh` is meant to be sourced, not run, so I'm not sure you can move that? Or is that also supported? `eval $(scripts/run start-cuda-mps)`?

Just a clarification: the main point of this PR is to define and document the structure of the scripts folder so all scripts follow the same (loose) conventions and development & testing patterns, specially python scripts. Additionally, it adds a single entry point to quickly see all the functionality provided by the scripts already existing in the repo, but this entry point is just an extra feature, and all the scripts should still be fully standalone. I made sure that python scripts written with the recommended pattern (use `typer` and name the app object as `cli`) both work as subcommands of the `./scripts/run` entry point and standalone.

scripts/python/run_serialization.py
```diff
@@ -10,72 +12,122 @@
...
+    runscript_dir: pathlib.Path
+    experiments_dir: pathlib.Path
+    output_root: pathlib.Path
+    max_threads: int
+
+    @classmethod
+    def defaults(cls) -> SerializationSettings:
+        # We hardcode the settings here for simplicity, but they could be
+        # extended to be read from a config file or command-line arguments
+        # if needed in the future.
```
jcanton, line 61:
```suggestion
    def defaults(cls) -> SerializationSettings:
        # ======================================
        # START DEFAULT USER CONFIGURATION
        # ======================================
        # We hardcode the settings here for simplicity, but they could be
        # extended to be read from a config file or command-line arguments
        # if needed in the future.
```

scripts/python/run_serialization.py
```diff
@@ -10,72 +12,122 @@
...
+        RUNSCRIPTS_DIR = BUILD_DIR / "run"
+        EXPERIMENTS_DIR = BUILD_DIR / "experiments"
+
+        # Output location for copied ser_data and tarballs
+        OUTPUT_ROOT = EXPERIMENTS_DIR / definitions.SERIALIZED_DATA_DIR
+
+        # Maximum concurrent threads for running experiments
+        MAX_THREADS: int = 5
+
+        return cls(
```
jcanton, line 103:
```suggestion
        MAX_THREADS: int = 5

        # ======================================
        # END DEFAULT USER CONFIGURATION
        # ======================================
        return cls(
```

jcanton:
just two minor comments to make the "config" block more clearly visible? (if you have better ideas go ahead)
otherwise it looks ok to me!

egparedes:
Closed in favor of #1232

PR #1204, opened 2026-04-21
Add AGENTS.md

msimberg:


muellch:
1) This is also good human onboarding documentation. In general the better the human documentation the better the AI agents understand what is going on. Every time one let's a new AI agent work on this project it is like onboarding a human.
2) Should we add something about higher level concepts like `field_operators` being pure functions, etc. etc. Or should this live in gt4py repo and then we link to that.
3) Should we somewhere explain to the AI that this is a Python implementation of Fortran ICON components?

None of these need to be addressed in this PR.

msimberg:
> 1. This is also good human onboarding documentation. In general the better the human documentation the better the AI agents understand what is going on. Every time one let's a new AI agent work on this project it is like onboarding a human.

Agree. Once I/we have removed the nix shell instructions I was planning to ask copilot to simply run tests and linters. If the instructions aren't good enough it will go in circles and try to rediscover how to do that.

> 2. Should we add something about higher level concepts like `field_operators` being pure functions, etc. etc. Or should this live in gt4py repo and then we link to that.

Not at all opposed to that. If it's just a single sentence maybe. But if it's just a single sentence I'm not sure an agent learns much from it. If it's more than a sentence it probably deserves its own document that can be referenced in AGENTS.md?

> 3. Should we somewhere explain to the AI that this is a Python implementation of Fortran ICON components?

This is worth adding. Maybe with a link to upstream git repo/model website.

msimberg:
I added a few pieces of high level information from @havogt's https://github.com/C2SM/icon4py/blob/scientific_validation_distributed_driver_hannes/CLAUDE.md, that is likely to help agents navigate the repo/concepts.

havogt (approved):

PR #1212, opened 2026-04-22
Fix cupy/numpy usage in metis decomposition

msimberg:
A big slowdown in metis decomposition with GPU backends comes from metis restructuring the c2e2c adjacency matrix that we pass into a format for internal use. If I understand correctly, pymetis happens to work with cupy arrays because of this indexing: https://github.com/inducer/pymetis/blob/1db6cd1d189ddda75b80cb24125c0e9de88d79d8/pymetis/__init__.py#L358. This accesses a single element from the cupy array which is brought back to the host as a scalar.

This is of course extremely slow with cupy arrays. This PR makes sure we only pass numpy arrays to pymetis.

Additionally, this changes the call to pymetis to use the lower level interface where pymetis doesn't have to transform the adjacency matrix to a format used by metis. We can do this a lot faster because, at the moment, we only support global and torus grids, which always have full connectivity (three neighbors for every cell). With this assumption we can do the setup a lot faster than what pymetis can do with weaker assumptions. When we support limited area grids we'll have to revisit this setup, but even if we use the adjacency matrix interface the major speedup comes from the cupy/numpy fix.

I did a small test using the standalone benchmark script in https://gist.github.com/msimberg/84ad8195688d3edca08c8ef839e3a0f8. It compares:
- passing a cupy array directly to pymetis
- converting to numpy
- converting to a list (this is what pymetis really wants, but it works with arrays as well)
- using the low level API

On one GH200 module it outputs the following for a random 1000000 vertex graph:

| array_lib  |  n_cells | nparts | method                   | total (ms) | prep (ms) |
|------------|----------|--------|--------------------------|------------|-----------|
| cupy       |  1000000 |      4 | adjacency=xp (no conv)   |  117225.73 |         - |
| cupy       |  1000000 |      4 | adjacency=ndarray        |    4538.66 |    0.0842 |
| cupy       |  1000000 |      4 | adjacency=list           |    2687.75 |  374.5756 |
| cupy       |  1000000 |      4 | xadj/adjncy (CSR)        |    2380.84 |    0.4656 |

The graph is random (not a real icon grid), but is likely representative enough to conclude: the implicit cupy array access is expensive.

Just for completeness, using numpy for everything the timings are:

| array_lib  |  n_cells | nparts | method                   | total (ms) | prep (ms) |
|------------|----------|--------|--------------------------|------------|-----------|
| numpy      |  1000000 |      4 | adjacency=xp (no conv)   |    4418.93 |         - |
| numpy      |  1000000 |      4 | adjacency=ndarray        |    4551.28 |    0.0001 |
| numpy      |  1000000 |      4 | adjacency=list           |    2742.63 |  353.4515 |
| numpy      |  1000000 |      4 | xadj/adjncy (CSR)        |    2375.42 |    0.3910 |

So the slowdown is not just from the `no conv` path running first or similar.

### DIFF
--- a/model/common/src/icon4py/model/common/decomposition/decomposer.py
+++ b/model/common/src/icon4py/model/common/decomposition/decomposer.py
@@ -67,7 +67,7 @@ class MetisDecomposer(Decomposer):
         # which avoids Python-side iteration and copies in pymetis._prepare_graph.
         #
         # xadj is an array [0, 3, 6, 9, ...] where xadj[i] is the start index of
-        # vertex i's neighbors, xadj[i+1] is the end index. Each cell always has
+        # cell i's neighbors, xadj[i+1] is the end index. Each cell always has
         # 3 neighbor cells (in global/torus grids).
         _, partition_index = pymetis.part_graph(
             nparts=num_partitions,
--- a/model/common/src/icon4py/model/common/decomposition/decomposer.py
+++ b/model/common/src/icon4py/model/common/decomposition/decomposer.py
@@ -9,6 +9,7 @@
 from typing import Protocol, runtime_checkable
 
 import gt4py.next as gtx
+import numpy as np
 
 from icon4py.model.common.utils import data_allocation as data_alloc
 
@@ -31,11 +32,6 @@ class Decomposer(Protocol):
 class MetisDecomposer(Decomposer):
     """
     A simple decomposer using METIS for partitioning a grid topology.
-
-    We use the simple pythonic interface to pymetis: just passing the adjacency matrix, which for ICON is
-    the full grid C2E2C neigbhor table.
-    if more control is needed (for example by using weights we need to switch to the C like interface)
-    https://documen.tician.de/pymetis/functionality.html
     """
 
     def __call__(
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

model/common/src/icon4py/model/common/decomposition/decomposer.py
```diff
@@ -60,9 +56,25 @@ def __call__(
...
+
         # The partitioning is done on all ranks, and this assumes that the
         # partitioning is deterministic.
-        _, partition_index = pymetis.part_graph(nparts=num_partitions, adjacency=adjacency_matrix)
+        #
+        # Passes the adjacency matrix in CSR format (xadj/adjncy) to pymetis,
+        # which avoids Python-side iteration and copies in pymetis._prepare_graph.
+        #
+        # xadj is an array [0, 3, 6, 9, ...] where xadj[i] is the start index of
+        # vertex i's neighbors, xadj[i+1] is the end index. Each cell always has
```
OngChia: Is "vertex" equivalent to "cell" in icon grid? Or maybe I misunderstood something here.

OngChia: if cell 0 has neighbors 1, 2, and 3; cell 1 has neighbors 4, 5, and 0. xadj will be [1, 3, 4, 0, ...]?

msimberg: Excellent question, I didn't think about that when writing this. Yes, "vertex" is equivalent to "cell". I'll change this to icon grid terminology, makes a lot more sense like that.

The xadj array is a table of indices into adjncy, so the neighbors indices of cell i are in at positions [xadj[i], xadj[i+1]) in adjncy.

In your example, adjncy is [1, 2, 3, 4, 5, 0, ...], i.e. c2e2c flattened into a 1d array. xadj is always the same [0, 3, 6, ...] as long as all cells have three neighbors, because xadj[0] == 0 is where the neighbors of cell 0 start in adjncy. xadj[1] is where the neighbors of cell 0 end (exclusive) in adjncy, and where neighbors of cell 1 start.

Makes sense?

msimberg: I changed "vertex" to "cell" in the comment. Let me know if you think the comment can be clarified further.

OngChia: Ahh, now I understand! Thank you for the detailed explanation.

OngChia: I think maybe you can put that example in the comment. At least, I felt hard to understand.

OngChia (approved):
Thanks for the interesting experiment. So, my understanding is that you partition a 1D grid in a ring shape in your experiment with cupy array, list, or numpy for the adjacency matrix. I tried your script. Regardless of the number of cells, there is a huge speed up when switching from cupy to numpy. Interesting. But at the same time, wondering why you did not see the same speedup with the icon grid. Overall, I agree that it is a good start.

msimberg:
> But at the same time, wondering why you did not see the same speedup with the icon grid. Overall, I agree that it is a good start.

I was probably unclear. I just didn't try it with a real icon grid. A friend of mine wrote the standalone benchmark script.

model/common/src/icon4py/model/common/decomposition/decomposer.py
msimberg: Do you think https://github.com/C2SM/icon4py/pull/1212/commits/1913300d7a29483e7289b5131c1f9a03176ce6f6 helps?

OngChia: Yes, that is clearer!

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1218, opened 2026-04-24
Add persistent gt4py cache for CSCS CI

### LINKED ISSUE #1133: Set up a persistent cache across jobs for CI
Currently there's a persistent gt4py cache within a job in CI which helps a bit (different workers and/or ranks can share the cache). However, every job still has to regenerate the cache from scratch which is expensive. Try to put a semi-persistent cache on scratch (clean every now and then to 1. avoid it growing too big and 2. to every now and then trigger recompilation).

msimberg:
Fixes #1133.

Sets up a persistent cache for cscs ci pipelines. There's a weekly cache shared across all jobs. This means that each week a new cache directory is populated based on the year and week number. The idea of this is that we sometimes trigger compilation for a new cache from scratch, but not for every job. The cache is shared between branches and the same uv.lock to avoid having too many files on scratch from every branch writing a separate cache such as in https://github.com/C2SM/icon4py/pull/1220. The caches are separated by uv.lock hash to avoid sharing cache when gt4py or another important dependency changes. The caches are separated by job because .venvs are different for each nox session. This avoids issues with hardcoded paths in compile_commands.json.


msimberg:
The github actions cache for gt4py seems to work nicely. I had `test-model (3.11, gtfn_cpu, dycore)` do a first run without a cache in 27 minutes. The second run with a cache ran in 10 minutes.

I've yet to try the cscs ci cache.

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

ci/default.yml
```diff
@@ -23,8 +23,9 @@ include:
         GRID: [simple, icon_regional]
 # test_model_stencils_x86_64:
 #   extends: [.test_model_stencils, .test_template_x86_64]
-test_model_stencils_aarch64:
-  extends: [.test_model_stencils, .test_template_aarch64]
+# TODO(msimberg): enable
+# test_model_stencils_aarch64:
+#   extends: [.test_model_stencils, .test_template_aarch64]
```
msimberg: To do.

ci/default.yml
```diff
@@ -60,8 +61,10 @@ test_tools_datatests_aarch64:
         SLURM_TIMELIMIT: '00:15:00'
   parallel:
     matrix:
-      - COMPONENT: [advection, diffusion, dycore, microphysics, muphys, common, driver, standalone_driver]
-        BACKEND: [embedded, dace_gpu, gtfn_cpu, gtfn_gpu]
+      # - COMPONENT: [advection, diffusion, dycore, microphysics, muphys, common, driver, standalone_driver]
+      - COMPONENT: [dycore, advection]
+        # BACKEND: [embedded, dace_gpu, gtfn_cpu, gtfn_gpu]
+        BACKEND: [dace_gpu, gtfn_cpu]
```
msimberg: To do: revert.

ci/distributed.yml
```diff
@@ -87,14 +90,15 @@ build_distributed:
...
   parallel:
     matrix:
-      - COMPONENT: [atmosphere/diffusion, atmosphere/dycore, common]
-        # TODO(msimberg): Enable dace_gpu when compilation doesn't take as long
-        # or when we can cache across CI jobs.
-        BACKEND: [embedded, gtfn_cpu, dace_cpu, gtfn_gpu]
+      # - COMPONENT: [atmosphere/diffusion, atmosphere/dycore, common]
+      - COMPONENT: [common]
+        # BACKEND: [embedded, gtfn_cpu, dace_cpu, gtfn_gpu]
+        BACKEND: [gtfn_cpu]
```
msimberg: Note, we might still need a long timelimit for dace jobs when the cache is populated. Hopefully the average time will be a lot better than the maximum timelimit though.

msimberg:
This should benefit significantly from https://github.com/GridTools/gt4py/pull/2565. The concern about number of files on scratch will be a much smaller concern (and possibly it allows using per-branch caches? still needs measurement).

ci/scripts/gt4py-cache.sh
```diff
@@ -0,0 +1,20 @@
...
+set -euo pipefail
+
+# First clean up files and directories older than 7 days in the base cache
+# directory. The backend directories can stay, but the date-based directories
+# are removed. There may be concurrent cleanup, ignore failures.
+find "${GT4PY_BUILD_CACHE_BASE_DIR}/icon4py" -mindepth 2 -maxdepth 2 -type d -mtime +7 -exec rm -rf {} + || true
+
+# Then set the cache directory for this run based on the backend and current date.
+DATE=$(date +%Y-%W)
+export GT4PY_BUILD_CACHE_DIR="${GT4PY_BUILD_CACHE_BASE_DIR}/icon4py/${BACKEND}/${DATE}"
```
philip-paul-mueller: Didn't you also need to export the persistent cache env?

philip-paul-mueller (commented):
What I am a bit concerned with is when you work on GT4Py, i.e. if your PR does not touch the ICON4Py stencil code but affects it.
I know the solution is simply to disable it, but there should be somewhere an explanation on how to do it.

ci/scripts/gt4py-cache.sh
msimberg: That is already exported in the general pipeline definitions (we already use it on main to share caches between pytest workers and MPI ranks within one job).

havogt:
> What I am a bit concerned with is when you work on GT4Py, i.e. if your PR does not touch the ICON4Py stencil code but affects it. I know the solution is simply to disable it, but there should be somewhere an explanation on how to do it.

I agree. A simple solution might be to add the gt4py version to the key. PRs not touching gt4py will use the same cache as main, PRs touching gt4py will get their own directory. But need to double-check that the gt4py version is correctly resolved (i.e. with the dev suffix if it's not a release version).

msimberg:
> What I am a bit concerned with is when you work on GT4Py, i.e. if your PR does not touch the ICON4Py stencil code but affects it. I know the solution is simply to disable it, but there should be somewhere an explanation on how to do it.

Yeah, that's a valid concern. How well do you think gt4py itself will invalidate caches when it itself is changed? The "weekly fresh cache" slightly helps protect against this. A per-branch cache might also help a bit more? But yeah, if one really wants a fresh cache then it would (at the moment) mean either disabling the scratch cache for your PR, or we can think about also allowing setting a specific variable to disable the cache (cscs-ci run allows passing extra env vars that will be forwarded to the job). Ideally this would look as close as possible to the github actions caching does (those caches can also be explicitly deleted, manually). File count was my main concern, but it might not even be a real concern.

msimberg:
> > What I am a bit concerned with is when you work on GT4Py, i.e. if your PR does not touch the ICON4Py stencil code but affects it. I know the solution is simply to disable it, but there should be somewhere an explanation on how to do it.
> 
> I agree. A simple solution might be to add the gt4py version to the key. PRs not touching gt4py will use the same cache as main, PRs touching gt4py will get their own directory. But need to double-check that the gt4py version is correctly resolved (i.e. with the dev suffix if it's not a release version).

Commented on the other PR with the same suggestion, but what about including a hash of uv.lock? Can be overly conservative (if we bump other packages unrelated to the gt4py cache), but should be a safe option which catches changes to both gt4py and dace?

philip-paul-mueller:
The question is "what is unrelated" to the GT4Py cache?
In the end it will also be really hard to tell "what will influence the cache".
So I think the safest solution would be to hash `uv.lock`.

edopao:
> setting a specific variable to disable the cache

I like the idea of allowing setting a specific variable to disable the cache, it could be useful for troubleshooting:
`cscs-ci run default GT4PY_BUILD_CACHE_ENABLED=false`

ci/default.yml
edopao: Also add `dace_`cpu, please. And in other places, like stencil tests, run also the dace backends. The dace backends were skipped in several places because too slow. We have instead a nightly dace pipeline that runs with the dace backends. That we can remove after this PR is merged.

msimberg: I'll try it out, but note that the worst case timings for the pipelines will still be as bad as before, when the cache is cleared (currently weekly or with the uv.lock hash changing, perhaps the uv.lock hash could be enough, but then we risk scratch cleanup policies messing up the cache; unsure what the best setup is). I think @havogt's changes to better parallelize gt4py compilation would also be very useful here.

@havogt do you think me might be able to get https://github.com/GridTools/gt4py/pull/2565 and your parallization changes in a 1.1.10 sometime soon or does that need more time?

Edit: this is the parallelization PR I'm talking about: https://github.com/GridTools/gt4py/pull/2587.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

ci/scripts/gt4py-cache.sh
```diff
@@ -0,0 +1,28 @@
...
+# Sets up a persistent gt4py cache directory based on backend and week to start
+# with a fresh cache every week. ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR is set as
+# the root and GT4PY_BUILD_CACHE_DIR is set to
+# ${ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR}/icon4py/${BACKEND}/uv-lock-<hash of uv.lock>-${DATE}.
+
+set -euo pipefail
+
+if [[ "${ICON4PY_CI_DISABLE_PERSISTENT_GT4PY_CACHE:-}" == "true" ]]; then
+    echo "Using non-persistent gt4py cache because ICON4PY_CI_DISABLE_PERSISTENT_GT4PY_CACHE is set"
+    export GT4PY_BUILD_CACHE_LIFETIME="session"
```
edopao: I had forgotten that it is just a matter of triggering the pipeline with:
`cscs-ci run default GT4PY_BUILD_CACHE_LIFETIME="session"`

The trigger variables should have precedence, according to the [documentation](https://docs.gitlab.com/ci/variables/#cicd-variable-precedence), so we do not need to introduce `ICON4PY_CI_DISABLE_PERSISTENT_GT4PY_CACHE` .

ci/default.yml
edopao:
```suggestion
        BACKEND: [dace_gpu, gtfn_gpu]
```

ci/distributed.yml
```diff
@@ -64,11 +64,12 @@ build_distributed:
     ICON4PY_TEST_DATA_PATH: "/icon4py/testdata"
     ICON4PY_ENABLE_GRID_DOWNLOAD: true
     ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
+    ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR: /gt4py_cache
     ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: false
     ICON4PY_DRIVER_LOGGING_LEVEL: critical
     GT4PY_BUILD_CACHE_LIFETIME: "persistent"
     PYTEST_ADDOPTS: "--durations=0"
-    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH"]'
+    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH","/iopsstor/scratch/cscs/svc_cwci02_cicd_ext:$ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR"]'
```
edopao, line 72: I don't know whether it could work:
```suggestion
    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH","$SCRATCH:$ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR"]'
```

ci/base.yml
```diff
@@ -94,7 +96,7 @@ build_image_aarch64:
   variables:
     RUNNER: "ci-runner"
     SYSTEM_TAG: "${FIRECREST_SYSTEM}-gh200"
-    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH"]'
+    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH","/iopsstor/scratch/cscs/svc_cwci02_cicd_ext:$ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR"]'
```
edopao, line 99: I don't know whether it could work:

```suggestion
    CSCS_ADDITIONAL_MOUNTS: '["/capstor/store/cscs/userlab/cwci02/icon4py/ci/testdata:$ICON4PY_TEST_DATA_PATH","$SCRATCH:$ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR"]'
```

ci/distributed.yml
msimberg: I tried this naively and it didn't work. I think the problem is that this gets interpreted on many levels through the cicd ext system/gitlab/etc. and I think putting the explicit directory is simpler, at least for now.

ci/scripts/gt4py-cache.sh
msimberg: Excellent idea, that would make it much more straightforward.

msimberg:
The latest runs are definitely not looking very good. There's a lot of https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14200851472#L655:
```
E           ../compute_inverse_on_edges.hpp:1:10: fatal error: gridtools/fn/backend/naive.hpp: No such file or directory
E               1 | #include <gridtools/fn/backend/naive.hpp>
E                 |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
which seems strange to me...

Most likely cause is that I've messed something up in the config, but @edopao does that look familiar to you for any reason? Worst case is that something in the concurrent cache writing still goes wrong.

edopao:
> The latest runs are definitely not looking very good. There's a lot of https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14200851472#L655:
> 
> ```
> E           ../compute_inverse_on_edges.hpp:1:10: fatal error: gridtools/fn/backend/naive.hpp: No such file or directory
> E               1 | #include <gridtools/fn/backend/naive.hpp>
> E                 |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```
> 
> which seems strange to me...
> 
> Most likely cause is that I've messed something up in the config, but @edopao does that look familiar to you for any reason? Worst case is that something in the concurrent cache writing still goes wrong.

I have no idea. This file should come from the `gridtools` python package.

ci/scripts/gt4py-cache.sh
```diff
@@ -0,0 +1,28 @@
...
+# First clean up files and directories older than 7 days in the base cache
+# directory. The backend directories can stay, but the date-based directories
+# are removed. There may be concurrent cleanup, ignore failures.
+find "${ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR}/icon4py" -mindepth 2 -maxdepth 2 -type d -mtime +7 -exec rm -rf {} + || true
+
+uv_lock_hash=$(sha256sum "./uv.lock" | awk '{print $1}')
+
+# Then set the cache directory for this run based on the backend and current date.
+DATE=$(date +%Y-%W)
+export GT4PY_BUILD_CACHE_DIR="${ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR}/icon4py/${BACKEND}/uv-lock-${uv_lock_hash}-${DATE}"
```
edopao: Note that it matters much, but maybe it would better to list the directories (with `ls`) in order of timestamp:
```suggestion
export GT4PY_BUILD_CACHE_DIR="${ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR}/icon4py/${BACKEND}/${DATE}-uv-lock-${uv_lock_hash}"
```

msimberg:
> > The latest runs are definitely not looking very good. There's a lot of https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14200851472#L655:
> > ```
> > E           ../compute_inverse_on_edges.hpp:1:10: fatal error: gridtools/fn/backend/naive.hpp: No such file or directory
> > E               1 | #include <gridtools/fn/backend/naive.hpp>
> > E                 |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which seems strange to me...
> > Most likely cause is that I've messed something up in the config, but @edopao does that look familiar to you for any reason? Worst case is that something in the concurrent cache writing still goes wrong.
> 
> I have no idea. This file should come from the `gridtools` python package.

Indeed, that's what I'd expect as well (and it's found on most pipelines...). I'll investigate.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default;GT4PY_BUILD_CACHE_LIFETIME=session;ICON4PY_CI_WIPE_GT4PY_CACHE=true

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default;GT4PY_BUILD_CACHE_LIFETIME=session;ICON4PY_CI_WIPE_GT4PY_CACHE=true

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

ci/scripts/gt4py-cache.sh
```diff
@@ -0,0 +1,35 @@
...
+job_name="${CI_JOB_NAME_SLUG}"
+
+# Then set the cache directory for this run based on the backend and current date.
+DATE=$(date +%Y-%W)
+export GT4PY_BUILD_CACHE_DIR="${ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR}/icon4py/${BACKEND}/uv-lock-${uv_lock_hash}-job-${job_name}-${DATE}"
+mkdir -p "${GT4PY_BUILD_CACHE_DIR}"
+
+echo "Using GT4PY_BUILD_CACHE_DIR=${GT4PY_BUILD_CACHE_DIR}"
+
+# TODO: This is here just for debugging, probably remove?
```
edopao: I don't know.. On the one hand, I find it useful to easily wipe the build cache (which I guess needs to be done by the CI user) without relying on the scratch cleanup policy. On the other hand, we would expose a variable that deletes stuff from the filesystem, so we introduce the risk for CI users of deleting folders by mistake.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

ci/scripts/gt4py-cache.sh
msimberg: Yeah, I'm undecided. I think ideally if one needs a fresh cache then one should just set cache lifetime to `session`, but I'm sure we'll sooner or later bump into a situation where we just need to start over with the cache. It's easily added again later. In the current version it's removed again.

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
@edopao I'd probably go ahead with the current version to give it some real-world time on other PRs. If you think it'd still be useful to have an explicit "wipe the cache" option I don't mind adding that. Of course let me know if you have other concerns.

For reference, the iopsstor scratch of svc_cwci02_cicd_ext now has ~100G space and 2M files used:
```
| /iopsstor/scratch/cscs/svc_cwci02_cicd_ext | LUSTRE |  97.1G |    - |      - |            - | 2160954 |    - |    -     |           - |
```
It's quite a lot of files, but there's no quota on either at the moment. If I remember correctly cleanup is 2 or 3 weeks, which is good enough for our use case (it's wiped after a week by us anyway).

model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
```diff
@@ -554,7 +554,13 @@ def _compare_metrics_fields_single_multi_rank(
         assert isinstance(field, state_utils.ScalarType)
         assert pytest.approx(field) == field_ref
     else:
-        if model_backends.is_cpu_backend(backend) and test_utils.is_dace(backend):
+        if test_utils.is_dace(backend) and (
```
edopao, line 557: I was not aware of this special case. We need to follow up on this next time we upgrade the gt4py version.

edopao (commented):
LGTM.

PR #1220, opened 2026-04-27
Add persistent gt4py cache for github workflows

msimberg:
PRs have their own cache which is written and restored for PRs only. Jobs on main will always write a fresh cache, never read a cache. On PRs, if the PR doesn't have a cache yet, the latest cache from main will be used.


msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

havogt (approved):
lgtm

.github/workflows/icon4py-test-model.yml
```diff
@@ -60,6 +60,20 @@ jobs:
...
+
+      - name: restore GT4Py cache
+        id: gt4py-cache-restore
+        if: github.ref != 'refs/heads/main'
+        uses: actions/cache/restore@v5
+        with:
+          path: ${{ env.GT4PY_BUILD_CACHE_DIR }}
+          key: gt4py-${{ runner.os }}-py${{ matrix.python-version }}-${{ matrix.backend }}-${{ matrix.component }}-branch-${{ github.head_ref || github.ref_name }}-${{ github.sha }}
+          restore-keys: |
+            gt4py-${{ runner.os }}-py${{ matrix.python-version }}-${{ matrix.backend }}-${{ matrix.component }}-branch-${{ github.head_ref || github.ref_name }}-
```
philip-paul-mueller: This is more me asking random questions.
It is probably similar to my remarks on the [other PR](https://github.com/C2SM/icon4py/pull/1218).
But, assumes you have done something in ICON4Py and run the CI, thus the cache is populated.
Then, you realize that you need to check if your feature also works with this new thing in the DaCe backend, thus you update `pyproject.toml`.
CI will then run, but as far as I can see because of the prefix matching you will find a cache entry that will be used.
But since you do not change the stencil code in ICON4Py it will not recompile a new, although it should.

I know this is kind of an exotic problem and it is probably not the most important one, but is it a problem (if not what do I miss) and do we care about it?

.github/workflows/icon4py-test-model.yml
```diff
@@ -60,6 +60,20 @@ jobs:
...
+            gt4py-${{ runner.os }}-py${{ matrix.python-version }}-${{ matrix.backend }}-${{ matrix.component }}-branch-main-
```
philip-paul-mueller: Another potential problem is the ordering of the keys that are checked.
As far as I can tell you assume that they are checked from newest to oldest.
To make an example, assume you have your PR and the ID of your first commit is `aaaa`, then you make a new one with the ID `ffff`, because there is only one match will use the `aaaa` cache.
The question is what cache the third commit use?
If they are checked against age it will use `fffff` if they are first alphabetically sorted it will use `aaaa`.
My question now is, is this a problem and do we care about it?

philip-paul-mueller (commented):
I have some questions (sorry for the wall of text).

.github/workflows/icon4py-test-model.yml
msimberg: It's a fair question here as well. My naive answer would be gt4py and dace should take care of that to make sure keys are unique if something changes :P but more realistically, would the simplest solution actually be just to include a hash of uv.lock into the key? That should catch gt4py and dace version changes, right?

.github/workflows/icon4py-test-model.yml
msimberg: It will use the newest. The only time the primary key would match is when the commit is unchanged, e.g. when a job is rerun. After that each of the restore keys is checked for prefix matches, and if there are multiple matches the newest one will be used. And the first restore key will be checked first and only if that has no matches will the second be checked.

I think we don't care, since it's supposed to be a cache so things should work with the cache or without the cache, or with a partial cache.

The important question is your other comment (https://github.com/C2SM/icon4py/pull/1220#discussion_r3167629891), where I now think including uv.lock in the key is likely the most elegant solution.

.github/workflows/icon4py-test-model.yml
msimberg: For reference (https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#cache-key-matchingm, emphasis mine):

> restore-keys allows you to specify a list of alternate restore keys to use when there is a cache miss on key. You can create multiple restore keys ordered from the most specific to least specific. The cache action searches the restore-keys in sequential order. When a key doesn't match directly, the action searches for keys prefixed with the restore key. **If there are multiple partial matches for a restore key, the action returns the most recently created cache.**

msimberg: I added a hash of uv.lock in https://github.com/C2SM/icon4py/pull/1220/commits/966f1ea15610fa3090ba37bb4d5cb7933217a274.

philip-paul-mueller: Thanks for linking, for some reason I did not found it (okay I did not put much effort into it).
But then it makes sense.

.github/workflows/icon4py-test-model.yml
philip-paul-mueller: Makes sense.

philip-paul-mueller (approved):
LGTM

.github/workflows/icon4py-test-model.yml
```diff
@@ -60,6 +60,20 @@ jobs:
         run: |
           .github/validation_grid_utils.py download
 
+      - name: set GT4Py cache directory
+        run: echo "GT4PY_BUILD_CACHE_DIR=${{ runner.temp }}/gt4py-cache" >> $GITHUB_ENV
```
edopao, line 67: I think that mixing `dace_cpu` and `dace_gpu` backends in the same cache folder does not work (at least it didn't last time I tried). This setting works under the assumption that one job runs only one backend, which is the case now. Overall, no problem, but better to keep it in mind or maybe add a note.

.github/workflows/icon4py-test-model.yml
```diff
@@ -60,6 +60,20 @@ jobs:
...
+
+      - name: restore GT4Py cache
+        id: gt4py-cache-restore
+        if: github.ref != 'refs/heads/main'
+        uses: actions/cache/restore@v5
+        with:
+          path: ${{ env.GT4PY_BUILD_CACHE_DIR }}
+          key: gt4py-${{ runner.os }}-${{ hashFiles('uv.lock') }}-py${{ matrix.python-version }}-${{ matrix.backend }}-${{ matrix.component }}-branch-${{ github.head_ref || github.ref_name }}-${{ github.sha }}
```
edopao, line 75: Just for me to understand. Is this line restoring the cache from the **parent** commit on main?

edopao (commented):
LGTM. Only a couple of questions to be sure I understand the cache configuration.

.github/workflows/icon4py-test-model.yml
edopao: On a second thought, the cache does not need to come from the parent commit apart from the changes in `uv.lock`, which you have already addressed. When testing a gt4py program in a different "context", the program jitting should take care of matching or not in the build cache.

msimberg: Exactly. My understanding is that the cache from main (if there's no other match) will be the newest matching cache, so not necessarily parent. With the uv.lock hashing this should still work well (in the sense that the cache should be _compatible_ but not necessarily _complete_).

.github/workflows/icon4py-test-model.yml
msimberg: Good point. I'll add a comment on this.

msimberg: Added comment in https://github.com/C2SM/icon4py/pull/1220/commits/02c22e1a27a0dc394b8d1e5395a34b1924794908.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1222, opened 2026-04-27
Fix unnecessary use of `xp.ones`

### LINKED ISSUE #1203: Check for x * xp.ones patterns in codebase
Replace `x * xp.ones` patterns in the codebase with `xp.full(x)` to directly initialize the array. For cupy compatibility, if x is an enum value, it may have to use `x.value`.

msimberg:
Fixes #1203.


OngChia (approved):
Looks good to me. Only one question, how about those `np.ones`? Should we also change them to `np.full`?

msimberg:
> Looks good to me. Only one question, how about those `np.ones`? Should we also change them to `np.full`?

Good catch, they should definitely be changed. Also array_ns.ones... changed those in the latest push as well. Thanks for catching that!

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

OngChia (approved):
LGTM

jcanton (approved):
LGTM

PR #1226, opened 2026-04-29
Fix typo in field names in driver_utils.py

msimberg:
The wrong field was used to initialize dual_normal_vert_y.

The bug was found by GLM 5.1: https://hackmd.io/cuEsccnkTniYTn0ABCa7LQ?view#H6-dual_normal_vert_y-uses-wrong-geometry-attribute-normal-vs-tangent.

Note that I'm not changing the names of the members in EdgeParams to use u/v instead of x/y as initially discussed. The *_x/y suffixes are as in ICON fortran and the dual_normal_vert etc. names also come from ICON fortran. If we change x/y then the whole name should be changed in one go, so I'm leaving that unchanged right now.

This change does not change the comparison to ICON fortran serialized data.


nfarabullini (approved):

nfarabullini:
cscs-ci run default

nfarabullini:
cscs-ci run distributed

msimberg:
Aah, I guess the typo comes from https://github.com/C2SM/icon-exclaim/blob/8b73908e1dfb5755565ce8f1b78917013d7eaaf8/src/serialization/mo_icon4py_verification.f90#L878... I'm hoping CI would actually fail with this "fix". 

Edit: wrong PR. This comment was about https://github.com/C2SM/icon4py/commit/bd8b164ecac9702bd6aa6e1778926f8851be7f58, which I was planning to "fix".

PR #1227, opened 2026-04-29
Remove dace orchestration functionality

msimberg:
Removes dace orchestration functionality as it's outdated and unmaintained.

This would allow some cleanups to the halo exchange functionality, but I'm not mixing those changes into this PR. This is pure removal.


msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

kotsaloscv (approved):

philip-paul-mueller:
@kotsaloscv I am actually not sure if `HaloExchangeWaitRuntime` can also be removed since it essentially calls `finish()`?
When I updated the halo exchanges it really confused me as it essentially only a layer that turns `self.exchange.exchange_and_wait(...)` into `self.exchange(..., wait=True)`.
To me it seems like a layer that was introduced to handle some limitation of DaCe's Python frontend.

havogt (approved):
lgtm

msimberg:
> @kotsaloscv I am actually not sure if `HaloExchangeWaitRuntime` can also be removed since it essentially calls `finish()`? When I updated the halo exchanges it really confused me as it essentially only a layer that turns `self.exchange.exchange_and_wait(...)` into `self.exchange(..., wait=True)`. To me it seems like a layer that was introduced to handle some limitation of DaCe's Python frontend.

I'll try to have a closer look after this is merged whether that can be removed or not...

PR #1191, opened 2026-04-10
Pull test data from CSCS object store

msimberg:

PR #1232, opened 2026-04-30
build: development scripts conventions

egparedes:
This PR refactors the development scripts folder so they follow the project conventions for different kind of standalone scripts. It also adds an optional single-entry point to run any of the scripts The design and instructions are in the README file.


scripts/python/validation_grid_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        definitions.Grids.MCH_CH_R04B09_DSL,
+        definitions.Grids.MCH_OPR_R04B07_DOMAIN01,
+        definitions.Grids.TORUS_50000x5000,
+    ]  # change to MCH_OPR_R04B07_DOMAIN01
+
+
+@cli.command(name="cache-key")
+def cache_key() -> None:
+    """Generate a cache key for the Github action cache based on grid file name and download URI."""
+    d = "_".join(grid.name + grid.uri for grid in get_validation_grids())
```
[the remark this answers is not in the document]
msimberg: This was changed on main btw (in #1191).

scripts/run
```diff
@@ -0,0 +1,128 @@
...
+IDE navigation, and static analysis). Shell scripts under ``sh/`` are
+auto-discovered and wrapped with a uniform argument-forwarding interface.
+
+Dependencies for all sub-commands are defined in the `scripts` table of
+the root `pyproject.toml`.
+
+Usage:
+    ./scripts/run --help
+    ./scripts/run lint --fix src/
+    ./scripts/run setup-env
```
msimberg: These aren't actual sub-commands (or it's not working for me?), so I'd remove them or replace them with something real.

scripts/run
```diff
@@ -0,0 +1,128 @@
...
+import typer
+
+if __name__ == "__main__":
+    try:
+        # First try to import the 'python' folder inside './scripts'
+        # and check that it is indeed the expected one.
+        import python
+
+        python_scripts_dir = str(pathlib.Path(__file__).resolve().absolute().parent / "python")
+        assert len(python.__path__) == 1 and python.__path__[0] == python_scripts_dir, (
```
msimberg: Should this raise an exception directly instead of asserting?

scripts/run
```diff
@@ -0,0 +1,128 @@
...
+        sys.exit(127)
+
+    # Build the CLI app and register all sub-commands from scripts
+    app = typer.Typer(
+        name="scripts",
+        help="Project dev-scripts toolbox.",
+        no_args_is_help=True,
+    )
+
+    # ── Python sub-commands ──────────────────────────────────────────────────────
```
msimberg: Ultimate bike-shed:

```suggestion
    # Python sub-commands
```
? Not because I particularly care, but only because this looks like an LLM-choice (I really don't mind leaving it as is, I just wouldn't make headings like this myself).

scripts/README.md
```diff
@@ -0,0 +1,97 @@
+# scripts/ — Dev-Scripts Toolbox
+
+Conventions for standardized project development scripts using Python, Bash, and (possibly) other recipes. All scripts should be executable and testable standalone, but a single entry point (`./scripts/run`) is provided as an extra feature, mostly for exploration.
```
msimberg: I don't think you need to say it's "mostly for exploration". It's very nice that everything is usable standalone, but it's likewise nice to have a common entrypoint.

msimberg (approved):
Only bikeshedding nits, this worked nicely for me locally. The grid download script definitely needs a fix before merging though.

scripts/python/compare_icon4py_openacc.py
```diff
@@ -6,19 +8,24 @@
...
-import matplotlib.pyplot as plt
-import numpy as np
+import typer
 
 
 VariantDescriptor: TypeAlias = tuple[str, dict[str, Any]]
 
+cli = typer.Typer()
+
+log = logging.getLogger(__name__)
```
edopao, line 28: Should we now use `typer.echo()` instead of `logging`?

edopao (commented):
Advanced, but very nice!

scripts/python/validation_grid_utils.py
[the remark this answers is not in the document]
egparedes: Updated.

scripts/python/compare_icon4py_openacc.py
egparedes: I didn't want to change the core of the scripts to avoid introducing bugs, just the minimal changes to make them compatible with the infrastructure. This script has plenty of room for improvement and your suggestion makes a lot of sense, but I would do it in a different PR

PR #1235, opened 2026-05-04
AMD profiling: cluster node variance investigation + tooling

dganellari:
extract_pmc.py: rewrite with three cache-plane columns (L1<->L2, L2-Fabric, HBM) matching rocprof-compute analyze section 4.1.9 within ~1%, validated on 5 nodes across both clusters and rocprof-compute versions 3.4.0 and 4.x.

CLUSTER_NODE_VARIANCE.md: cross-cluster comparison (aac6 vs beverin), within-cluster silicon binning (13-26% spread), per-kernel HBM tables, firmware/driver comparison, reproducer commands, methodology gotchas.

setup_env.sh: uenv-agnostic (autodetect ROCM_VERSION + rocprofiler-dev path).
setup_env_rocm72.sh: beverin icon4py-rocm72 venv setup with gt4py PR 2578.
capture_env.sh, trace_power.sh: env capture and under-load power/clock trace.

Sbatch scripts (cluster-paired aac6/rocm72): gt4py_timer, rocprof_compute, trace_power. Plus job-capture-env.sh, job-rocprof-compute-improved.sh aac6 helpers.

benchmark_solver.sh: switch from .venv to .venv_rocm.
PROFILING_RESULTS.md, DEEP_ANALYSIS.md: updated extract_pmc references.


amd_scripts/benchmark_solver.sh
```diff
@@ -14,7 +14,7 @@ cd $ICON4PY_GIT_ROOT
 # Set necessasry flags for compilation
 source amd_scripts/setup_env.sh
 
-source .venv/bin/activate
```
iomaganaris, line 17: I think that to make this work out of the box we have to update the `amd_scripts/install_icon4py_venv.sh` script and the rest of the scripts that use the `.venv` as well

amd_scripts/extract_pmc.py
```diff
@@ -50,11 +74,33 @@ def main():
...
+        ("req", "TCC_REQ_sum"),
+        ("read", "TCC_READ_sum"),
+        ("write", "TCC_WRITE_sum"),
+        ("rdreq", "TCC_EA0_RDREQ_sum"),
+        ("rdreq_32b", "TCC_EA0_RDREQ_32B_sum"),
+        ("rdreq_dram", "TCC_EA0_RDREQ_DRAM_sum"),
+        ("wrreq", "TCC_EA0_WRREQ_sum"),
+        ("wrreq_64b", "TCC_EA0_WRREQ_64B_sum"),
+        ("wrreq_dram", "TCC_EA0_WRREQ_DRAM_sum"),
+        ("bubble", "TCC_BUBBLE_sum"),
```
iomaganaris, line 105: Are these generated by default using `rocprof-compute` or do we have to adjust which counters are included?

amd_scripts/setup_env_rocm72.sh
```diff
@@ -0,0 +1,148 @@
+#!/usr/bin/env bash
```
iomaganaris: This file looks more or less like a copy of `amd_scripts/install_icon4py_venv.sh`. I should either remove the existing script or avoid adding a new one and add the necessary functionality to the old one. I would keep it as simple as possible in any case without mentions of certain PRs, branches and directories

amd_scripts/job-rocprof-compute-improved.sh
```diff
@@ -0,0 +1,82 @@
...
+rocprof-compute profile --no-roof --name rcu_${GT4PY_BUILD_CACHE_DIR} ${ROCPROF_COMPUTE_KERNEL_NAME_FILTER} --format-rocprof-output csv --kernel-names -R FP64 -p ${OUTPUT_DIR} -- \
+    python3 -m pytest -sv \
+    -m continuous_benchmarking \
+    -p no:tach \
+    --backend=dace_gpu \
+    --grid=${ICON_GRID} \
+    model/atmosphere/dycore/tests/dycore/stencil_tests/test_vertically_implicit_dycore_solver_at_predictor_step.py \
+    -k "test_TestVerticallyImplicitSolverAtPredictorStep[compile_time_domain-at_first_substep[False]__is_iau_active[False]__divdamp_type[32]]"
+
+git checkout model/common/src/icon4py/model/common
```
iomaganaris: This is used to revert the changes to `$HOME/icon4py/model/common/src/icon4py/model/common/model_options.py`?

amd_scripts/job-rocprof-compute-improved.sh
```diff
@@ -0,0 +1,82 @@
+#!/bin/bash
```
iomaganaris: This file needs renaming maybe or it should be integrated to one of the other `aac6` scripts

amd_scripts/benchmark_solver.sh
dganellari: .venv_rocm → .venv reverted.

amd_scripts/extract_pmc.py
dganellari: the 11 counters are in rocprof-compute's default set on both 7.1 and 7.2 (no -c override needed) — added a code comment to make that explicit.

amd_scripts/setup_env_rocm72.sh
dganellari: setup_env_rocm72.sh deleted. install_icon4py_venv.sh now finds the rocprofiler-compute requirements file with a wildcard pattern instead of a hardcoded ROCm 7.1 path, so the same script works on either uenv. Also dropped the three sbatch_*_rocm72.sh siblings.

amd_scripts/job-rocprof-compute-improved.sh
dganellari: the git checkout was reverting the in-place sed injection of gpu_block_size_2d — but the file is now deleted, replaced by the sbatch_*_aac6.sh scripts which don't need the patch.

amd_scripts/job-rocprof-compute-improved.sh
dganellari: job-rocprof-compute-improved.sh deleted; superseded by sbatch_*_aac6.sh.

iomaganaris (approved):
LGTM for merging now but let's discuss also the findings

PR #1239, opened 2026-05-06
Update Plottingscripts

philip-paul-mueller:
This PR adds the OpenACC baseline to the repo and ensures that the data is available. Furthermore, it removes the sole entry in `gt4py_input` since that file is not present.


philip-paul-mueller:
cscs-ci run default

philip-paul-mueller:
cscs-ci run distributed

edopao (dismissed):

edopao:
Adding @egparedes since he recently did some work in the scripts.

egparedes:
@edopao @philip-paul-mueller Both downloading the file or add it to repo would be ok for me. If the file easily available online and it might get out of sync at some point, then I'd rather download it and cache it within the script. If it's not longer available online and/or will never change, then we should add it to the repo, in a subfolder inside a general  `scripts/data` folder.

philip-paul-mueller:
As far as I understand it you can not simply use a different file, because the file and the script have both to reflect the FORTRAN timer structure.
Thus downloading (the newest version) might not work, because if there changes something then the script will stop working because of the changed names.
Thus I think checking in the JSON file is the best approach.

iomaganaris (approved):
LGTM

edopao:
I can approve, there is just the comment about renaming the folder to `scripts/data`. I think Enrique can then merge this PR bypassing the CI.

philip-paul-mueller:
@egparedes I relocated the script to the `data/` folder and updated the README.md.
I think that it is not worth running the CI could you force merge it?

egparedes (approved):
LGTM

PR #1244, opened 2026-05-06
Fix license specification in muphys pyproject.toml

msimberg:
Use "text" instead of "file", as in other pyproject.toml files.

https://hackmd.io/cuEsccnkTniYTn0ABCa7LQ?both#H11-muphys-pyprojecttoml-license-points-to-non-existent-file


msimberg:
I'd merge this without cscs ci if there's no objection...

egparedes (approved):
LGTM

PR #1245, opened 2026-05-06
Improve return code handling in py2fgen generated code

havogt:
Fixes and old issue: if the Python startup fails (typically because icon4py (and its dependencies) is not in the active Python environment), then Fortran progresses, but Python is skipped. Simulation ends with "success" but obviously wrong results. Since cffi returns a hard-coded `0` always on an init fail, we make the success path returning `1`.

From the comments:
>rc convention (chosen so init-fail is distinguishable from success — cffi
forces the result slot to 0 on init-fail, see _cffi_start_and_call_python
in _embedding.h of cffi):
  0 -> embedding init failed
  1 -> success
  2 -> Python wrapper raised an exception


havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

egparedes (approved):
LGTM but I find confusing to use 0 for errors instead of the opposite. I guess it's really impossible to change with FFI, so I'm approving it anyway.

havogt:
Yes, annoying. hard-coded here https://github.com/python-cffi/cffi/blob/c5212d08d659acc1ac1299b576ad859a8efdb3c1/src/cffi/_embedding.h#L478

PR #1248, opened 2026-05-06
Bump minimum supported python to 3.12

msimberg:
As a follow-up to #1197 which adds support for python 3.14, this bumps the minimum version to 3.12 and applies some modernizations along the way.


msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/states/utils.py
```diff
@@ -16,14 +16,14 @@
...
+FloatType: TypeAlias = ta.wpfloat | ta.vpfloat | float  # noqa: UP040
+IntegerType: TypeAlias = gtx.int32 | gtx.int64 | int  # noqa: UP040
+ScalarType: TypeAlias = FloatType | bool | IntegerType  # noqa: UP040
 
 T = TypeVar("T", ta.wpfloat, ta.vpfloat, float, bool, gtx.int32, gtx.int64)
 
-GTXFieldType: TypeAlias = gtx.Field[DimsT, T]
-FieldType: TypeAlias = gtx.Field[DimsT, T] | data_alloc.NDArray
+GTXFieldType: TypeAlias = gtx.Field[DimsT, T]  # noqa: UP040
+FieldType: TypeAlias = gtx.Field[DimsT, T] | data_alloc.NDArray  # noqa: UP040
```
egparedes, line 26: I guess these type aliases are left in the old form just because of gt4py doesn't support the new form, right? I would add a TODO here documenting the issue, and also as a reminder to change it once gt4py is updated.

model/common/src/icon4py/model/common/field_type_aliases.py
```diff
@@ -16,12 +16,12 @@
 
 T = TypeVar("T", wpfloat, vpfloat, float, bool, gtx.int32, gtx.int64)
 
-CellField: TypeAlias = Field[Dims[dims.CellDim], T]
-EdgeField: TypeAlias = Field[Dims[dims.EdgeDim], T]
-VertexField: TypeAlias = Field[Dims[dims.VertexDim], T]
-KField: TypeAlias = Field[Dims[dims.KDim], T]
-KHalfField: TypeAlias = Field[Dims[dims.KHalfDim], T]
+CellField: TypeAlias = Field[Dims[dims.CellDim], T]  # noqa: UP040
```
egparedes, line 19: Same comment as above

.opencode
```diff
@@ -0,0 +1 @@
+/home/mjs/src/icon4py/.opencode
```
egparedes: This file should be removed, right?

shell.nix
```diff
@@ -0,0 +1 @@
+/home/mjs/src/icon4py/shell.nix
```
egparedes: The file should be removed, right?

egparedes (commented):
Looks good to me, just a comment to document a ToDo item. Copilot comments are also sensible.

model/common/src/icon4py/model/common/states/utils.py
msimberg: I didn't check what gt4py supports, but the main reason for still using TypeAlias here is that the new `type` keyword doesn't allow using a `type T = ...` in the `type GTXFieldType = gtx.Field[DimsT, T]` declaration.

So if gt4py is not a restriction, we could use the new form, but would have to repeat the type union of wpfloat, vpfloat, etc. for GTXFieldType and FieldType. Not the end of the world here, a bit uglier in model/common/src/icon4py/model/common/field_type_aliases.py (https://github.com/C2SM/icon4py/pull/1248#discussion_r3493457529).

msimberg:
Side note: `icon/26.7:v1` is now deployed on santis with python 3.13, so we could finally go ahead with this.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run merge

msimberg:
Waiting for #1299 to be merged first (uv.lock will conflict).

msimberg:
cscs-ci run merge

msimberg:
cscs-ci run merge

Issue #1137, opened 2026-03-26
Improve type checking coverage


msimberg:
Currently a large number of files and checks are disabled for mypy: https://github.com/C2SM/icon4py/blob/10105f8ef0a4906aa67711530addfb7d3a195a6d/pyproject.toml#L137-L230.

This will have to be done in stages. Pick _any_ disabled check or file and try to resolve type errors that come up (it's probably best to prioritize library code, i.e. not the tests, but everything benefits from checking eventually).

msimberg:
One interesting case worth focusing on: https://github.com/C2SM/icon4py/pull/1251#discussion_r3225212926. TL;DR `dims` should generally be a tuple, so one-element dims should be `(dim,)` instead of `(dim)`. Type hints are there but likely not being checked.

PR #1251, opened 2026-05-07
Fix miscellaneous typos and other minor issues

msimberg:
Fixes the most trivial issues from https://hackmd.io/cuEsccnkTniYTn0ABCa7LQ. As far as I can tell, nothing important was broken.


msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

model/common/src/icon4py/model/common/metrics/metrics_attributes.py
```diff
@@ -443,23 +443,23 @@
         standard_name=DEEPATMO_DIVH,
         long_name="",
         units="",
-        dims=(dims.KDim),
+        dims=(dims.KDim,),
```
jcanton, line 446: can we add these to mypy or open an issue about adding them to mypy?

jcanton (approved):

model/common/src/icon4py/model/common/metrics/metrics_attributes.py
msimberg: I think it's implicit in #1137, but I'll add a note about this particular case to the issue.

Edit: FieldMetaData does have type hints and dims has `Sequence[gtx.Dimension]`, so I think if we were checking for it it should be caught.

msimberg: https://github.com/C2SM/icon4py/issues/1137#issuecomment-4429467817

PR #1253, opened 2026-05-07
py2fgen: all_bindings generate .f90 and .c separately

havogt:
Create all_bindings as executable module which can generate .f90 and .c files separately. ICON integration will call this script instead of py2fgen directly.

Additional: general cleanups in py2fgen and bindings, e.g. (almost) remove a test dependency from py2fgen to bindings.


havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run default

bindings/src/icon4py/bindings/all_bindings.py
```diff
@@ -6,22 +6,63 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+"""CLI entrypoint that generates the combined icon4py Fortran/C bindings.
 
+Run as ``python -m icon4py.bindings.all_bindings --output-f90 X --output-c Y``.
```
egparedes, line 11: Just nitpicking, feel free to ignore, but isn't `bindings.all_bindings` a bit redundant? Maybe something like `cli.py` would be a better name. Actually, it could be just `__main__.py` and then can call it just like `python -m icon4py.bindings`

tools/src/icon4py/tools/py2fgen/_utils.py
```diff
@@ -54,10 +46,11 @@ def write_file_if_changed(
...
     False if the existing file already has the same content.
     """
-    path = outdir / fname
-    if not force and path.exists() and path.read_text() == content:
+    if not force and path.exists() and path.read_text(encoding="utf-8") == content:
         return False
-    write_file(content, outdir, fname)
+    path.parent.mkdir(exist_ok=True, parents=True)
+    with path.open("w", encoding="utf-8") as f:
+        f.write(content)
```
egparedes: For a coherent use of `pathlib` and the statement above, I would do:

```suggestion
    path.write_text(content, encoding="utf-8")
```

egparedes (approved):
LGTM

bindings/src/icon4py/bindings/all_bindings.py
havogt: I am not sure. It's actually "all bindings for ICON Fortran". Currently there are also the muphys bindings which are in this package, but not part of all_bindings. Since I won't have time to do this change, I'd prefer to do it in a later cleanup.

havogt:
cscs-ci run default

havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

PR #1254, opened 2026-05-08
Save json namelist during serialization

jcanton:
and bug-fix the run_serialization script with lazy loading


jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

scripts/python/run_serialization.py
```diff
@@ -514,12 +515,28 @@ def run_experiment(
...
 @cli.command()
 def run_serialization() -> None:
     """Run the serialization experiment series."""
+
+    # Import here to reduce startup time for the CLI
+    import icon4py.model.testing.datatest_utils as _dt_utils
+    import icon4py.model.testing.definitions as _definitions
+
+    globals()["dt_utils"] = _dt_utils
+    globals()["definitions"] = _definitions
```
msimberg: I guess

```suggestion
    # Import here to reduce startup time for the CLI
    global dt_utils
    global definitions
    
    import icon4py.model.testing.datatest_utils as dt_utils
    import icon4py.model.testing.definitions
```
did not work?

In any case, deferring to @egparedes for authoritative comments on this approach.

scripts/python/run_serialization.py
```diff
@@ -514,12 +515,28 @@ def run_experiment(
...
+
     settings = SerializationSettings.defaults()
     settings.output_root.mkdir(parents=True, exist_ok=True)
 
+    require_cli("f90nml")
```
msimberg: Is the idea with this that the user might not have `uv sync`-ed and if the data has already been serialized it's awkward to run f90nml after the data has been serialized?

If yes, why not instead use
```
import f90nml
nml = f90nml.read(...)
json.dump(nml.todict(), ...)
```
?

The f90nml eecutable does internally this: https://github.com/marshallward/f90nml/blob/34da4c873d3eb169da7892b54f868b077d4a9c27/f90nml/cli.py#L175-L178. I think just doing that directly in the script instead of calling the `f90nml` script would simplify things and avoid this check.

msimberg (changes requested):
Not much to say, except I think the f90nml would be simplified by importing f90nml instead of calling the script (but I might miss a reason why it should be done the current way).

scripts/python/run_serialization.py
jcanton: ruff didn't like it, but the dict approach isn't much better either...
so I've implemented the suggestiond with a `# noqa: PLW0603`

scripts/python/run_serialization.py
jcanton: I think initially we didn't want to add another dependency, but now there's the `group scripts`  in pyproject so this can be implemented as you suggested. done

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

scripts/python/run_serialization.py
```diff
@@ -517,6 +517,12 @@ def run_experiment(
 @cli.command()
 def run_serialization() -> None:
     """Run the serialization experiment series."""
+
+    # Import here to reduce startup time for the CLI
+    global dt_utils, definitions  # noqa: PLW0603
```
msimberg:
```suggestion
    global dt_utils, definitions  # noqa: PLW0603 [global-statement]
```

since I now know about `CODING_GUIDELINES.md`. Specifically:
> Additionally, you should add a short descriptive code if possible (check [ruff rules](https://docs.astral.sh/ruff/rules/) and [mypy error codes](https://mypy.readthedocs.io/en/stable/error_code_list.html) for reference):

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

scripts/python/run_serialization.py
jcanton: add to `AGENTS.md`? https://github.com/C2SM/icon4py/pull/1254/commits/7b92092e4dc3b60c4c8259022b872c531c2a9468

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

msimberg (approved):

scripts/python/run_serialization.py
msimberg: #1256

PR #1257, opened 2026-05-11
Make grid and testdata download explicitly true

iomaganaris:
- Fixes bencher CI issue: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504672/-/jobs/14297540064
```
    def _download_grid_file(grid: definitions.GridDescription) -> pathlib.Path:
        full_name = dt_utils.get_grid_filepath(grid)
        grid_directory = full_name.parent
        grid_directory.mkdir(parents=True, exist_ok=True)
        if config.ENABLE_GRID_DOWNLOAD:
            uri = dt_utils.get_grid_archive_url(definitions.TESTDATA_ROOT_URL, grid)
            data_handling.download_and_extract(
                uri,
                grid_directory,
            )
        else:
            # If grid download is disabled, we check if the file exists
            # without locking. We assume the location is managed by the user
            # and avoid locking shared directories (e.g. on CI).
            if not full_name.exists():
>               raise FileNotFoundError(
                    f"Grid file {full_name} does not exist and grid download is disabled."
                )
E               FileNotFoundError: Grid file /icon4py/testdata/grids/icon_grid_0021_R02B06_G/icon_grid_0021_R02B06_G.nc does not exist and grid download is disabled.
```
- Update bencher version

### DIFF
--- a/.github/workflows/icon4py-test-model.yml
+++ b/.github/workflows/icon4py-test-model.yml
@@ -12,8 +12,8 @@ on:
 env:
   ICON4PY_TEST_DATA_PATH: "${{ github.workspace }}/testdata"
   NUM_PROCESSES: auto
-  ICON4PY_ENABLE_GRID_DOWNLOAD: true
-  ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
+  ICON4PY_ENABLE_GRID_DOWNLOAD: false
+  ICON4PY_ENABLE_TESTDATA_DOWNLOAD: false
   PYTEST_ADDOPTS: "--durations=0"
 
 jobs:
--- a/.github/workflows/icon4py-test-model.yml
+++ b/.github/workflows/icon4py-test-model.yml
@@ -12,8 +12,8 @@ on:
 env:
   ICON4PY_TEST_DATA_PATH: "${{ github.workspace }}/testdata"
   NUM_PROCESSES: auto
-  ICON4PY_ENABLE_GRID_DOWNLOAD: false
-  ICON4PY_ENABLE_TESTDATA_DOWNLOAD: false
+  ICON4PY_ENABLE_GRID_DOWNLOAD: true
+  ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
   PYTEST_ADDOPTS: "--durations=0"
 
 jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/icon4py-test-model.yml
iomaganaris, line 61: Not sure whether we should remove this step or not

msimberg: Ah, sorry @iomaganaris that I led you down a path I didn't intend for. I'd leave downloads disabled in the github workflows since we have a clean way to cache that (and caching is good to avoid hitting our ceph storage every job).

In the cscs ci workflows our caching is effectively the filesystem (which is also good, but which we can't use in the github workflows).

iomaganaris: I see. I also agree that using the filesystem is a good caching approach since the files should already be there and shouldn't change much so maybe we should just make sure that all the necessary files exist in the proper directories instead of enabling downloads?

msimberg: The old approach was to always manually populate the filesystem for CSCS CI, which is why the download options were off, but I think it's nicer to keep it enabled now that we have (hopefully) working concurrent downloads etc. And yeah, they should not be changed very often, but when they do it's annoying to have to enable downloads and/or ask someone to upload the files to the filesystem. This is more important for serialized data than for grid files, but I think we can keep both enabled in CSCS CI.

I'd rather try to make the github workflows cache more dynamic (it currently reacts to grid name changes/locations correctly, but one still has to manually add a grid to the list of grids to download). I think it's not very urgent to update that though and it works well enough exactly because the grids rarely change. I'd definitely not try to change that in this PR.

Does that make sense?

iomaganaris:
cscs-ci run benchmark-bencher

.github/workflows/icon4py-test-model.yml
iomaganaris: Yes, it makes sense 👍 I have only enabled the downloads for the bencher cscs-ci then

msimberg (approved):

iomaganaris:
cscs-ci run benchmark-bencher

kotsaloscv (approved):

iomaganaris:
cscs-ci run benchmark-bencher

iomaganaris:
cscs-ci run benchmark-bencher

iomaganaris:
cscs-ci run benchmark-bencher

iomaganaris:
cscs-ci run benchmark-bencher

iomaganaris:
cscs-ci run default

iomaganaris:
cscs-ci run distributed

PR #1259, opened 2026-05-11
suggest keyword guidelines

jcanton:


msimberg (approved):
Already discussed offline, but I agree with this as a general/default rule. As a guideline, it can still be broken when appropriate. But @egparedes and @havogt have more experience with python to say if this is a bad idea to have as a default.

jcanton:
instead of guidelines, we will replace withruff rules https://docs.astral.sh/ruff/rules/too-many-positional-arguments/

PR #1161, opened 2026-03-26
add warm bubble initial condition

OngChia:
- Add weisman_klemp() initial condition implementing the WK82 atmospheric test case (Warm bubble) with piecewise vertical integration from the tropopause, moisture profile
  with qv_max capping, hydrostatic adjustment, and warm bubble perturbation on torus geometry
- Extend ```standalone_driver``` to support the Weisman-Klemp experiment via ```experiment_name```      
  parameter, with flat topography initialization and WK-specific vertical grid config
  ```(nlev=64, model_top_height=30000, stretch_factor=0.85)```                            
 - Add integration test test_weisman_klemp_initial_condition validating ```rho```,```exner```, ```theta_v```, ```vn```, ```w```, ```qv```, and ```exner_pr``` against Fortran serialized reference data

PR #1263, opened 2026-05-12
Experiment configs

jcanton:
replace hardcoded fixtures with configs from json namelists


jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

bindings/tests/bindings/test_diffusion_wrapper.py
```diff
@@ -47,35 +45,34 @@ def test_diffusion_wrapper_granule_inputs(
...
+    smagorinski_scaling_factor4 = cfg.smagorinski_scaling_factor4
+    smagorinski_scaling_height = cfg.smagorinski_scaling_height
+    smagorinski_scaling_height2 = cfg.smagorinski_scaling_height2
+    smagorinski_scaling_height3 = cfg.smagorinski_scaling_height3
+    smagorinski_scaling_height4 = cfg.smagorinski_scaling_height4
+    zdiffu_t = cfg.apply_zdiffusion_t
+    denom_diffu_v = cfg.velocity_boundary_diffusion_denominator
+    max_nudging_coefficient = cfg.max_nudging_coefficient
+    itype_sher = cfg.shear_type
+    ndyn_substeps = cfg.ndyn_substeps
```
msimberg: Probably not for this PR, but it'd be neat if we could just pass the `cfg` object directly to e.g. `diffusion_init`.

For now though, do you think it'd make sense to remove all of these local variables here except for `cfg = experiment.config.diffusion`, and just refer to the variables later as `cfg.<config option>`? In my opinion these local variables add little value now that we have a real config object here.\

Same comment applies to all the tests.

scripts/python/run_serialization.py
```diff
@@ -514,6 +514,12 @@ def run_experiment(
         raise
 
 
+def require_cli(command_name) -> None:
```
msimberg: This is old, right? Can you check fi there are other outdated changes included in the diff?

model/atmosphere/subgrid_scale_physics/microphysics/tests/microphysics/integration_tests/test_saturation_adjustment.py
```diff
@@ -25,30 +25,29 @@
...
     from icon4py.model.testing import serialbox as sb
 
 
 @pytest.mark.embedded_static_args
 @pytest.mark.datatest
 @pytest.mark.parametrize(
-    "experiment, model_top_height",
-    [(definitions.Experiments.WEISMAN_KLEMP_TORUS, 30000.0)],
+    "experiment",
+    [(definitions.Experiments.WEISMAN_KLEMP_TORUS)],
```
msimberg:
```suggestion
    [definitions.Experiments.WEISMAN_KLEMP_TORUS],
```

model/atmosphere/subgrid_scale_physics/microphysics/tests/microphysics/integration_tests/test_single_moment_six_class_gscp_graupel.py
```diff
@@ -38,31 +38,25 @@
 @pytest.mark.embedded_static_args
 @pytest.mark.datatest
 @pytest.mark.parametrize(
-    "experiment, model_top_height",
+    "experiment",
     [
-        (definitions.Experiments.WEISMAN_KLEMP_TORUS, 30000.0),
+        (definitions.Experiments.WEISMAN_KLEMP_TORUS),
```
msimberg:
```suggestion
        definitions.Experiments.WEISMAN_KLEMP_TORUS,
```

model/testing/src/icon4py/model/testing/fixtures/benchmark.py
```diff
@@ -116,13 +116,8 @@ def metrics_field_source(
         geometry_source=geometry_field_source,
         topography=gtx.as_field((dims.CellDim,), data=topo_c),  # type: ignore[arg-type]  # NDArrayObject is not exported from gt4py
         interpolation_source=interpolation_field_source,
+        config=metrics_factory.MetricsConfig(),
```
msimberg, line 123: Do defaults make sense here or would it be better to make an explicit choice here?

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -96,3 +106,217 @@ def create_icon_serial_data_provider(
...
+    nonhydrostatic_nml = nml_data["nonhydrostatic_nml"]
+    interpol_nml = nml_data["interpol_nml"]
+    diffusion_nml = nml_data["diffusion_nml"]
+    turbdiff_nml = nml_data["turbdiff_nml"]
+    run_nml = nml_data["run_nml"]
+
+    # *** MetricsConfig ***
+    rayleigh_coeff = nonhydrostatic_nml["rayleigh_coeff"]
+    if isinstance(rayleigh_coeff, list):
+        rayleigh_coeff = rayleigh_coeff[0]
```
msimberg: Why do we need this? Should the fortran namelist to json conversion rather take care of this case?

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -96,3 +106,217 @@ def create_icon_serial_data_provider(
...
+
+    metrics_config = metrics_factory.MetricsConfig(
+        exner_expol=nonhydrostatic_nml["exner_expol"],
+        vwind_offctr=nonhydrostatic_nml["vwind_offctr"],
+        thslp_zdiffu=nonhydrostatic_nml["thslp_zdiffu"],
+        thhgtd_zdiffu=nonhydrostatic_nml["thhgtd_zdiffu"],
+        rayleigh_type=constants.RayleighType(nonhydrostatic_nml["rayleigh_type"]),
+        rayleigh_coeff=rayleigh_coeff,
```
msimberg, line 209: Inline like `rayleigh_damping_height`?

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -96,3 +106,217 @@ def create_icon_serial_data_provider(
...
+    )
+
+    # *** DiffusionConfig ***
+    lhdiff_smag_w = diffusion_nml["lhdiff_smag_w"]
+    if isinstance(lhdiff_smag_w, list):
+        hdiff_smag_w = lhdiff_smag_w[0]
+
+    lsmag_3d = diffusion_nml["lsmag_3d"]
+    if isinstance(lsmag_3d, list):
+        smag_3d = lsmag_3d[0]
```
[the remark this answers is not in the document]
msimberg: This seems like a real type. Is it `smag_3d` or `lsmag_3d`?

.gitignore
```diff
@@ -12,6 +12,7 @@ testdata
 simple_mesh*.nc
 /prof
 .profile
+pyrightconfig.json
```
msimberg, line 15: Is this something that should be checked in instead?

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -202,6 +204,35 @@ class Experiment:
...
+    @property
+    def config(self) -> ExperimentConfig:
+        """Lazily load the experiment configuration on first access."""
+        if self._config is None:
+            if self._process_props is None:
+                self._process_props = decomp_defs.SingleNodeProcessProperties()
+            dt_utils.download_experiment(self, self._process_props)
+            self._config = dt_utils.create_experiment_configuration(self, self._process_props)
+        # Return a deep copy so that tests cannot mutate the shared cached config.
+        return copy.deepcopy(self._config)
```
msimberg: I think this should likely rather be provided through an `ExperimentConfig` test fixture, no? Or did you try that and it can't be done for some reason?

If it's a fixture you can make it depend on process_props and experiment.

msimberg (changes requested):
Very nice cleanup. I've made a few minor comments. I think you could more aggressively remove local
```
param = cfg.param
```
variables? Additionally, unless there's a technical reason to avoid it, I think the experiment config should come from a fixture. I think it'd clean things up further quite nicely.

model/testing/src/icon4py/model/testing/datatest_utils.py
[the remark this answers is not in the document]
jcanton: they are both bools both in fortran and in icon4py, but the `l` was dropped from the name

scripts/python/run_serialization.py
jcanton: should be this one only, auto-merge not properly done...

model/testing/src/icon4py/model/testing/fixtures/benchmark.py
jcanton: up for discussion, but this is not parameterized with the `experiment` fixture and just used the defaults, lower down in `metrics_field_source` you see the same with `VerticalGridConfig`.
essentially I just moved defaults from here to Config.
should I do the same for VerticalGridConfig in the next test?

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: this and others are for some reason lists instead of single values:
```
        "damp_height": [
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0,
            45000.0
        ],
        "rayleigh_type": 2,
        "rayleigh_coeff": [
            0.1,
            0.1,
            0.1,
            0.1,
            0.1,
            0.1,
            0.1,
            0.1,
            0.1,
            0.1
        ],
```

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: I de-inlined `rayleigh_damping_height` instead, like I've done for all other lists

.gitignore
jcanton: @havogt didn't like it when I did, but now we have the symlink to `CLAUDE.md` so I'm not sure what's the policy on that...

model/testing/src/icon4py/model/testing/definitions.py
jcanton: the thought was to keep only one fixture: `experiment`, because this is as clean as can be.
then experiment contains: description, config and possibly in the future data (so we don't need the serialization fixtures).
should this be part of the `experiment` *fixture* only or also part of the `Experiment` *class* as I've done now? also up for discussion
`process_props` is needed to know which tarball/directory to read the config (and data) from

bindings/tests/bindings/test_diffusion_wrapper.py
jcanton: I guess we cannot pass `cfg` to `diffusion_init` because that needs to be called from fortran

I kept the list in that way to have a smaller diff for this PR, but in principle I agree with you.
I don't have a strong opinion about it, but I liked that at the top of the test you translate from icon4py names to fortran names which are used in the wrappers, kind of neat.

model/atmosphere/dycore/tests/dycore/integration_tests/test_solve_nonhydro.py
```diff
@@ -963,7 +963,7 @@ def test_non_hydrostatic_params(savepoint_nonhydro_init):
 
 @pytest.mark.embedded_remap_error
 @pytest.mark.datatest
-@pytest.mark.parametrize("at_initial_timestep", [(True)])
+@pytest.mark.parametrize("at_initial_timestep", [True])
```
nfarabullini, line 971: what's the difference between `at_initial_timestep` and `at_first_timestep` ? Can we just not use the latter?

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -212,6 +212,10 @@ def create_experiment_configuration(
     )
 
     # *** VerticalGridConfig ***
+    rayleigh_damping_height = nonhydrostatic_nml["damp_height"][0]
+    if isinstance(nonhydrostatic_nml["damp_height"], list):
+        rayleigh_damping_height = rayleigh_damping_height[0]
```
nfarabullini:
```
nonhydrostatic_nml["damp_height"][0]
```
implies that `nonhydrostatic_nml["damp_height"]` is a list, which would make the if statement below always true, or am I reading this wrong?

Shouldn't this be:
```
    rayleigh_damping_height = nonhydrostatic_nml["damp_height"][0]
    if isinstance(nonhydrostatic_nml["damp_height"][0], list):
        rayleigh_damping_height = rayleigh_damping_height[0]
```
or something similar?

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/common/tests/common/grid/unit_tests/test_geometry.py
havogt, line 162: this is not a unit test

scripts/python/validation_grid_utils.py
```diff
@@ -54,9 +64,18 @@ def cache_key() -> None:
...
+
+    d = "_".join(exp.name + str(exp.version) for exp in get_validation_experiments())
+    hexdigest = hashlib.md5(d.encode()).hexdigest()
+    print(hexdigest)
+
+
 @cli.command(name="download")
 def download_validation_grids() -> None:
-    """Effectively download the validation grid files."""
+    """Download the validation grid files."""
```
msimberg: I like this one!

model/testing/src/icon4py/model/testing/fixtures/benchmark.py
msimberg: Sounds like a bigger change to unify fixtures for benchmarks and other tests... It probably should eventually use the `experiment`/`experiment_description` fixture, but not now...

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -96,3 +111,219 @@ def create_icon_serial_data_provider(
...
+        root_url = definitions.TESTDATA_ROOT_URL
+        archive_filename = get_experiment_archive_filename(experiment_description, comm_size)
+        archive_path = definitions.EXPERIMENT_DATA_DIR + "/" + archive_filename
+        uri = get_experiment_archive_url(root_url, archive_path)
+        destination_path = get_datapath_for_experiment(experiment_description, processor_props)
+        data_handling.download_test_data(destination_path.parent, uri)
+    except KeyError as err:
+        raise RuntimeError(
+            f"No data for communicator of size {comm_size} exists, check on the server"
+        ) from err
```
msimberg: I just realized this branch might be wrong. I think `KeyError` might have come from when we had a map from comm size to download URLs. I can try to remove this separately. We should likely just not catch any exceptions here and let it bubble up, or not catch specific exceptions.

.github/workflows/icon4py-test-model.yml
```diff
@@ -60,6 +60,25 @@ jobs:
...
+          path: testdata/ser_icondata
+          key: ${{ steps.experiment-cache-key.outputs.key }}
+
+      - name: download experiment data
+        if: steps.experiment-cache.outputs.cache-hit != 'true'
+        env:
+          ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
+        run: |
+          ./scripts/python/validation_grid_utils.py download-experiments
+
```
msimberg: Discussed offline already before implementing, but if it takes ~2 minutes per job to download this cache I think that's acceptable, for now... The cleaner solution is to separate the json configs out from the tarballs so that they can be downloaded without all the serialized data. I think that belongs in a separate PR so that we can unlock e.g. @yiluchen1066 on #1161 to use the json config files.

msimberg (approved):

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: good catch, I'll remove the exception in  this PR

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -60,16 +60,16 @@ class SingleMomentSixClassIconGraupelConfig:
 
     #: liquid auto conversion mode. Originally defined as isnow_n0temp (PARAMETER) in gscp_data.f90 in ICON. I keep it because I think the choice depends on resolution.
     liquid_autoconversion_option: mphys_options.LiquidAutoConversionType = (
-        mphys_options.LiquidAutoConversionType.KESSLER
+        mphys_options.LiquidAutoConversionType.SEIFERT_BEHENG
```
msimberg, line 63: I missed this before and it brings me back to the question about whether any default makes sense. Doesn't have to be resolved now, but the simple question for now is why was this changed?

msimberg (approved):

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
jcanton: good that you left a comment here, I only privately wrote to @OngChia 
this parameter is not in the namelists: it's hardcoded in the source code apparently (and probably the description above it isn't even totally correct, talking about snow)
I changed the default value from KESSLER to SEIFERT_BEHENG because the class is frozen, this is not in the namelists, and the only place where it's used (the graupel test) used SEIFERT_BEHENG hardcoded explicitly.
If that's not ok we need to think of a good way to change this / make it properly selectable

jcanton: https://github.com/C2SM/icon4py/pull/1263/changes#diff-a1f0d46df11e9457f92c5815c5f558e509c49905366f65f5897a4300d08ac691L102

PR #1266, opened 2026-05-13
Fix gtfn codegen for `_compute_rayleigh_w` by using int literals

yiluchen1066:
The `gtfn` backend does not lower `IntEnum` members referenced inside a `@gtx.field_operator` body — it emits the attribute path verbatim into the generated C++, which fails to compile: 

```
error: use of undeclared identifier 'RayleighType'; did you mean 'rayleigh_type'?
error: member reference base type 'int' is not a structure or union
```

Replace `RayleighType.CLASSIC` and `RayleighType.KLEMP` with bare integer literals. Ans those values stay in sync with `constants.RayleighType`

Regression introduced by PR#1201.


yiluchen1066:
cscs-ci run default

yiluchen1066:
cscs-ci run distributed

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -214,13 +213,13 @@ def _compute_rayleigh_w(
...
     z_sin_diff = maximum(0.0, vct_a - damping_height)
     z_tanh_diff = vct_a_1 - vct_a  # vct_a(1) - vct_a
-    if rayleigh_type == RayleighType.CLASSIC:
+    if rayleigh_type == 1:  # RayleighType.CLASSIC
         rayleigh_w = (
             rayleigh_coeff
             * (sin(pi_const / 2.0 * z_sin_diff / maximum(0.001, vct_a_1 - damping_height))) ** 2
         )
 
-    elif rayleigh_type == RayleighType.KLEMP:
```
msimberg, line 223: Does this work instead:
```suggestion
    if rayleigh_type == RayleighType.CLASSIC.value:
        rayleigh_w = (
            rayleigh_coeff
            * (sin(pi_const / 2.0 * z_sin_diff / maximum(0.001, vct_a_1 - damping_height))) ** 2
        )

    elif rayleigh_type == RayleighType.KLEMP.value:
```
?

yiluchen1066: This also worked! thanks! this looks better

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -214,13 +214,13 @@ def _compute_rayleigh_w(
...
     z_tanh_diff = vct_a_1 - vct_a  # vct_a(1) - vct_a
-    if rayleigh_type == RayleighType.CLASSIC:
+    if rayleigh_type == RayleighType.CLASSIC.value:
         rayleigh_w = (
             rayleigh_coeff
             * (sin(pi_const / 2.0 * z_sin_diff / maximum(0.001, vct_a_1 - damping_height))) ** 2
         )
 
-    elif rayleigh_type == RayleighType.KLEMP:
+    elif rayleigh_type ==RayleighType.KLEMP.value:
```
msimberg:
```suggestion
    elif rayleigh_type == RayleighType.KLEMP.value:
```

msimberg (approved):
Looks good to me, modulo the formatting issue.

model/common/src/icon4py/model/common/metrics/metric_fields.py
yiluchen1066: Ohhhh, previously I only tried RayleighType.CLASSIC.value locally. But actually it breaks more than the original.

  ```.value ``` is an attribute access on the enum object. Inside a ```@gtx.field_operator```, GT4Py translates the body to its foast IR for any non-embedded backend, and foast can't
  apply .value to a captured enum object:
```
  MissingAttributeError: Object does not have attribute 'value'
    if rayleigh_type == RayleighType.CLASSIC.value:
```

  It only looks fine locally because the embedded backend (the --backend default) runs the operator as plain Python — that's why CI's diffusion tests still errored. So:

  - RayleighType.CLASSIC → only gtfn C++ codegen fails
  - .value → foast fails → breaks roundtrip and gtfn
  - 1 / 2 → works everywhere

  GT4Py can't constant-fold enum attribute access in field operators today, so I went with bare literals + a NOTE comment pinning them to constants.RayleighType.

msimberg: Sad :( thanks for checking! Then we stick with the other version for now...

msimberg (approved):

yiluchen1066:
cscs-ci run default

yiluchen1066:
cscs-ci run distributed

yiluchen1066:
cscs-ci run default

yiluchen1066:
cscs-ci run distributed

PR #1272, opened 2026-05-19
Optimize `compute_zdiff_gradp`

msimberg:
This PR makes the `compute_zdiff_gradp` metrics kernel fast on GPU by replacing the per-element host access in nested Python loops with a single batched `searchsorted` over all edge rows. The implementation uses the array namespace (numpy/cupy), so the same code runs on CPU and GPU without backend dispatch.

`compute_zdiff_gradp` now takes a precomputed `z_me` edge-level field and returns `zdiff_gradp` and `vertoffset_gradp`. It reverses each cell column to ascending order, masks levels above `flat_idx`, and offsets every row by a float64 constant large enough to keep cross-row queries separate. One flattened `searchsorted` call finds the bracket index for all `(edge, level)` queries at once; a clip to `[flat_idx, nlev-1]` gives the same `jk_start` as the sequential reference loop.

The change relies on two grid invariants documented in the module docstring: `z_ifc` is strictly decreasing in the vertical (enforced by the grid builder), and `z_me` is non-increasing per edge. Under those invariants the searchsorted result is non-decreasing in `jk` and matches the reference first-match scan. Synthetic probe inputs that violate the invariants are marked `xfail` with the reason recorded in the test file.

`z_me` is extracted into its own registered `ProgramFieldProvider` with `do_exchange=True`, so `compute_wgtfac_e`, `compute_flat_max_idx`, and `compute_zdiff_gradp` consume a single exchanged `z_me` instead of computing it inline. `compute_wgtfac_e` now accepts `z_me` directly and no longer performs an exchange. The field providers in `states/factory.py` gained an optional `exchange_fields` filter so the `compute_zdiff_gradp` provider can exchange only `zdiff_gradp` and not its inputs.

Tests in `model/common/tests/common/metrics/unit_tests/test_compute_zdiff_gradp.py` were updated: the existing datatest now runs against the new signature, a deterministic element-by-element reference mirrors the main-branch semantics, random small and full-scale cases compare the vectorized result against that reference, and probe inputs exercise the documented invariants.

Benchmark at full scale (491520 edges x 60 levels, GH200, minimum over repetitions):

| backend | time |
|---|---|
| cupy (this PR) | 21.3 ms |
| numpy (this PR) | 4.6 s |
| main (cupy, per-element host loop) | minutes |

🤖 *Opened by an agent on behalf of @msimberg*.


msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default;MODEL_SUBPACKAGES=common:driver;SESSIONS=model;MODEL_SUBSETS=datatest
🤖 *Posted by an agent on behalf of @msimberg*.

msimberg:
cscs-ci run default;MODEL_SUBPACKAGES=common:driver;SESSIONS=model;MODEL_SUBSETS=datatest;LEVELS=integration
🤖 *Posted by an agent on behalf of @msimberg*.

PR #1275, opened 2026-05-19
Py2fgen: fix fortran types, generate header only on demand

havogt:
## Main changes

**Boolean ABI.** `BOOL` is now `logical(kind=c_bool)` (Fortran) ↔ `unsigned char` (C, 1-byte) ↔ `np.bool_` (Python). The previous `logical(c_int)` ↔ `int` ↔ `np.int32` chain was not standard-conforming and round-tripped boolean arrays through a copy, making them effectively read-only; after this PR boolean arrays are a direct writable view. The C side is `unsigned char` rather than `_Bool` because NVHPC writes Fortran `.TRUE.` as byte `0xFF` and cffi's `_Bool` validator rejects that.

  **Caller action required:** declare boolean dummy args as `logical(kind=c_bool)` (not the default `logical`); convert from default `logical` at the call site via `LOGICAL(x, c_bool)`.

**Generated Fortran array dummies gain `contiguous` and `intent(inout)`.**

  **Caller awareness:** passing a non-simply-contiguous section (e.g. middle-scalar slices like `a(:,1,:)`) to a `contiguous` dummy causes some compilers (notably NVHPC) to synthesize a host-side copy, which breaks `host_data use_device` on device-resident arrays. Pass a simply-contiguous reference instead (trailing-scalar slices, or a contiguous-pointer rank-remap).

**Header generation on demand.** The default `py2fgen` CLI no longer writes `.h` alongside the compiled `.so`. The icon4py-internal `bindings/all_bindings.py` driver gained an `--output-h` flag for callers that need the header.


havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

havogt:
cscs-ci run default

havogt:
## Code Review (xhigh, multi-agent)

Generated by `/code-review xhigh` — 9 finder angles → dedup → 1-vote 3-state verify → sweep. 16 raw → 13 deduped → 10 survived verification + 5 from sweep. Mix of correctness, reuse/simplification/efficiency, and altitude findings; ranked correctness-first.

### Correctness (3)

1. **`_codegen.py:178` (Python wrapper template) — string comparison against `MemorySpace` works only by `StrEnum` accident.**
   `{% if arg.memory_space == "host" %}False{% else %}on_gpu{% endif %}` compares the enum value to the raw string `"host"`. If `_definitions.MemorySpace` is ever switched from `eve.StrEnum` to a plain `enum.Enum`, the test silently becomes always-False and every HOST array starts receiving `on_gpu` from the caller — `as_array` then dispatches to cupy for host buffers. *Fix:* compare against `_definitions.MemorySpace.HOST` via the Jinja template's `_definitions=` plumbing.

2. **`tools/tests/tools/py2fgen/test_generator.py:~136` — degenerate substring assertion in `test_render_c_source_embeds_python_wrapper_and_header`.**
   `assert "square_from_function" in sources.c` is strictly weaker than `assert "square_from_function_wrapper" in sources.c` (one is a substring of the other). The deleted `test_render.py` originally anchored the same logic with `'#include "custom_header.h"' in sources.c` to verify the header is **embedded** (not `#include`-ed). That anchor was dropped when the test moved. *Fix:* restore an embedded-header assertion (or check absence of `#include "*.h"` in `sources.c`).

3. **`tools/tests/tools/py2fgen/fortran_samples/test_bool.f90:26` — `call exit(1)` is a libgfortran extension, not standard Fortran.**
   `compile_fortran_code` already passes `-cpp` and doesn't add strict flags today, so this passes on gfortran. Add `-std=f2008` or run under NVHPC strict mode and the program won't link `exit`, silently dropping the bool round-trip coverage that the rest of this PR exists to guard. *Fix:* use `stop 1` (standard) or `error stop 1`.

### Reuse (3)

4. **`_codegen.py:290` and `:356` — `on_gpu` Fortran type hardcoded as `"logical(c_bool), value :: on_gpu"` / `"logical(c_bool) :: on_gpu"`.**
   Every other parameter in the same generator goes through `to_iso_c_type(_definitions.BOOL)`; the C-header side already routes via `to_c_type(_definitions.BOOL)` (line 244 of this PR). The two hardcoded Fortran strings sat as `logical(c_int)` until this PR had to flip them by hand; the next BOOL kind change will need a third manual flip if they stay literal. *Fix:* `f"{to_iso_c_type(_definitions.BOOL)}, value :: on_gpu"`.

5. **`_conversion.py:30` — `C_STR_TYPE_TO_NP_DTYPE` duplicates the C-type vocabulary already in `_codegen.BUILTIN_TO_CPP_TYPE`.**
   Both tables list `unsigned char`, `int`, `long`, `float`, `double` independently; the BOOL ABI migration in this PR had to update both in lock-step. A single inverse map (e.g. `{cpp: to_np_dtype(kind) for kind, cpp in BUILTIN_TO_CPP_TYPE.items()}`) eliminates that drift. *Fix:* derive `C_STR_TYPE_TO_NP_DTYPE` from `BUILTIN_TO_CPP_TYPE` + `test_utils.to_np_dtype`.

6. **`_conversion.py:65` — duplicated dtype-resolution block between `_unpack_numpy` and `_unpack_cupy`.**
   ```py
   c_type = ffi.getctype(ffi.typeof(ptr).item)
   dtype = C_STR_TYPE_TO_NP_DTYPE.get(c_type)
   if dtype is None:
       raise ValueError(f"Unsupported C data type: {c_type}")
   ```
   appears verbatim in both unpackers. *Fix:* extract a `_resolve_dtype(ffi, ptr)` helper.

### Simplification (4)

7. **`_conversion.py:152` — `_as_array_mapping()` factory is vestigial after the `dtype` parameter was dropped.**
   No closed-over state remains; the factory just returns an `lru_cache`-wrapped `impl`. *Fix:* inline the cached function directly at the single call site in `default_mapping`.

8. **`_cli.py:89` — `c_source_exists` / `shared_lib_exists` computed unconditionally before the `if skip_compilation` branch.**
   Each is consumed in only one branch; hoisting them adds an unused stat in each path and blurs which value belongs to which branch. *Fix:* move each `.exists()` inside its branch.

9. **`test_utils.py:46` — `from_np_dtype` is effectively private after this PR.**
   All external call sites in `test_test_utils.py` / `test_conversion.py` were removed alongside the `dtype` arg of `as_array`. *Fix:* rename to `_from_np_dtype` or inline into `array_to_array_info`.

10. **`_conversion.py:161` — `_int_to_bool(value, ffi) -> bool` is just `return bool(value)`.**
    The `ffi` parameter exists only to satisfy `MapperType`. *Fix:* a `lambda value, ffi: bool(value)` registered in `default_mapping`, or omit the dedicated function entirely.

### Altitude (4 + 1 plausible)

11. **`_generator.py:67` — `render(plugin, *, rpath="")` plumbs `rpath` into `extra_link_args`, but `render` only emits C source; `extra_link_args` is consumed only by `compile()`.**
    A caller passing `render(plugin, rpath="/opt/lib")` gets `.c` source with no rpath baked in (it's not encodable in C source). The parameter is dead in `render`'s only code path; it's a leftover from the deleted `_render.py`. *Fix:* drop the `rpath` parameter from `render`, keep it only on `compile`.

12. **`_conversion.py:35` — the `BOOL <-> unsigned char` correspondence lives as three independent string literals.**
    Forward map in `_codegen.BUILTIN_TO_CPP_TYPE`, reverse map in `_conversion.C_STR_TYPE_TO_NP_DTYPE`, plus the comment. No enforced relationship between forward and reverse. *Fix:* a single `BOOL_C_TYPE = "unsigned char"` constant (or derive one map from the other) so a future rename to `uint8_t` is a one-line change.

13. **`_cli.py:53` — the standalone `py2fgen` CLI lost the ability to emit `.h`.**
    Only `bindings/all_bindings.py` (the project-internal entrypoint) was updated with an `--output-h` flag; the public CLI no longer writes a header when `--skip-compilation` is passed. PR description says headers are produced "on demand" but the public CLI has no flag to request one. *Fix:* add `--output-h` to the CLI in symmetry with the bindings entrypoint.

14. **`_generator.py:115` — `generate_and_compile_cffi_plugin`'s docstring still claims it creates `lib{library_name}.so`** but is now silent about the deliberate **removal** of the `.h` side effect. The `test_compile_and_run_cffi_plugin_from_C` test in this PR had to start writing the header manually for this reason. *Fix:* update the docstring to call out the dropped header artifact, or restore it under a flag.

15. **(PLAUSIBLE) `_codegen.py:244` — `on_gpu` flag is `unsigned char` on the C side but is forwarded into `ArrayInfo` tuples as a Python int, not a bool.**
    User-facing BOOL scalars get `_int_to_bool`; `on_gpu` bypasses that path and rides through every `ArrayInfo` as `arg[2]: int`. The `ArrayInfo` type alias advertises `on_gpu: bool`; user code doing `if array_info[2] is True` (vs plain `if array_info[2]`) silently misbehaves. *Fix:* route `on_gpu` through the same conversion as a regular BOOL scalar, or convert in the wrapper template.

---

*Run cost: 23 agents, ~1.56M tokens, 22 min.*

havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

havogt:
cscs-ci run default

havogt:
cscs-ci run distributed

egparedes (approved):
I've only looked at the python changes and the PR description, and both look correct. It'd be better for a thorough review to split the changes into several smaller PRs, but it's up to you to do the extra work. I'm approving the PR preliminary here and you decide the best way to merge it.

PR #1277, opened 2026-05-20
Add pypi release workflow and instructions

msimberg:
Adds workflow to release packages on pypi (and test.pypi.org). Adds a release procedure under docs meant to be followed by whoever is making a release.


scripts/python/discover_packages.py
```diff
@@ -0,0 +1,150 @@
...
+    for dep in deps:
+        base = (
+            dep.split(">=")[0]
+            .split("~=")[0]
+            .split("==")[0]
+            .split("<")[0]
+            .split(">")[0]
+            .split("[")[0]
+            .strip()
+        )
```
egparedes: Maybe is overkill here, but using `pypackaging` (https://packaging.pypa.io/en/stable/index.html) is normally the safer way to parse version and constraint specifiers

scripts/python/discover_packages.py
```diff
@@ -0,0 +1,150 @@
...
+    queue = sorted(node for node in all_nodes if in_degree[node] == 0)
+    result: list[str] = []
+    while queue:
+        node = queue.pop(0)
+        result.append(node)
+        for dependent in sorted(reverse_graph.get(node, [])):
+            in_degree[dependent] -= 1
+            if in_degree[dependent] == 0:
+                queue.append(dependent)
+                queue.sort()
```
egparedes: In the same spirit as the previous comment, the standard library already implements this (https://devdocs.io/python~3.12/library/graphlib#graphlib.TopologicalSorter) but it might be a bit overkill here

egparedes (approved):
LGTM. I only have a couple of comments which might improve the quality of the script, but the proposals  might be overkill here, so I'll leave it to you to decide.

msimberg:
Thanks @egparedes, very good comments, I've applied both. I'll merge this as well without expensive CSCS CI since none of the main projects were changed.

PR #1282, opened 2026-05-26
Remove hardcoded numpy use from `grid_refinement.py` and `metric_fields.py`

msimberg:


model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -941,8 +940,8 @@ def compute_exner_w_implicit_weight_parameter(
 
     for jk in range(k_start, nlev):
         zdiff2_sliced = zdiff2[horizontal_start_cell:, jk]
-        index_for_k = np.where(zdiff2_sliced < 0.6)[0]
-        max_value_k = np.maximum(
+        index_for_k = array_ns.nonzero(zdiff2_sliced < 0.6)[0]
```
nfarabullini, line 943: We should actually change all `array_ns.where()` to `arrays_ns.nonzero()` if only the condition parameter is specified. Unless you want to make these edits here, I'll open a new issue

nfarabullini (approved):

model/common/src/icon4py/model/common/metrics/metric_fields.py
msimberg: It's a good point! I'll make a separate PR fixing that for this case and other instances, if any.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

model/common/src/icon4py/model/common/metrics/metric_fields.py
msimberg: https://github.com/C2SM/icon4py/pull/1286

PR #1281, opened 2026-05-22
Remove old driver

msimberg:
#1376 added tests to bring the standalone_driver testing on par with the old driver, which means there's nothing blocking removing the old driver anymore. The standalone driver rename is in #1283.

PR #1283, opened 2026-05-26
Rename standalone driver to driver

msimberg:
Depends on #1281.


msimberg:
cscs-ci run default

model/driver/README.md
```diff
@@ -24,7 +24,7 @@ export GRID_FOLDER=<path to the folder holding grids>
 export CONFIG_FOLDER=<path to the configuration directory>
 
 # command line arguments
-icon4py-standalone-driver \
+icon4py-driver \
```
msimberg, line 27: Bikeshedding-worthy.

jcanton: I like

msimberg: I suppose I could've expanded a bit... I do like `icon4py-driver`. I also think plain `icon4py` is a good option. Things like `icon4py-app` or similar are not too bad either, but I think for that the package should also be renamed. And driver is likely the most familiar name to the community so I wouldn't change it to something more generic without scientists also liking the name (though they can have strange opinions as well...).

So the question I was trying to pose here is: should it be called something other than `icon4py-driver`?

jcanton: I had understood exactly (because we had talked about it?) and my very verbose comment was a vote in favour of the `icon4py-driver` name (at least for now, we can re-think it but only alongside a package name as you wrote).
and if nobody else votes that's the name that wins ;-)
joke aside, it was also the name given to the old driver, so maybe somebody already bike-shedded on it months ago (or not)
@OngChia @yiluchen1066 @egparedes @havogt ?

DropD: I would keep it for now. I'm not sure we don't want to use `icon4py` as a base command for more than just running sims.

Plus, we can change it in the future by creating an alias and stay backwards compatible.

msimberg:
cscs-ci run default

jcanton (approved):
since nobody complains...

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
Merging https://github.com/C2SM/icon4py/pull/1426 first for the GHEX/nanobind issue that showed up on this PR (completely unrelated to the changes here).

PR #1285, opened 2026-05-27
Update development status from alpha to beta

msimberg:
I would no longer consider icon4py alpha software. This updates the status to beta for all packages, except the driver and standalone_driver packages which stay at alpha. See https://pypi.org/classifiers/ for available classifiers.


nfarabullini (approved):
I would say that dycore and diffusion can be classified as production stable. I don't have particular feelings about standalone driver, as I feel like it's in between alpha and beta

msimberg:
I changed driver and standalone_driver back to alpha (driver will anyway be deleted soon: #1281).

msimberg:
Merging without CI since this is just metadata changes.

PR #1286, opened 2026-05-27
Replace one-argument wheres with nonzero

msimberg:
As pointed out by @nfarabullini in https://github.com/C2SM/icon4py/pull/1282#discussion_r3302974984, `array_ns.where` with only one argument is exactly `array_ns.nonzero`. The documentation states (https://numpy.org/doc/stable/reference/generated/numpy.where.html):
> When only condition is provided, this function is a shorthand for np.asarray(condition).nonzero(). Using [nonzero](https://numpy.org/doc/stable/reference/generated/numpy.nonzero.html#numpy.nonzero) directly should be preferred, as it behaves correctly for subclasses. The rest of this documentation covers only the case where all three arguments are provided.

### DIFF
--- a/model/common/src/icon4py/model/common/grid/grid_refinement.py
+++ b/model/common/src/icon4py/model/common/grid/grid_refinement.py
@@ -8,7 +8,6 @@
 import logging
 from typing import Final
 
-import numpy as np
 from gt4py import next as gtx
 
 from icon4py.model.common import dimension as dims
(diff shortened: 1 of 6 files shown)


nfarabullini:
cscs-ci run distributed

nfarabullini:
cscs-ci run default

nfarabullini (approved):

PR #1291, opened 2026-05-28
Test `--tach` in CI

msimberg:
`--tach` invokes tach as pytest plugin which uses the tach import/dependency graph to figure out which tests are reachable from the files changed in a given PR, allowing to skip unaffected tests, therefore making CI faster.


jcanton:
closing this for now because it turned out it's not as easy as we thought: tach ignores too much because of the monorepo structure, and other options require investigation

PR #1292, opened 2026-05-28
Use common base image for default and distributed CI pipelines

msimberg:
Combines the two base images into one. The ubuntu base image is taken from the distributed pipeline with 25.10. NVHPC is still needed for GPU compilation tests, so NVHPC is also used. The NVHPC version is unchanged.

Also adds cuda13 extra to icon4py-bindings.


msimberg:
cscs-cin run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

havogt:
- [ ] Reviewing this "I" noticed that at some point we seem to have lost compilation of the GPU bindings tests that were exercised with nvfortran and enabled in a very weird way which I never cleaned up: `compiler=os.environ["NVFORTRAN_COMPILER"]` in `test_cli.py:230`. Not sure if this alone is a good reason to keep the nvhpc toolkit around...
- [ ] I believe `PY2F_GPU_TESTS: 1` is dead because the trigger is the above env var currently.
- [ ] Any reason to go to 25.10 and not on a LTS release?

msimberg:
cscs-ci run default

msimberg:
> * [ ]  Reviewing this "I" noticed that at some point we seem to have lost compilation of the GPU bindings tests that were exercised with nvfortran and enabled in a very weird way which I never cleaned up: `compiler=os.environ["NVFORTRAN_COMPILER"]` in `test_cli.py:230`. Not sure if this alone is a good reason to keep the nvhpc toolkit around...
> 
>     * [ ]  I believe `PY2F_GPU_TESTS: 1` is dead because the trigger is the above env var currently.

Argh, ok. Yes, I would prefer not to lose that, I did not realize that was implicitly enabled like that. I'll see if I can reenable that. If it needs to use NVHPC instead, I'll use that instead for both pipelines, not really a problem.

>     * [ ]  Any reason to go to 25.10 and not on a LTS release?

25.10 (specifically I think pmix) was not compatible with pmix that we have on the system when I set this up in #692. It's not intended to stay like that and now I can try LTS. I will look at that separately though.

havogt:
2 more comments from my friend:
  - NUM_PROCESSES: 0 (commit 8957585a6) is inert for the distributed job. NUM_PROCESSES is only consumed in noxfile.py (pytest -n {NUM_PROCESSES}), but .test_distributed_aarch64 invokes pytest directly via ci-mpi-wrapper.sh with no -n. So the distributed command never enabled xdist in the first place, and this variable has no effect today. Harmless, but it's not load-bearing — if the goal was "no xdist under MPI," that's already true by construction. Worth a comment so nobody treats it as the thing protecting MPI from xdist.
  - PYVERSION_PREFIX (base.yml:11) is now a dead variable. It was only ever used in BASE_IMAGE_${PYVERSION_PREFIX}, which this PR replaced with plain BASE_IMAGE. Nothing references it anymore — a one-line cleanup.

msimberg:
> 2 more comments from my friend:
> 
>     * NUM_PROCESSES: 0 (commit [8957585](https://github.com/C2SM/icon4py/commit/8957585a65b88653a6e492eab245ccd7ec8fdb84)) is inert for the distributed job. NUM_PROCESSES is only consumed in noxfile.py (pytest -n {NUM_PROCESSES}), but .test_distributed_aarch64 invokes pytest directly via ci-mpi-wrapper.sh with no -n. So the distributed command never enabled xdist in the first place, and this variable has no effect today. Harmless, but it's not load-bearing — if the goal was "no xdist under MPI," that's already true by construction. Worth a comment so nobody treats it as the thing protecting MPI from xdist.

I'll leave the NUM_PROCESSES=0 just to explicitly override it. I will work on making the mpi tests run in a nox session as well after this.

>     * PYVERSION_PREFIX (base.yml:11) is now a dead variable. It was only ever used in BASE_IMAGE_${PYVERSION_PREFIX}, which this PR replaced with plain BASE_IMAGE. Nothing references it anymore — a one-line cleanup.

Good catch, this can go.

msimberg:
cscs-ci run default

msimberg:
It looks like the `test_py2fgen_compilation_and_execution_gpu` test was completely skipped (both github actions and CSCS CI; the former doesn't PY2F_GPU_TESTS, the latter doesn't run unit tests). I'm adding the tools_and_bindings unittests session to CSCS CI (I think we need to run it somewhere...) and for now it will be unconditionally run. With #1294 we'll be able to select if it runs or not.

I'll go back to the nvhpc install as well so that the test can actually run correctly.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run benchmark-bencher

msimberg:
@havogt did the py2fgen gpu test ever pass in CI? It's failing now that it's been enabled, I don't know if there's something obvious in the output to you: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14737928408?

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run benchmark-bencher

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run benchmark-bencher

bindings/pyproject.toml
```diff
@@ -50,6 +50,7 @@ version = "0.2.0rc2"
 [project.optional-dependencies]
 cuda11 = ['cupy-cuda11x>=13.0', 'gt4py[cuda11]']
 cuda12 = ['cupy-cuda12x>=13.0', 'gt4py[cuda12]']
+cuda13 = ['cupy-cuda13x>=14.0', 'gt4py[cuda13]']
```
havogt, line 53:
```suggestion
cuda12 = ['gt4py[cuda12]']
cuda13 = ['gt4py[cuda13]']
```
I have it in #1299 as well

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run benchmark-bencher

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run benchmark-bencher

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run benchmark-bencher

msimberg:
There are now no GPU benchmark alerts (after I downgraded NVHPC back to 24.11 as it is on main), except one that seems to be noise (it passed after a rerun and the history seems to have similar outliers).

A few of the default/distributed tests failed with cache/compilation-related issues that also went away after reruns. I'll need to keep an eye on them if they're recurring.

msimberg:
Waiting for final reviews before triggering CI again.

ci/docker/base.Dockerfile
```diff
@@ -60,41 +83,98 @@ RUN wget -q ${HPC_SDK_URL} -O /tmp/nvhpc.tar.gz && \
...
-    NVIDIA_VISIBLE_DEVICES=all \
-    NVIDIA_DRIVER_CAPABILITIES=compute,utility
+ENV CUDA_PATH=${HPC_SDK_PATH}/cuda
 
-ENV PATH=${HPC_SDK_PATH}/compilers/bin:${HPC_SDK_PATH}/comm_libs/mpi/bin:${PATH} \
+ENV PATH=${HPC_SDK_PATH}/compilers/bin:${PATH} \
     MANPATH=${HPC_SDK_PATH}/compilers/man:${MANPATH} \
-    LD_LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LD_LIBRARY_PATH}
+    LD_LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LD_LIBRARY_PATH} \
+    LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LIBRARY_PATH}
```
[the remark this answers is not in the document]
msimberg: This is not a problem. Nvidia binaries and libraries are injected from the host system at runtime. nvidia-cuda-mps-control is one of them, injected into `/usr/bin/nvidia-cuda-mps-control`, so available in path.

havogt:
There is so much spam here now. What happened to your question https://github.com/C2SM/icon4py/pull/1292#issuecomment-4646500518? Did you disable that again? It was working at some point in the past, but not sure when it got dropped.

ci/default.yml
```diff
@@ -28,8 +28,11 @@ test_model_stencils_aarch64:
 .test_tools_datatests:
```
msimberg: I should rename this... it's not only datatests now, but also unit tests.

msimberg:
> There is so much spam here now. What happened to your question [#1292 (comment)](https://github.com/C2SM/icon4py/pull/1292#issuecomment-4646500518)? Did you disable that again? It was working at some point in the past, but not sure when it got dropped.

The test is now enabled and passing (in https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14761646007#L465).

The `test_tools_datatests` job in CSCS CI was running only datatests (makes sense...). I expanded it to run the unittests as well, which includes the `test_py2fgen_compilation_and_execution_gpu` test. The `unittests` job ran in a few minutes, so I think it's worth it to enable that always.

The change that made it work was changing the memory space from `HOST` to `MAYBE_DEVICE` in `tools/tests/tools/py2fgen/wrappers/simple.py`, which I have no idea if it's the right fix.

ci/docker/base.Dockerfile
```diff
@@ -1,43 +1,66 @@
-FROM ubuntu:22.04
+FROM ubuntu:25.10
```
edopao, line 1: Would it  be possible to use an LTS release (26.04)?

ci/docker/base.Dockerfile
```diff
@@ -1,43 +1,66 @@
...
+        libreadline-dev \
+        libsensors-dev \
+        libsqlite3-dev \
+        libssl-dev \
+        libtool \
+        libuv1-dev \
+        libyaml-dev \
+        llvm \
+        gfortran \
+        gfortran-12 \
```
edopao: Do you need both gfortran and gfortran-12?

ci/docker/base.Dockerfile
```diff
@@ -60,41 +83,98 @@ RUN wget -q ${HPC_SDK_URL} -O /tmp/nvhpc.tar.gz && \
...
+ENV CUDA_PATH=${HPC_SDK_PATH}/cuda
 
-ENV PATH=${HPC_SDK_PATH}/compilers/bin:${HPC_SDK_PATH}/comm_libs/mpi/bin:${PATH} \
+ENV PATH=${HPC_SDK_PATH}/compilers/bin:${PATH} \
     MANPATH=${HPC_SDK_PATH}/compilers/man:${MANPATH} \
-    LD_LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LD_LIBRARY_PATH}
+    LD_LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LD_LIBRARY_PATH} \
+    LIBRARY_PATH=${CUDA_PATH}/lib64:${HPC_SDK_PATH}/math_libs/lib64:${LIBRARY_PATH}
+
+ENV NVIDIA_VISIBLE_DEVICES=all
```
edopao, line 97: Apparently`NVIDIA_VISIBLE_DEVICES=all` works on the shared partition, already in baseline. No comment, just an observation.

bindings/pyproject.toml
edopao: I see that `bindings` depends on `common`, so this proposed change could go there, so we have it in one place only.

edopao (commented):
LGTM, I like this change!

havogt:
>The change that made it work was changing the memory space from HOST to MAYBE_DEVICE in tools/tests/tools/py2fgen/wrappers/simple.py, which I have no idea if it's the right fix.

Makes sense!

ci/docker/base.Dockerfile
msimberg: Unfortunately not, at least not without further investigation. I tried it quickly but got container startup errors. I'd like to update it as soon as possible but separately from this PR. Thanks for the reminder though, I opened https://github.com/C2SM/icon4py/issues/1308 for it.

ci/docker/base.Dockerfile
msimberg: No! Good catch.
```suggestion
        gfortran-12 \
```

ci/docker/base.Dockerfile
msimberg: Yes! Keep in mind that `NVIDIA_VISIBLE_DEVICES` is not the same as `CUDA_VISIBLE_DEVICES`. I think on the shared partition there's no real isolation, it's literally just `CUDA_VISIBLE_DEVICES` doing the work.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

edopao (approved):

havogt (approved):
lgtm

ci/docker/base.Dockerfile
msimberg: It turns out I do need gfortran installed, or at least a binary called `gfortran`, because of https://github.com/C2SM/icon4py/blob/7709a6b621c87cdd87055bbae3d03dd894f6887e/tools/tests/tools/py2fgen/test_cli.py#L74. The TODO is of course relevant, but I won't deal with that here.

I'll leave gfortran-12 just to be consistent with gcc-12 and gfortran only for the test, for now...

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1287, opened 2026-05-27
Add parallel MPI test scheduler with subcommunicators for increased concurrency

msimberg:
This adds a custom pytest scheduler that replaces the MPI communicator with a subcommunicator. This allows running more than one MPI tests concurrently through pytest. The `--mpi-subcomm-size` sets the size of the subcommunicator (as opposed to srun/mpirun -n, i.e. MPI_COMM_WORLD, currently). Tests are distributed evenly across the subcommunicators.

A downside is that pytest output is uglier since one rank from each subcommunicator now prints regular pytest output (the results are not collected to global rank 0, only rank 0 of each subcommunicator).

This allows oversubscribing slurm jobs to get much higher throughput on the MPI CI test jobs.

This of course does not help e.g. with the standalone driver package because it only runs a few tests and does not benefit from oversubscription. However, e.g. `common` benefits quite a bit from this.

As an example, `test_model_distributed: [common, dace_gpu, integration]` ran in 11 minutes on this branch (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504675/-/jobs/14674985636) vs 34 minutes on another branch (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504675/-/jobs/14630893557). Warm cache in both cases.

PR #1295, opened 2026-05-29
Driver config

jcanton:
no more hardcoded config for standalone driver.
also enable all experiments that were used in the old driver for tests (IC and full timestep)


model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -85,6 +85,30 @@ class SingleMomentSixClassIconGraupelConfig:
...
 
+    @classmethod
+    def from_fortran_dict(
+        cls, atmo_dict: dict[str, Any], **overrides: Any
+    ) -> SingleMomentSixClassIconGraupelConfig:
+        run_nml = atmo_dict["run_nml"]
+
+        nwp_phy_nml = atmo_dict.get("nwp_phy_nml")
+        nwp_tuning_nml = atmo_dict.get("nwp_tuning_nml")
+        if nwp_phy_nml is not None and nwp_tuning_nml is not None:
```
msimberg, line 98: Is this likely to happen? I.e. should this be silently ignored or raise an exception instead?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -31,13 +46,113 @@ class DriverConfig:
...
+]:
+    with (config_file_path / f"{fortran_config.NAMELIST_ATM_FNAME}.json").open() as f:
+        atmo_dict = json.load(f)
+    with (config_file_path / f"{fortran_config.NAMELIST_MASTER_FNAME}.json").open() as f:
+        master_dict = json.load(f)
+
+    profiling_stats = ProfilingStats() if enable_profiling else None
+
+    interpolation_config = interpolation_factory.InterpolationConfig.from_fortran_dict(atmo_dict)
+    assert interpolation_config.max_nudging_coefficient is not None
```
msimberg: I think this belongs in InterpolationConfig `__post_init__`, with a raise instead of assert? Or is this specific to the standalone driver config which requires this to be set, but other uses don't require it to be set?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -31,13 +46,113 @@ class DriverConfig:
...
+
+    return (
+        driver_cfg,
+        vertical_grid_config,
+        diffusion_config,
+        advection_config,
+        nonhydro_config,
+        interpolation_config,
+        metrics_config,
+    )
```
msimberg: This might deserve a TODO. We discussed that if this uses `ExperimentConfig` we end up with a circular dependency. But eventually I think this should be a proper data class/config object instead of a tuple. Not necessary to change now, but I think it's worth commenting that this is not the ideal solution.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -31,13 +46,113 @@ class DriverConfig:
...
+            )
+        output_path.mkdir(parents=True, exist_ok=False)
+
+    if process_props is not None:
+        if mpi_decomp.mpi4py is not None and mpi_decomp.mpi4py.MPI.COMM_WORLD.Get_size() > 1:
+            comm = mpi_decomp.mpi4py.MPI.COMM_WORLD
+            output_path = pathlib.Path(
+                comm.bcast(str(output_path) if process_props.rank == 0 else None, root=0)
+            )
+            comm.Barrier()
```
msimberg:
```suggestion
    if process_props is not None and process_props.comm_size > 1:
        output_path = pathlib.Path(
            comm.bcast(str(output_path) if process_props.rank == 0 else None, root=0)
        )
        process_props.comm.Barrier()
```

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -179,168 +167,48 @@ def create_experiment_configuration(
...
-        thhgtd_zdiffu=nonhydrostatic_nml["thhgtd_zdiffu"],
-        rayleigh_type=constants.RayleighType(nonhydrostatic_nml["rayleigh_type"]),
-        rayleigh_coeff=rayleigh_coeff,
-        divdamp_trans_start=nonhydrostatic_nml["divdamp_trans_start"],
-        divdamp_trans_end=nonhydrostatic_nml["divdamp_trans_end"],
-        divdamp_type=nonhydrostatic_nml["divdamp_type"],
-        igradp_method=nonhydrostatic_nml["igradp_method"],
-    )
+    atmo_dict = _read_namelist_json(config_path / f"{fortran_config.NAMELIST_ATM_FNAME}.json")
+    master_dict = _read_namelist_json(config_path / f"{fortran_config.NAMELIST_MASTER_FNAME}.json")
```
msimberg: Possibly worth moving filename and path to helper functions.

msimberg (commented):
Just a few preliminary comments, but this is already a very nice direction.

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -85,6 +85,30 @@ class SingleMomentSixClassIconGraupelConfig:
...
+                ice_stickeff_min=nwp_tuning_nml["tune_zceff_min"],
+                power_law_coeff_for_ice_mean_fall_speed=nwp_tuning_nml["tune_zvz0i"],
+                exponent_for_density_factor_in_ice_sedimentation=nwp_tuning_nml["tune_icesedi_exp"],
+                power_law_coeff_for_snow_fall_speed=nwp_tuning_nml["tune_v0snow"],
+                rain_mu=nwp_phy_nml["mu_rain"],
+                rain_n0=nwp_phy_nml["rain_n0_factor"],
+                snow2graupel_riming_coeff=nwp_tuning_nml["tune_zcsg"],
+                **overrides,
+            )
+        return None  # Throw an error if trying to acces this
```
DropD: Looking at call sites for of this classmethod (and the class' constructor), it seems that `None` would typically not be an expected value, while a default constructed instance would be unproblematic.

Is there a reason why this should not simply return defaults if one of `nwp_phy_nml` or `nwp_tuning_nml` is not found? Would the fortran code fail in that case or just work with defaults?

If so, I agree with the comment, that this should raise an exception instead for returning `None`.

DropD: Comparing with other `from_fortran_dict` implementations, the usual pattern seems to be to implicitly raise a `KeyError` if some necessary config value / section is not found (by using `[]` access). That makes it confusing that here, care is taken to use `.get()` but then return `None`, which might cause a harder to trace exception later on.

model/common/src/icon4py/model/common/utils/fortran_config.py
```diff
@@ -0,0 +1,30 @@
...
+MASTER_DICT_FNAME: Final = f"{NAMELIST_MASTER_FNAME}.json"
+
+
+def list_to_value(obj: list[_T] | _T) -> _T:
+    # Some parameters are allocated as `max_dom`-sized lists, with one value
+    # per domain. ICON4Py (for now) only runs on one domain.
+    # Most parameters have the same value for all elements, others (such as
+    # num_levels) have a default value different from domain[0].
+    # Tracers are an even different case where there is one value per tracer,
+    # but with the current version of ICON4Py all tracers get the same config.
```
DropD: I would recommend adding something like the following here:

```
# TODO: stop using this for per-tracer values when enabling that functionality
```

so that whoever implements this in the future can search for TODOs related to tracers easily.

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
jcanton: I'm happy to change if you and/or @msimberg have a better idea, here's the explanation related to the comment above https://github.com/C2SM/icon4py/pull/1295/changes/BASE..f439373e68e5e63e301c0fa8a77a124eea3d3d55#r3332380835

some experiments/testcases (the ones that ran microphysics) have the two namelists `nwp_phy_nml, nwp_tuning_nml`, the others don't (they're missing from the file containing the dict). so either I make a conditional at the call site in standalone_driver/config.py (https://github.com/C2SM/icon4py/pull/1295/changes/BASE..f439373e68e5e63e301c0fa8a77a124eea3d3d55#diff-1374141d85d171d3fa2c5711b656072708d65ff1080beaf0cdeedcc64f105bd7R128) or something here so that the Config is fake-read but throws an error if you try to access it (which you never should be doing because you're running an experiment without microphysics).
I cannot throw an error if the namelists are missing, because that's normal for the experiments without microphysics.

jcanton: ah, and using the class default values would end up in silent errors difficult to debug because again, you should not be accessing this, if you end up accessing it and getting the defaults you're doing something wrong and do not notice

model/common/src/icon4py/model/common/utils/fortran_config.py
jcanton: good point

model/common/src/icon4py/model/common/grid/vertical.py
```diff
@@ -119,6 +125,27 @@ class VerticalGridConfig:
...
+            top_height_limit_for_maximal_layer_thickness=sleve_nml["htop_thcknlimit"],
+            lowest_layer_thickness=sleve_nml["min_lay_thckn"],
+            model_top_height=sleve_nml["top_height"],
+            flat_height=sleve_nml["flat_height"],
+            stretch_factor=sleve_nml["stretch_fac"],
+            rayleigh_damping_height=fortran_config.list_to_value(nonhydrostatic_nml["damp_height"]),
+            htop_moist_proc=nonhydrostatic_nml["htop_moist_proc"],
+            SLEVE_decay_scale_1=sleve_nml["decay_scale_1"],
+            SLEVE_decay_scale_2=sleve_nml["decay_scale_2"],
+            SLEVE_decay_exponent=sleve_nml["decay_exp"],
```
msimberg, line 145: Nit/question, can these become:

```suggestion
            sleve_decay_scale_1=sleve_nml["decay_scale_1"],
            sleve_decay_scale_2=sleve_nml["decay_scale_2"],
            sleve_decay_exponent=sleve_nml["decay_exp"],
```
or something else? Or is `SLEVE_*` too idiomatic to change.

model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -1837,9 +1837,15 @@ def rho_now(self):
     def vn_now(self):
         return self._get_field("vn_now", dims.EdgeDim, dims.KDim)
 
+    def w_now(self):
+        return self._get_field("w_now", dims.CellDim, dims.KDim)  # shape (ncells, nlev+1)
```
msimberg:
```suggestion
        return self._get_field("w_now", dims.CellDim, dims.KDim)
```
?

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -168,18 +168,23 @@ class ExperimentDescription:
...
     grid: GridDescription
-    version: int = 4
+    version: int = 5
 
 
 @dataclasses.dataclass
 class ExperimentConfig:
-    driver: driver_config.DriverConfig
+    # NOTE: This has a duplicate in standalone_driver/config.py to avoid circular imports.
+    file_path: pathlib.Path
```
msimberg: What is `file_path` actually referring to at this level?

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -6,6 +6,8 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+from __future__ import annotations
+
```
msimberg, line 10: Leftover?

model/common/src/icon4py/model/common/utils/fortran_config.py
```diff
@@ -0,0 +1,57 @@
...
+    Fortran→Python name translation is driven by the required ``_fortran_name_map``
+    class variable: ``{fortran_key: python_field_name}``.
+    """
+    name_map: dict[str, str] = cls._fortran_name_map  # type: ignore[attr-defined]
+    known_fields = {f.name for f in dataclasses.fields(cls)}  # type: ignore[arg-type]
+    kwargs: dict[str, Any] = {}
+    for key, value in source.items():
+        python_name = name_map.get(key, key)
+        if python_name in known_fields:
+            kwargs[python_name] = value
```
msimberg: `_fortran_name_map` should probably be `fortran_name_map` if it's exposed like this. `params_from_dict` should possibly also be a method on a base class, but...

What's not clear to me how and why `params_from_dict` is different and required compared to `from_fortran_dict`?

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
DropD: I understand now why it's like this, thanks!

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/testcases/from_file.py
```diff
@@ -0,0 +1,139 @@
...
+            state.tracer[i].ndarray[:, :] = array_ns.asarray(tracers_raw[:nc, :, i])
+
+    return state
+
+
+def read_from_file(
+    *,
+    parameters: FromFileParameters,
+    grid: icon_grid.IconGrid,
+    geometry_field_source: grid_geometry.GridGeometry,  # unused; kept for API parity
```
msimberg: I think the `# unused; kept for API parity` comments need to be turned into a protocol or similar, to make it clear why these are passed.

msimberg (commented):
Some more comments, I've not had time to look through all the changes yet.

model/common/src/icon4py/model/common/grid/vertical.py
jcanton: I don't have a strong opinion about this but I think it helps separate them from the other vertical config params, see in the config above (line 111 https://github.com/C2SM/icon4py/pull/1295/changes/BASE..63e3d28b810856e3560b310720c85b0e3c82ec98#diff-9027d878732ea8eb8c21ca240d2387993a7e354ad26d9a5179009b312a4332d8L105)

reason is SLEVE is one type of verical coordinate and there are others, that we didn't port. So in principle this could have its own config, separated from or child of the VerticalGridConfig (but not in this PR).

so long story short: they can be renamed, but maybe uppercase helps?

model/common/src/icon4py/model/common/utils/fortran_config.py
jcanton: I'm trying to solve the following problem:
`from_fortran_dict` is used on the `atm` and `master` dicts/namelist which always have all keys (with default or user selected val) because they are auto-generated as dumps.

the `input_dict` instead is the dict/namelist where the user has written some key/val pairs, but I don't know which ones a-priori.
There may be present ones, missing ones (for which using the default value is the right choice) and the same input dict contains key/vals for different icon4py Configs, and some have been renamed to another variable name :-)

help?

model/testing/src/icon4py/model/testing/definitions.py
jcanton: it's the path to the config file(s), as in `experiment.config.file_path`, how do I make it clearer? `config_files_path` / `path_to_config_files` ?

tach.toml
```diff
@@ -75,7 +75,7 @@ depends_on = [
 
 [[modules]]
 path = "icon4py.model.testing"
-depends_on = [{ path = "icon4py.model.common" }]
+depends_on = [{ path = "icon4py.model.common" }, { path = "icon4py.model.standalone_driver" }]
```
jcanton: remove this ?!!?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/testcases/from_file.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def read_edge_k(name: str) -> data_alloc.NDArray:
+        return array_ns.asarray(array_ns.squeeze(ser.read(name, sp).astype(float))[:ne, :])
+
+    state = prognostics.initialize_prognostic_state(grid=grid, allocator=backend, ntracer=ntracer)
+    state.rho.ndarray[:, :] = read_cell_k("rho_now")
+    state.exner.ndarray[:, :] = read_cell_k("exner_now")
+    state.theta_v.ndarray[:, :] = read_cell_k("theta_v_now")
+    state.vn.ndarray[:, :] = read_edge_k("vn_now")
+    state.w.ndarray[:, :] = read_cell_k("w_now")  # shape (nc, num_levels + 1)
```
jcanton:
```suggestion
    state.w.ndarray[:, :] = read_cell_k("w_now")
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/testcases/from_file.py
```diff
@@ -0,0 +1,139 @@
...
+
+    All array manipulation uses only ``numpy`` and the GT4Py field API;
+    there is intentionally no dependency on ``icon4py.model.testing``.
+    """
+    fname = f"icon_pydycore_rank{rank}"
+    ser = serialbox.Serializer(serialbox.OpenModeKind.Read, str(data_path), fname)
+    sp = ser.savepoint["prognostics"].id[1].location["initial-state"].as_savepoint()
+    log.debug("Reading prognostics initial-state from %s / %s", data_path, fname)
+
+    nc = grid.num_cells
```
jcanton:
```suggestion
    num_cells = grid.num_cells
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/testcases/from_file.py
```diff
@@ -0,0 +1,139 @@
...
+    ne = grid.num_edges
```
jcanton:
```suggestion
    num_edges = grid.num_edges
```

jcanton:
move `from_file`(s) one level up and rename testcases -> analytical

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/testcases/from_file.py
DropD: How about `typing.Annotated[grid_geometry.GridGeometry, OnlyApiParity]`, where `OnlyApiParity` is an empty class with a docstring explaining what this is all about?

tach.toml
```diff
@@ -70,7 +70,6 @@ depends_on = [
     { path = "icon4py.model.atmosphere.diffusion" },
     { path = "icon4py.model.atmosphere.dycore" },
     { path = "icon4py.model.common" },
-    { path = "icon4py.model.testing" },              # TODO(): remove this dependency when driver is fully standalone
```
nfarabullini, line 73: can we assign. someone to this TODO?

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -168,18 +168,22 @@ class ExperimentDescription:
...
     long_name: str
     grid: GridDescription
-    version: int = 4
+    version: int = 5
 
 
 @dataclasses.dataclass
 class ExperimentConfig:
-    driver: driver_config.DriverConfig
+    # NOTE: This has a duplicate in standalone_driver/config.py to avoid circular imports.
```
nfarabullini, line 176: do we still use the class in `standalone_driver/config.py` and if so, does it have a different name?

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -130,217 +144,77 @@ def download_experiment(
...
-        divdamp_order=dycore_states.DivergenceDampingOrder(nonhydrostatic_nml["divdamp_order"]),
-        divdamp_type=dycore_states.DivergenceDampingType(nonhydrostatic_nml["divdamp_type"]),
-        l_vert_nested=run_nml["lvert_nest"],
-        deepatmos_mode=nml_data["dynamics_nml"]["ldeepatmo"],
-        extra_diffu=nonhydrostatic_nml["lextra_diffu"],
-        rhotheta_offctr=nonhydrostatic_nml["rhotheta_offctr"],
-        veladv_offctr=nonhydrostatic_nml["veladv_offctr"],
+    if experiment_description == definitions.Experiments.MCH_CH_R04B09:
+        # The MCH_CH_R04B09 experiment uses an advection scheme that is not supported by ICON4Py.
+        # Hopefully you are not actually trying to access this.
```
nfarabullini:
```suggestion
```
???

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -130,217 +144,77 @@ def download_experiment(
...
+        logger.warning(
+            "Loading advection config with default values for experiment %s, as the original config is not supported by ICON4Py",
+            experiment_description.name,
+        )
+        advection_config = advection.AdvectionConfig()
+    else:
```
nfarabullini:
```suggestion
advection_config = advection.AdvectionConfig()
if experiment_description != definitions.Experiments.MCH_CH_R04B09:
```
and add log message (if necessary) in the end

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -130,217 +144,77 @@ def download_experiment(
...
-    return obj[0] if isinstance(obj, list) else obj
-
-
 def create_experiment_configuration(
     experiment_description: definitions.ExperimentDescription,
     processor_props: decomposition.ProcessProperties,
 ) -> definitions.ExperimentConfig:
-    """
-    Create configuration objects from the experiment's namelist JSON file.
+    # NOTE: This has a duplicate in standalone_driver/config.py to avoid circular imports.
```
nfarabullini, line 151: since this happens in multiple places, can't we move stuff from `standalone_driver/config.py` somewhere else?

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
nfarabullini: remove comment?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/gauss3d.py
```diff
@@ -0,0 +1,152 @@
...
+    from icon4py.model.standalone_driver import driver_states
+
+log = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class Gauss3DParameters:
+    u0: float = 0.0
+    t0: float = 300.0
+    brunt_vais: float = 0.01
```
nfarabullini, line 44: can you specify where these default values were taken from?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -0,0 +1,267 @@
...
+    p_sfc: float = 100000.0
+    baroclinic_amplitude: float = 0.0
+    u0: float = 35.0
+    temp0: float = 288.0
+    eta_0: float = 0.252
+    eta_t: float = 0.2
+    gamma: float = 0.005
+    dtemp: float = 4.8e5
+    lon_perturbation_center: float = math.pi / 9.0
+    lat_perturbation_center: float = 2.0 * math.pi / 9.0
```
nfarabullini, line 62: same here. Where did these values come from?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
```diff
@@ -0,0 +1,392 @@
...
+        )
+        quadratic_c = -(fac2 * fac3 / ddqz_z_half[:, k + 1] + fac2 * d_exner_dz_ref_ic[:, k + 1])
+
+        exner[:, k] = (
+            quadratic_b + array_ns.sqrt(quadratic_b**2 + 4.0 * quadratic_a * quadratic_c)
+        ) / (2.0 * quadratic_a)
+        theta_v[:, k] = temp_v[:, k] / exner[:, k]
+        rho[:, k] = (
+            exner[:, k] ** phy_const.CVD_O_RD * phy_const.P0REF / (phy_const.RD * theta_v[:, k])
+        )
```
nfarabullini, line 84: you don't return anything here

model/common/src/icon4py/model/common/topography/config.py
```diff
@@ -0,0 +1,104 @@
...
+                log.info("Analytical topography for Jablonowski-Williamson test case")
+                parameters = fortran_config.params_from_dict(
+                    jw_topo.JablonowskiWilliamsonParameters, testcase_nml
+                )
+            case "gauss3D":
+                log.info("Analytical topography for Gaussian hill")
+                parameters = fortran_config.params_from_dict(
+                    gausshill_topo.GaussianHillParameters, testcase_nml
+                )
+            case name:
```
nfarabullini, line 80: can you explain how you get `name` here (instead of just `_` ?

model/common/src/icon4py/model/common/topography/smoothing.py
```diff
@@ -34,19 +32,15 @@ def smooth_topography(
...
 ) -> data_alloc.NDArray:
-    """
-    Computes the smoothed (laplacian-filtered) topography needed by the SLEVE
-    coordinate.
-    """
+    """Compute the smoothed (laplacian-filtered) topography needed by the SLEVE coordinate."""
     array_ns = data_alloc.array_namespace(topography)
     smooth_topo = topography.copy()
-    # TODO(@halungge): if the input topopgraphy is properly exchanged, which it should this is not needed here.
+    # TODO(@halungge): if the input topography is properly exchanged, which it should, this is not needed here.
```
nfarabullini:
```suggestion
    # TODO(): if the input topography is properly exchanged, which it should, this is not needed here.
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/config.py
```diff
@@ -0,0 +1,133 @@
...
+                parameters=from_file_ic.FromFileParameters(
+                    data_path=data_path / fortran_config.SER_DATA_SUBDIR,
+                    ntracer=ntracer,
+                ),
+            )
+
+        testcase_nml = input_dict.get("nh_testcase_nml", {})
+        parameters: (
+            jw_ic.JablonowskiWilliamsonParameters | gauss_ic.Gauss3DParameters
+        )  # otherwise mypy complains
```
nfarabullini:
```suggestion
        )  # mypy does not automatically catch type
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/from_file.py
```diff
@@ -0,0 +1,136 @@
...
+        return array_ns.asarray(array_ns.squeeze(ser.read(name, sp).astype(float))[:num_edges, :])
+
+    state = prognostics.initialize_prognostic_state(grid=grid, allocator=allocator, ntracer=ntracer)
+    state.rho.ndarray[:, :] = read_cell_k("rho_now")
+    state.exner.ndarray[:, :] = read_cell_k("exner_now")
+    state.theta_v.ndarray[:, :] = read_cell_k("theta_v_now")
+    state.vn.ndarray[:, :] = read_edge_k("vn_now")
+    state.w.ndarray[:, :] = read_cell_k("w_now")
+
+    if ntracer > 0:
```
nfarabullini, line 91: is ntracer ever 0?

nfarabullini (commented):
my two cents

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/from_file.py
DropD: It's the default value

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
DropD: Or alternatively add where returning `None` is expected to trigger the error, to help the next dev to touch this?

model/common/src/icon4py/model/common/topography/config.py
DropD: `_` does not have any special meaning to the Python interpreter, it's a variable name like any other. It's a convention to use `_` for the "catchall" case when you don't use the value. Here, the value is used (it's printed out), so it makes sense to name it.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
DropD: The return type is `None`, not returning anything in Python is equivalent to `return None` on the last line. Seems to match up perfectly.

model/common/src/icon4py/model/common/topography/smoothing.py
DropD: Does the name after the TODO mean "assigned to" or does it mean "reported by"? I always assumed the latter but could not find anything in the coding guidelines.

tach.toml
jcanton: I removed the TODO because I resolved it, or? 😛

model/testing/src/icon4py/model/testing/definitions.py
jcanton: I'm not sure I understand the question.
yes we use the duplicated `ExperimentConfig` class which is defined in `standalone_driver/config.py` 

https://github.com/C2SM/icon4py/pull/1295/changes#diff-1374141d85d171d3fa2c5711b656072708d65ff1080beaf0cdeedcc64f105bd7R90

this is because the standalone driver needs to read the config independently from the tests (and can't import from model.testing)

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: no due to circular imports, it's "just" these two: `create_experiment_configuration` and the corresponding class `ExperimentConfig`.
we (@msimberg and I) don't like it too much either, but could not come up with a better solution for this issue

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
jcanton: (I only moved this function from a file to another, didn't actually write it myself in this PR).
I have found it confusing too at some point: this one is editing variables in place, instead of getting them as args and returning the edited versions as outputs at the end.
I'm not personally a fan...
wanna make a small separate PR to change this here and anywhere else it may be done this way?

model/common/src/icon4py/model/common/topography/smoothing.py
jcanton: I always assumed it means "assigned to", so I'll change to myself and try to remember to investigate this, now there are enough tests in place to check it

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: I edited this by adding the options to the AdvectionConfig, they are sane values, just not ported to ICON4Py, so if anybody ever tries to use them the advection granule will error out (by design, there were already measures in place)

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/from_file.py
jcanton: yes, can be zero when we're not using tracer advection

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: correction: that idea didn't work because `convert_config_to_advection` is called right away..
I'm implementing your suggestion for now, but this needs a more robust solution (soon ™️ )

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

tach.toml
nfarabullini: ah brain fart on my side, I read it the other way around

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
nfarabullini: I don't quite understand why we edit all of those variables if we don't end up returning any of them and they are not global or am I missing something?

model/testing/src/icon4py/model/testing/datatest_utils.py
nfarabullini: ok then can you add it as an issue to investigate on the board? Maybe I pick it up in the next cool down

ci/base.yml
```diff
@@ -71,7 +71,7 @@ build_image_aarch64:
     ICON4PY_ENABLE_GRID_DOWNLOAD: true
     ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
     ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR: /gt4py_cache
-    ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: false
+    ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: true
```
msimberg: Reminder to revert this before merging. Same in distributed.yml.

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
```diff
@@ -0,0 +1,43 @@
...
+    try:
+        cell_x = grid_manager.coordinates[dims.CellDim]["lon"].ndarray
+    except KeyError:
+        cell_x = grid_manager.coordinates[dims.CellDim]["x"].ndarray
+
+    array_ns = data_alloc.array_namespace(cell_x)
+
+    topo = array_ns.zeros_like(cell_x)
+
+    return topo
```
msimberg:
```suggestion
    return array_ns.zeros_like(cell_x)
```
?

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
```diff
@@ -0,0 +1,43 @@
...
+def flat_topography(
+    *,
+    config: FlatTopographyConfig,
+    grid_manager: gm.GridManager,
+) -> data_alloc.NDArray:
+
+    try:
+        cell_x = grid_manager.coordinates[dims.CellDim]["lon"].ndarray
+    except KeyError:
+        cell_x = grid_manager.coordinates[dims.CellDim]["x"].ndarray
```
msimberg: Is this meant to deal with sphere vs torus? If yes, can we instead match/case on the grid type?

model/common/src/icon4py/model/common/topography/analytical/gaussian_hill.py
```diff
@@ -0,0 +1,51 @@
...
+    cell_x = grid_manager.coordinates[dims.CellDim]["x"].ndarray
+    cell_y = grid_manager.coordinates[dims.CellDim]["y"].ndarray
+
+    array_ns = data_alloc.array_namespace(cell_x)
+
+    dist = ((cell_x - mount_x) ** 2 + (cell_y - mount_y) ** 2) ** 0.5
+
+    topo = mount_height * array_ns.exp(-((dist / mount_width) ** 2))
+
+    return topo
```
msimberg:
```suggestion
    return mount_height * array_ns.exp(-((dist / mount_width) ** 2))
```
?

ci/base.yml
jcanton: REMEMBER to restore this

ci/distributed.yml
```diff
@@ -65,7 +65,7 @@ build_distributed:
     ICON4PY_ENABLE_GRID_DOWNLOAD: true
     ICON4PY_ENABLE_TESTDATA_DOWNLOAD: true
     ICON4PY_CI_GT4PY_BUILD_CACHE_BASE_DIR: /gt4py_cache
-    ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: false
+    ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: true
```
jcanton: REMEMBER to restore this

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
jcanton: turns out you can do
```
def foo(a):
  a = 4

a = 2
foo(a)
print(a)
> 4
```

this is how that function works 🫨

model/common/src/icon4py/model/common/topography/smoothing.py
nfarabullini: it does mean "assigned to"

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: good idea: https://github.com/C2SM/icon4py/issues/1305

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

nfarabullini (approved):

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -85,6 +85,35 @@ class SingleMomentSixClassIconGraupelConfig:
...
+                exponent_for_density_factor_in_ice_sedimentation=nwp_tuning_nml["tune_icesedi_exp"],
+                power_law_coeff_for_snow_fall_speed=nwp_tuning_nml["tune_v0snow"],
+                rain_mu=nwp_phy_nml["mu_rain"],
+                rain_n0=nwp_phy_nml["rain_n0_factor"],
+                snow2graupel_riming_coeff=nwp_tuning_nml["tune_zcsg"],
+                **overrides,
+            )
+        # If the required namelists are not present it means that this
+        # experiment was not run with microphysics.
+        # Throw an error if trying to acces this config in the granule instead
```
msimberg:
```suggestion
        # Throw an error if trying to access this config in the granule instead
```

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -85,6 +85,35 @@ class SingleMomentSixClassIconGraupelConfig:
...
+        # of initializing with default values, which may hide difficult to find
+        # bugs.
+        return None
```
msimberg: This is not consistent with the return type. Haveing `None` here will of course propagate further up, but I think that's probably a net benefit.

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
```diff
@@ -0,0 +1,42 @@
...
+from icon4py.model.common.utils import data_allocation as data_alloc
+
+
+if TYPE_CHECKING:
+    from icon4py.model.common.grid import grid_manager as gm
+
+
+@dataclasses.dataclass
+class FlatTopographyConfig:
+    # Empty config class, used for match/case statement in topography.config
```
msimberg: Could be called "tag type" in the comment (because it acts as a tag for dispatching). This might be more of a C++-ism though.

This is more FYI, no need to change if "tag type" is more confusing.

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
```diff
@@ -0,0 +1,42 @@
...
+
+    match grid_manager.grid.geometry_type:
+        case icon_grid.GeometryType.ICOSAHEDRON:
+            cell_x = grid_manager.coordinates[dims.CellDim]["lon"].ndarray
+        case icon_grid.GeometryType.TORUS:
+            cell_x = grid_manager.coordinates[dims.CellDim]["x"].ndarray
+
+    array_ns = data_alloc.array_namespace(cell_x)
+
+    return array_ns.zeros_like(cell_x)
```
msimberg, line 43: Super-nit: all you really need is the horizontal cell dimension? So possibly even this would work:

```suggestion
    return array_ns.zeros((grid_manager.grid.num_cells,), dtype=ta.wpfloat)
```

_except_ that then you need to get `array_ns` from somewhere else (I'm not sure if we have `array_ns`/`allocator`/`backend` available from `grid_manager` or `grid` (I think not).

This is optional.

model/common/src/icon4py/model/common/topography/config.py
```diff
@@ -0,0 +1,109 @@
...
+                )
+            case "gauss3D":
+                log.info("Analytical Gaussian hill topography")
+                config = fortran_config.config_dataclass_from_dict(
+                    gausshill_topo.GaussianHillConfig, testcase_nml
+                )
+            case name:
+                raise ValueError(f"Unknown or missing test case name: {name!r}")
+
+        return cls(config=config)
```
msimberg: I think this is probably one of those cases where

```suggestion
        return cls(config)
```
is perfectly fine.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/config.py
```diff
@@ -0,0 +1,131 @@
...
+            return from_file_ic.read_from_file(
+                config=config.config,
+                grid=grid,
+                interpolation_field_source=interpolation_field_source,
+                metrics_field_source=metrics_field_source,
+                backend=backend,
+                exchange=exchange,
+            )
+        case _:
+            raise TypeError(f"Unknown IC config type: {type(config.config)!r}")
```
msimberg: Probably doesn't hurt to spell it out explicitly:
```suggestion
            raise TypeError(f"Unknown initial conditions config type: {type(config.config)!r}")
```
?

tach.toml
msimberg, line 73: I think this should be reverted since it's unrelated, though it doesn't matter much since the old driver package will be deleted.

msimberg (commented):
Only minor nits remain.

@jcanton already started looking into separating the standalone driver entry point into two levels where main.main takes config file paths and a lower level takes an actual config object. This separation should make the standalone driver tests nicer because individual config options can be overridden as needed (such as the end date). I really like this cleanup, but in the interest of unblocking @yiluchen1066 and @DropD with their config-related changes, that should be left for a separate PR. This is also the reason why @jcanton increased the tolerances in the standalone driver single rank test (because it now has to do more time steps, and the error increases).

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
jcanton: yeah... that was why I lazily did it like that...since we stopped passing `array_ns` around 🤷 
should we add  `array_ns`/`allocator`/`backend` available from `grid_manager` ?

tach.toml
jcanton: it is not unrelated: it's related to me doing in this PR something that should have been done separately (as too often happens) and moving reading `ICON4PY_DRIVER_LOGGING_LEVEL` from `testing/config` `into driver_utils.py`

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/common/src/icon4py/model/common/topography/analytical/flat_topography.py
msimberg: Let's keep it like this for the moment. I'm not convinced the grid_manager (or grid) should kepe that around permanently.

tach.toml
msimberg: Ah indeed, thanks for the correction.

In my head effectively every package depends on testing for the tests, but I don't know if that needs to be visible in pyproject.toml. This is at least consistent with other packages for now.

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -85,6 +85,35 @@ class SingleMomentSixClassIconGraupelConfig:
     #: coefficient for snow-graupel conversion by riming. Originally defined as csg in mo_nwp_tuning_config.f90 in ICON.
     snow2graupel_riming_coeff: ta.wpfloat = 0.5
 
+    @classmethod
+    def from_fortran_dict(
+        cls, atmo_dict: dict[str, Any], **overrides: Any
+    ) -> SingleMomentSixClassIconGraupelConfig:
```
DropD, line 93:
```suggestion
    ) -> SingleMomentSixClassIconGraupelConfig | None:
```

DropD: If this signature is a problem for some reason, I would suggest:

```python
class NoMicrophysics(SingleMomentSixClassIconGraupelConfig):
    def __getattr__(self, name: str) -> typing.Never:
        raise NoMicrophysicsError

    def __setattr__(self, name: str, value: Any) -> typing.Never:
        raise NoMicrophysicsError

class NoMicrophysicsError(Exception):
    """Raise this if microphysics was not configured but the configuration for it is accessed anyway."""
    ...
```

And then return `NoMicrophysics()` instead of `None`, so the signature is unchanged, no downstream `is not None` is required and downstream code will raise a descriptive exception if it doesn't deal with this case explicitly

DropD: Also, is this module excluded from typing checks? I would have expected mypy to catch the mismatching return type annotation!

DropD (changes requested):

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
msimberg: Unfortunately the majority of icon4py is still excluded from type checks. See #1137. If you have spare time in between other tasks, even enabling type checks a single new file would be useful!

jcanton: I implemented the second approach, thanks for the suggestion, I am already working on a separate PR to handle this better and more generally at the `read_config` level though, so this is really only temporary

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

DropD (approved):

ci/distributed.yml
```diff
@@ -108,7 +108,7 @@ build_distributed:
     - if: ($COMPONENT == 'common' && ($BACKEND == 'dace_cpu' || $BACKEND == 'dace_gpu' || $BACKEND == 'gtfn_gpu')) || ($COMPONENT == 'standalone_driver' && $BACKEND == 'dace_gpu')
```
msimberg: I think you need one more round with this:
```suggestion
    - if: ($COMPONENT == 'common' && ($BACKEND == 'dace_cpu' || $BACKEND == 'dace_gpu' || $BACKEND == 'gtfn_gpu')) || $COMPONENT == 'standalone_driver'
```
`dace_cpu`, `gtfn_cpu`, and `gtfn_gpu` used to be ok with the 50 minutes in the fallback case, but they also need more time now. It's likely easiest to group them all in the same case. We can reduce this again with #1287. (standalone_driver + gtfn_cpu happened to have a working cache and ran in only ~3 minutes: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504675/-/jobs/14808045018; but that was the best case).

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

PR #1302, opened 2026-06-08
Use nox sessions for MPI tests in CSCS CI

msimberg:
The MPI tests were previously launched manually with pytest because nox will create a venv for each session. Because mpirun/srun is outside the nox session, the nox sessions overlap and create problems when every rank tries to initialize/update the venv.

This adds an MPI nox session that uses a separate nox envdir per rank (set in the noxfile; if `--envdir` is passed, that takes precedence and the caller is responsible). I considered trying to reuse the env from the e.g. the first rank, or an external venv, but this either means callers have to know to use it, or introduces synchronization requirements within the nox sessions (e.g. ranks > 0 have to wait to make sure the first rank has created the venv). There's a slight downside with uv creating n separate venvs for each rank, but I find it's much easier to ensure the venvs are correct like this.

#1292 is now merged into this PR. #1292 needs to be merged first.


msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

msimberg:
cscs-ci run distributed

ci/distributed.yml
```diff
@@ -80,29 +78,33 @@ build_distributed:
     FI_PROVIDER: cxi
     OMPI_MCA_pml: cm
     OMPI_MCA_mtl: ofi
+    # TODO(msimberg): these will come from base template after #1292 is merged
+    # (must happen before this branch is merged).
```
msimberg: To do: resolve this after #1292 is merged.

noxfile.py
```diff
@@ -21,6 +21,14 @@
 nox.options.default_venv_backend = "uv"
 nox.options.sessions = ["test_model", "test_tools_and_bindings"]
 
+_rank = (
+    os.environ.get("PMI_RANK")
+    or os.environ.get("OMPI_COMM_WORLD_RANK")
+    or os.environ.get("SLURM_PROCID")
+)
+if _rank is not None:
+    nox.options.envdir = f".nox-{_rank}"
```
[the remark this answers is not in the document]
msimberg: I think this is a non-concern. Ranks don't really have to be numeric, and if they contain `/` so be it... the prefix is still relative to the current directory.

msimberg:
cscs-ci run distributed

ci/scripts/ci-mpi-wrapper.sh
```diff
@@ -20,9 +20,9 @@ fi
 log_file="${CI_PROJECT_DIR:+${CI_PROJECT_DIR}/}pytest-log-rank-${rank}.txt"
 
 if [[ "${rank}" -eq 0 ]]; then
```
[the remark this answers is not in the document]
msimberg: It's probably a good thing if it fails on non-integer ranks... That would be a very unusual environment.

msimberg:
cscs-ci run distributed

ci/docker/checkout_mpi.Dockerfile
```diff
@@ -5,11 +5,14 @@ COPY . /icon4py
...
+ENV UV_MANAGED_PYTHON=1
+ENV UV_CACHE_DIR=/opt/uv-cache
 ENV MPI4PY_BUILD_BACKEND=scikit-build-core
 ENV GHEX_USE_GPU=ON
 ENV GHEX_GPU_TYPE=NVIDIA
 ENV GHEX_GPU_ARCH=90
 ENV GHEX_TRANSPORT_BACKEND=MPI
-RUN uv sync --extra all --extra cuda12 --python=$PYVERSION
+RUN uv sync --extra all --extra cuda12 --python=$PYVERSION && \
+    chmod -R a+rwX "$UV_CACHE_DIR"
```
egparedes: Interesting, is this needed to reuse the cache in the container image?

ci/distributed.yml
```diff
@@ -80,29 +78,30 @@ build_distributed:
...
-    - echo "running with $(python --version)"
     - source ci/scripts/start-cuda-mps.sh
     - source ci/scripts/gt4py-cache.sh
   script:
-    - ci/scripts/ci-mpi-wrapper.sh pytest -sv -k mpi_tests --with-mpi --backend=$BACKEND model/$COMPONENT --level=$LEVEL
+    - ci/scripts/ci-mpi-wrapper.sh nox -s "test_model_mpi-3.13($SELECTION, $COMPONENT)" -- --backend=$BACKEND --level=$LEVEL
   parallel:
     matrix:
-      - COMPONENT: [atmosphere/diffusion, atmosphere/dycore, atmosphere/advection, common, standalone_driver]
+      - COMPONENT: [diffusion, dycore, advection, common, standalone_driver]
```
egparedes, line 37: Just a question: why dropping the "namespace" from the list of components?

egparedes (approved):
LGTM. Just a couple of questions

ci/docker/checkout_mpi.Dockerfile
msimberg: Yeah. The cache at container image build-time is created as root, and when the test jobs are run they run as the CI service user. The service user wouldn't be able to read the cache unless the permissions are changed to allow anyone to read it (I suppose writing might not be necessary, I didn't check if I can remove that).

ci/distributed.yml
msimberg: Only to align with the non-MPI model nox session, which currently only use the last component of the subdirs. I'm just reusing the existing `MODEL_SUBPACKAGE_PATHS`. Since there was no nox session we had to pass the full (minus `model/`) path for the MPI tests previously.

It did cross my mind if it wouldn't be more explicit to just use the full path, but changing that is not for this PR...

ci/distributed.yml
msimberg: I merged #1292 into this PR so the conflicts are resolved, but need to double check once #1292 is actually merged.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1304, opened 2026-06-10
refactor standalone_driver

jcanton:
refactor driver cli/testing interface from a single `main.main()` entry point to one for testing and one dedicated to CLI

also allows for overriding config variables read from the fortran namelists such as end_date / n_time_steps 
which allows for restoring the original tolerances that were altered by https://github.com/C2SM/icon4py/pull/1295 since the JW experiment ran for 3 time steps instead of 1


jcanton:
cscs-ci run default

jcanton:
ah, I cannot run CI though...

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -54,6 +54,7 @@ class DriverConfig:
     dtime: datetime.timedelta
     start_date: datetime.datetime
     end_date: datetime.datetime
+    n_time_steps: int | None = None  # when set, it overrides date-based computation
```
msimberg: What do you think about configuring this as
```suggestion
end_<need to find a good name for this>: datetime.datetime | NumTimeSteps
```
instead? If this is something you like, it might even be:
```suggestion
end_<...>: RelativeTime <i.e. timedelta> | AbsoluteTime <i.e. datetime> | NumTimeSteps <i.e. int>
```
And if we go this direction, probably we should have:
```
type SimulationLengthSpec = <relative> | <absolute> | <n steps>
```

Food for thought, not a requirement for this PR.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -544,34 +512,25 @@ def initialize_driver(
         process_props=process_props,
     )
```
msimberg: Would you object to having this as an explicit call after calling `setup_environment`? (In which case maybe `setup_environment` needs a rename...).

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
msimberg: Does it make sense to log in each branch whether MPI is enabled or not? Or log once after logging is set up? I worry about stuff like the binary wheel incompatibility (https://github.com/C2SM/icon4py/blob/main/README.md#compiling) making this silently fall back to each rank running a separate simulation.

model/standalone_driver/tests/standalone_driver/integration_tests/test_initial_conditions.py
```diff
@@ -49,15 +54,26 @@ def test_initial_conditions(
...
+    grid_manager = driver_utils.create_grid_manager(
         grid_file_path=grid_file_path,
-        config_file_path=config_file_path,
-        log_level=next(iter(driver_utils._LOGGING_LEVELS.keys())),
-        output_path=tmp_path / "ci_driver_output",
-        backend_like=backend_like,
+        vertical_grid_config=config.vertical_grid,
+        allocator=allocator,
+        process_props=process_props,
+    )
```
msimberg, line 57: Not 100% sure if this is possible now, but (github doesn't like my `suggestion` block as a diff, so you get it as a full code block instead):

```python
@pytest.mark.datatest
def test_initial_conditions(
    experiment: definitions.Experiment, # <--- config from here
    data_provider: sb.IconSerialDataProvider,
    tmp_path: pathlib.Path,
    process_props: decomp_defs.ProcessProperties,
    backend_like: model_backends.BackendLike,
) -> None:
    backend = model_options.customize_backend(program=None, backend=backend_like)
    allocator = model_backends.get_allocator(backend)

    grid_file_path = grid_utils._download_grid_file(experiment.experiment_description.grid)
    config_file_path = dt_utils.get_path_for_experiment(eexperiment_description, process_props)

    config = experiment.config.with_overrides(driver={"output_path": tmp_path / "ci_driver_output"})

    grid_manager = driver_utils.create_grid_manager(
        grid_file_path=grid_file_path,
        vertical_grid_config=config.vertical_grid,
        allocator=allocator,
        process_props=process_props,
    )
```

If/when we have grid/grid_manager fixtures you could also get those for free.

And separately (the test already does this, so feel free to ginore, but I noticed since you now have to create a `backend`): use the `backend` fixture instead? It effectively does exactly what your oneliner `backend = customize_backend(...)` does.

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
```diff
@@ -65,15 +66,33 @@ def test_standalone_driver(
...
+    config = driver_config.read_config(config_file_path)
+    config = config.with_overrides(
+        driver={
+            "output_path": tmp_path / "ci_driver_output",
+            "end_date": datetime.datetime.fromisoformat(timeloop_date_exit).replace(
+                tzinfo=datetime.timezone.utc
+            ),
+        }
+    )
+
```
msimberg: Same comment here as in the initial conditions test: backend fixture, config from experiment fixture?

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_initial_conditions.py
```diff
@@ -81,17 +87,30 @@ def test_initial_conditions_compare_single_multi_rank(
         f"running on {process_props.comm} with {process_props.comm_size} ranks and atol = {atol}, rtol = {rtol}"
     )
 
+    backend = model_options.customize_backend(program=None, backend=backend_like)
```
msimberg: Same comment about config and backend.

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_initial_conditions.py
```diff
@@ -81,17 +87,30 @@ def test_initial_conditions_compare_single_multi_rank(
...
+    allocator = model_backends.get_allocator(backend)
+
     grid_file_path = grid_utils._download_grid_file(experiment_description.grid)
     config_file_path = dt_utils.get_path_for_experiment(experiment_description, process_props)
 
+    config = driver_config.read_config(config_file_path)
+
+    serial_process_props = decomp_defs.SingleNodeProcessProperties()
+    serial_config = config.with_overrides(
+        driver={"output_path": tmp_path / "ci_driver_output_serial_rank0"}
```
msimberg: Configure this with `rank_{process_props.rank}` I think? Doesn't every rank run the serial test separately? Or only rank zero? If every rank does it, we can of course (try to) change it so that only rank zero does the serial test.

Edit: Then again this isn't new, so feel free to ignore. Since the test also doesn't write output, it probably doesn't matter much. But might be good to keep track of at least in case we ever enable output on these tests (I don't think we should, but still...).

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_standalone_driver.py
```diff
@@ -68,22 +69,51 @@ def test_standalone_driver_compare_single_multi_rank(
         f"running on {process_props.comm} with {process_props.comm_size} ranks and atol = {atol}, rtol = {rtol}"
     )
 
+    backend = model_options.customize_backend(program=None, backend=backend_like)
```
msimberg: <config + backend comment> (see other comments).

msimberg (changes requested):
Minor comments. I really like the split a lot.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton: we can also get rid of setup_environment and move these two out of here

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
msimberg: I think this would effectively be equivalent:
```
    process_props = decomposition_defs.get_process_properties(
        decomposition_defs.get_runtype(with_mpi=mpi_decomp.mpi4py is not None)
    )
```

model/standalone_driver/tests/standalone_driver/integration_tests/test_initial_conditions.py
jcanton: good point, I think I overdid it setting up the standalone_driver tests (single and distributed) which instead need to read the config independently from experiment.config to test the reading, but here it's not necessary

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
jcanton: here I'd leave the config reading explicit so we test that the reading works, since it goes through a different (and cough cough untested) code path since we had to duplicate the ExperimentConfig and reading constructor

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -53,7 +59,7 @@ class DriverConfig:
     profiling_stats: ProfilingStats | None
     dtime: datetime.timedelta
```
jcanton: this can be changed to `RelativeTime`

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -34,6 +34,12 @@
 log = logging.getLogger(__name__)
 
 
+RelativeTime = datetime.timedelta
+AbsoluteTime = datetime.datetime
+NumTimeSteps = int
+EndSimulation = RelativeTime | AbsoluteTime | NumTimeSteps
```
msimberg:
```suggestion
RelativeTime: TypeAlias = datetime.timedelta: TypeAlias
AbsoluteTime = datetime.datetime
NumTimeSteps: TypeAlias = int
EndSimulation: TypeAlias = RelativeTime | AbsoluteTime | NumTimeSteps
```
for now. After #1248, they could be `type RelativeTime = datetime.timedelta` etc.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -95,17 +92,25 @@ class ModelTimeVariables:
     cfl_watch_mode: bool = dataclasses.field(init=False)
 
     def __post_init__(self, config: driver_config.DriverConfig) -> None:
-        self.n_time_steps = int((config.end_date - config.start_date) / config.dtime)
+        match config.end_simulation:
+            case driver_config.NumTimeSteps(n):
+                self.n_time_steps = n
+                self.simulation_date = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
```
msimberg: Should `start_date` not be used in this case as well?

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -463,8 +464,13 @@ def display_driver_setup_in_log_file(
...
     log.info(f"Start date             : {config.start_date}")
-    log.info(f"End date               : {config.end_date}")
-    log.info(f"Number of timesteps    : {n_time_steps}")
+    match config.end_simulation:
+        case int(n):
+            log.info(f"Number of timesteps    : {n}")
+        case datetime.timedelta() as relative:
+            log.info(f"Simulation duration    : {relative.total_seconds()} s")
+        case datetime.datetime() as absolute:
+            log.info(f"End date               : {absolute}")
```
msimberg: This needs to be updated to the type aliases.

Also, is duration in seconds a useful measure (for humans in a log file)? I don't know if the default `datetime.timedelta` format could be used instead?

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
msimberg: Ah good point, this applies to all my comments about backend/config fixtures I think. The one from experiment is now the wrong type, but might happen to work because it has the same shape as the config from standalone_driver. I'm ok with leaving these as you had them. This needs a bit more thought. Sorry for the noise.

msimberg (changes requested):

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
jcanton: specify run "mode"
print `self.end_date`

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: and compute `self.end_datetime` in all cases

jcanton: rename `simulation_date` to `current_datetime`

jcanton (commented):
review from a friend Hi! Nice refactoring — the split of main() into main (CLI glue) + initialize_driver + run_driver is a big improvement and with_overrides on a frozen ExperimentConfig is much cleaner than in-place mutation. A few things I noticed: 1. BUG: main.py — decomposition_defs and mpi_decomp used but not imported. Lines 65-66 reference decomposition_defs.get_process_properties(...) decomposition_defs.get_runtype(...) and mpi_decomp.mpi4py but neither name is imported into main.py. The old decomp_defs was removed and decomposition_defs/mpi_decomp are only available in standalone_driver module scope — they aren't re-exported. This will raise NameError at runtime when the CLI entry point is invoked. Fix: add `from icon4py.model.common.decomposition import definitions as decomposition_defs mpi_decomposition as mpi_decomp` 2. driver_states.py:98 — NumTimeSteps case sets simulation datetime to epoch. simulation_datetime is hardcoded to datetime(1970 1 1 ...) instead of config.start_date. Already noted in thread and author agreed to fix. 3. config.py:37-40 — type aliases should use TypeAlias. Already flagged by msimberg. Consider also whether dtime on DriverConfig should be typed as RelativeTime (as author self-noted). 4. driver_utils.py:467-473 — match uses raw types instead of type aliases. match arms use int(n) datetime.timedelta() datetime.datetime() instead of driver_config.NumTimeSteps / RelativeTime / AbsoluteTime aliases. Already flagged. Also consider logging timedelta with str(relative) rather than .total_seconds() for human readability. 5. test_initial_conditions.py:60-61 — experiment param declared but body uses experiment_description. The function takes experiment: definitions.Experiment (line 51) for experiment.config but grid/config paths still reference experiment_description.grid / experiment_description via parametrized fixture. This works but is confusing — function depends on two separate fixture sources for related data. Minor suggestion: since experiment already contains experiment_description consider using experiment.experiment_description.grid and experiment.experiment_description for path lookups and dropping separate experiment_description fixture. 6. test_parallel_standalone_driver.py:87 106 — grid manager created from config not serial_config/mpi_config. Reads config.vertical_grid instead of serial_config.vertical_grid / mpi_config.vertical_grid. Functionally identical since vertical_grid is not overridden but inconsistent with test_parallel_initial_conditions.py pattern. 7. test_standalone_driver.py — experiment_description still declared as separate fixture parameter. Same pattern as #5. Not a bug but inconsistent with test_initial_conditions.py which uses experiment fixture for config.

model/standalone_driver/src/icon4py/model/standalone_driver/main.py
```diff
@@ -54,58 +51,47 @@ def main(
...
-        config_file_path=config_file_path,
-        output_path=output_path,
-        log_level=log_level,
-        print_distributed_debug_msg=print_distributed_debug_msg,
-        backend_like=icon4py_backend,
-        force_serial_run=force_serial_run,
+    backend = model_options.customize_backend(
+        program=None, backend=driver_utils.get_backend_from_name(icon4py_backend)
     )
+    allocator = model_backends.get_allocator(backend)
```
jcanton, line 71: from a friend: `decomposition_defs` and `mpi_decomp` are used at line 65-66 but neither is imported in this file. The old `from icon4py.model.common.decomposition import definitions as decomp_defs` was removed without replacement. These names exist in `standalone_driver` module scope but aren't re-exported. Add:
```python
from icon4py.model.common.decomposition import definitions as decomposition_defs, mpi_decomposition as mpi_decomp
```

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: from a friend: When `end_simulation=NumTimeSteps(n)`, `simulation_datetime` is set to `datetime(1970, 1, 1, ...)` instead of `config.start_date`. The simulation clock starts at epoch rather than the configured start date. Should be `self.simulation_datetime = config.start_date` (consistent with the other two match arms).

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -463,8 +464,13 @@ def display_driver_setup_in_log_file(
     log.info(f"Experiment name        : {config.experiment_name}")
     log.info(f"Time step (dtime)      : {config.dtime.total_seconds()} s")
     log.info(f"Start date             : {config.start_date}")
-    log.info(f"End date               : {config.end_date}")
-    log.info(f"Number of timesteps    : {n_time_steps}")
+    match config.end_simulation:
+        case int(n):
```
jcanton: from a friend: These match arms use raw types (`int(n)`, `datetime.timedelta()`, `datetime.datetime()`) instead of the type aliases defined in `config.py` (`NumTimeSteps`, `RelativeTime`, `AbsoluteTime`). The match in `driver_states.py` correctly uses `driver_config.NumTimeSteps(n)` etc. — this function should do the same for consistency.

Also consider logging the timedelta as `str(relative)` rather than `.total_seconds()` for human readability.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -34,6 +34,12 @@
 log = logging.getLogger(__name__)
 
 
+RelativeTime = datetime.timedelta
```
jcanton: from a friend: These type aliases should use `TypeAlias` from `typing` for proper type checking:
```python
from typing import TypeAlias

RelativeTime: TypeAlias = datetime.timedelta
AbsoluteTime: TypeAlias = datetime.datetime
NumTimeSteps: TypeAlias = int
EndSimulation: TypeAlias = RelativeTime | AbsoluteTime | NumTimeSteps
```

model/standalone_driver/tests/standalone_driver/integration_tests/test_initial_conditions.py
```diff
@@ -43,21 +48,31 @@
...
+    experiment: definitions.Experiment,
     data_provider: sb.IconSerialDataProvider,
     tmp_path: pathlib.Path,
     process_props: decomp_defs.ProcessProperties,
     backend_like: model_backends.BackendLike,
+    backend: model_backends.Backend,
 ) -> None:
+    allocator = model_backends.get_allocator(backend)
+
     grid_file_path = grid_utils._download_grid_file(experiment_description.grid)
```
jcanton: from a friend: The function declares `experiment: definitions.Experiment` (for `experiment.config`) but body still references `experiment_description.grid` and `config_file_path = dt_utils.get_path_for_experiment(experiment_description, ...)` where `experiment_description` comes from the parametrized fixture — a second, independent fixture source. Since `experiment` (via `definitions.Experiment`) already wraps `experiment_description`, consider:
```python
grid_file_path = grid_utils._download_grid_file(experiment.experiment_description.grid)
config_file_path = dt_utils.get_path_for_experiment(experiment.experiment_description, process_props)
```
and dropping the separate `experiment_description` fixture/param.

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_standalone_driver.py
```diff
@@ -39,51 +38,82 @@
...
+            "end_simulation": end_simulation,
+        }
+    )
+    serial_grid_manager = driver_utils.create_grid_manager(
         grid_file_path=grid_file_path,
-        config_file_path=config_file_path,
-        output_path=tmp_path / "ci_driver_output_serial_rank0",
-        icon4py_backend=backend_like,
-        force_serial_run=True,
+        vertical_grid_config=config.vertical_grid,
```
jcanton: from a friend: Grid manager is created with `config.vertical_grid` rather than `serial_config.vertical_grid` (also at line 108 with `config.vertical_grid` instead of `mpi_config.vertical_grid`). Functionally identical since `vertical_grid` is never overridden, but inconsistent with `test_parallel_initial_conditions.py` which uses `serial_config.vertical_grid` / `mpi_config.vertical_grid`.

jcanton (commented):
from a friend — inline comments below

model/standalone_driver/src/icon4py/model/standalone_driver/main.py
jcanton, line 71: from a friend: `decomposition_defs` and `mpi_decomp` are used at line 65-66 but neither is imported in this file. The old `from icon4py.model.common.decomposition import definitions as decomp_defs` was removed without replacement. These names exist in `standalone_driver` module scope but aren't re-exported.

**Fix:** Add
```python
from icon4py.model.common.decomposition import definitions as decomposition_defs, mpi_decomposition as mpi_decomp
```

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: from a friend: When `end_simulation=NumTimeSteps(n)`, `simulation_datetime` is set to `datetime(1970, 1, 1, ...)` instead of `config.start_date`. The simulation clock starts at epoch rather than the configured start date.

**Fix:** `self.simulation_datetime = config.start_date` (consistent with the other two match arms).

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
jcanton: from a friend: Match arms use raw types (`int(n)`, `datetime.timedelta()`, `datetime.datetime()`) instead of the type aliases defined in `config.py` (`NumTimeSteps`, `RelativeTime`, `AbsoluteTime`). The match in `driver_states.py:95-104` correctly uses `driver_config.NumTimeSteps(n)` etc. — this function should do the same for consistency.

Also consider logging the timedelta as `str(relative)` rather than `.total_seconds()` for human readability.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
jcanton: from a friend: These type aliases should use `TypeAlias` from `typing` for proper mypy support:
```python
from typing import TypeAlias

RelativeTime: TypeAlias = datetime.timedelta
AbsoluteTime: TypeAlias = datetime.datetime
NumTimeSteps: TypeAlias = int
EndSimulation: TypeAlias = RelativeTime | AbsoluteTime | NumTimeSteps
```
(Already noted by msimberg in the thread.)

model/standalone_driver/tests/standalone_driver/integration_tests/test_initial_conditions.py
jcanton: from a friend: Function declares `experiment: definitions.Experiment` (for `experiment.config` at line 63) but body still references the standalone `experiment_description` parametrized fixture for grid/config paths. Since `experiment` wraps `experiment_description`, consider simplifying to:
```python
grid_file_path = grid_utils._download_grid_file(experiment.experiment_description.grid)
config_file_path = dt_utils.get_path_for_experiment(experiment.experiment_description, process_props)
```

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_standalone_driver.py
jcanton: from a friend: Grid manager uses `config.vertical_grid` (from the un-overridden base config) rather than `serial_config.vertical_grid`. Same at line 108 with `config.vertical_grid` instead of `mpi_config.vertical_grid`. Functionally identical since `vertical_grid` is never overridden, but inconsistent with `test_parallel_initial_conditions.py` which correctly uses the override-specific config.

model/standalone_driver/src/icon4py/model/standalone_driver/main.py
```diff
@@ -12,19 +12,16 @@
...
-    initial_condition,
-    standalone_driver,
-)
+from icon4py.model.common import model_backends, model_options
+from icon4py.model.standalone_driver import config as driver_config, driver_utils, standalone_driver
 
 
 log = logging.getLogger(__name__)
 
+app = typer.Typer()
```
msimberg: Discovered after adding this while looking at the scripts subdir:

```suggestion
app = typer.Typer(no_args_is_help=True)
```

model/standalone_driver/src/icon4py/model/standalone_driver/main.py
```diff
@@ -54,58 +51,47 @@ def main(
...
+        config=config,
+        grid_manager=grid_manager,
+        process_props=process_props,
+        backend=backend,
+    )
 
 
 if __name__ == "__main__":
-    typer.run(main)
+    app()
```
msimberg: Same here, discovered after pushing the previous commit:
```suggestion
    sys.exit(app())
```

msimberg (commented):
Two minor diffs to (my incomplete) typer changes.

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -34,6 +34,12 @@
 log = logging.getLogger(__name__)
 
 
+RelativeTime: TypeAlias = datetime.timedelta
```
msimberg, line 37: Not for this PR, just noting it: This may need to move to a more common place for dtime refactorings (to not use bare floats). dtime/dt is also a `RelativeTime`.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -51,9 +57,9 @@ class DriverConfig:
 
     experiment_name: str
     profiling_stats: ProfilingStats | None
-    dtime: datetime.timedelta
-    start_date: datetime.datetime
-    end_date: datetime.datetime
+    dtime: RelativeTime
+    start_datetime: AbsoluteTime
+    end_simulation: EndSimulation
```
msimberg: 100% bikeshedding (100% ignorable): For some reason `SimulationEnd` sounds more natural to me (or `EndOfSimulation`). `EndSimulation` sounds very imperative to me, i.e. like an action (end it) rather than a thing (the end).

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -70,15 +76,17 @@ def from_fortran_dict(
...
         return cls(
             experiment_name=master_model_nml["model_namelist_filename"]
             .removeprefix("NAMELIST_")
             .removesuffix("_sb_atm"),
             dtime=datetime.timedelta(seconds=dtime),
-            start_date=datetime.datetime.fromisoformat(start_date_str.replace("Z", "+00:00")),
-            end_date=datetime.datetime.fromisoformat(end_date_str.replace("Z", "+00:00")),
+            start_datetime=datetime.datetime.fromisoformat(
+                start_datetime_str.replace("Z", "+00:00")
+            ),
```
msimberg: Not for this PR, I only noticed now: https://github.com/C2SM/icon4py/issues/1321.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -115,8 +136,8 @@ def dtime_in_seconds(self) -> ta.wpfloat:
     def substep_timestep(self) -> ta.wpfloat:
         return ta.wpfloat(self.dtime_in_seconds / self.ndyn_substeps_var)
 
-    def next_simulation_date(self) -> None:
-        self.simulation_date += self.dtime
+    def next_simulation_datetime(self) -> None:
```
msimberg: Not for this PR and not important, but noting in case it comes up in the future: this should probable be something like `advance_simulation_datetime` since it updates the internal state and doesn't return the "next simulation datetime".

model/standalone_driver/tests/standalone_driver/mpi_tests/test_parallel_initial_conditions.py
```diff
@@ -106,13 +125,22 @@ def test_initial_conditions_compare_single_multi_rank(
...
 
+    mpi_config = experiment.config.with_overrides(
+        driver={"output_path": tmp_path / f"ci_driver_output_mpi_rank_{process_props.rank}"}
+    )
+    mpi_grid_manager = driver_utils.create_grid_manager(
+        grid_file_path=grid_file_path,
+        vertical_grid_config=mpi_config.vertical_grid,
+        allocator=allocator,
+        process_props=process_props,
+    )
```
msimberg: Nit, but for consistency I think it'd be nice to name these `multi_rank_config` etc. On one hand it's not MPI that's important (we can have MPI without multiple ranks) and on the other `mpi_config` says to me "this is the config for MPI itself", not "the config used when MPI is enabled".

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
msimberg: Most likely cheap friend reverted this after #1295 was merged and this branch updated?

msimberg (changes requested):
Only the `mpi_config` and tolerance TODOs are blocking from my perspective. The others are optional and/or preferable to leave for the future (if at all).

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
jcanton: https://github.com/C2SM/icon4py/issues/1323

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: done

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
jcanton: fixed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

msimberg (approved):
Looks good to me, but something is not right with the standalone driver test: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/jobs/14888741985#L635. Is it possibly still incorrectly running more than one time step, hence the diff is still very large?

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

PR #1310, opened 2026-06-11
Optional configs handling

jcanton:
Cleanup the hacky handling of optional `Config`s (such as tracer advection and graupel).
and while we are at it, make every granule and its config optional, so all are consistent now


model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -102,8 +108,8 @@ class ExperimentConfig:
     topography: topography.TopographyConfig
     nonhydrostatic: solve_nh.NonHydrostaticConfig
     diffusion: diffusion.DiffusionConfig
-    advection: advection.AdvectionConfig
-    graupel: graupel.SingleMomentSixClassIconGraupelConfig
+    tracer_advection: tracer_advection.AdvectionConfig | None = None
+    graupel: graupel.SingleMomentSixClassIconGraupelConfig | None = None
```
msimberg: I'm unsure about the consequences of the following, but did you try out:
```
ExperimentConfig:
  AdvectionConfig:
    do_tracer_advection: bool
    ...
  GraupelConfig:
    do_physics: bool
```
?

I think this approach might simplify some of the uses of the advection and graupel configs (since they'd never be `None`). On the other hand the configs would always have to be created with default values, with the flags deciding if they should be used or not.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -66,6 +66,12 @@ class DriverConfig:
     ndyn_substeps: int = 5
     enable_statistics_output: bool = False
     ntracer: int = 0
+    do_tracer_advection: bool = True
+    do_physics: bool = True
+
+    def __post_init__(self) -> None:
```
jcanton: from a friend: Using `object.__setattr__` to mutate a frozen dataclass is functional but cuts against the purpose of freezing. Since `from_fortran_dict` already accepts `**overrides`, consider computing `ntracer=0` there and passing it explicitly when `do_tracer_advection` is False, instead of patching it in `__post_init__`:
```python
@classmethod
def from_fortran_dict(cls, ..., do_tracer_advection=True, ...):
    ntracer = 0 if not do_tracer_advection else fortran_config.list_to_value(run_nml["ntracer"])
    return cls(
        ...,
        ntracer=ntracer,
        do_tracer_advection=do_tracer_advection,
    )
```
Then drop `__post_init__` entirely.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -346,42 +351,49 @@ def initialize_granules(
...
-                interpolation_attributes.RBF_VEC_COEFF_E
+    tracer_advection_granule: advection.Advection | None = None
+    if config.driver.do_tracer_advection:
+        assert config.tracer_advection is not None
+        tracer_advection_granule = advection.convert_config_to_advection(
+            grid=grid,
+            backend=backend,
+            config=config.tracer_advection,
+            interpolation_state=advection_states.AdvectionInterpolationState(
+                geofac_div=interpolation_field_source.get(interpolation_attributes.GEOFAC_DIV),
```
jcanton, line 406: from a friend: `assert config.tracer_advection is not None` inside `if config.driver.do_tracer_advection:` will pass in dev mode but is stripped with `python -O`. At -O the next line would silently pass `None` into `convert_config_to_advection`. Consider a regular `if`/`raise` or just trust the invariant since the guard ensures it.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -72,26 +72,17 @@ class DriverStates(NamedTuple):
     diagnostic: diagnostics.DiagnosticState
 
 
-@dataclasses.dataclass
 class ModelTimeVariables:
```
jcanton, line 76: from a friend: `ModelTimeVariables` is no longer a dataclass — the field declarations (`n_time_steps`, `dtime`, etc.) are gone. These attributes are now dynamically assigned in `_init_from_config`, which means type checkers (mypy/Pyright) and IDEs won't know they exist. Add type annotations for the fields, even if they can't be dataclass fields:
```python
class ModelTimeVariables:
    n_time_steps: int
    dtime: datetime.timedelta
    # ... rest of fields ...
    
    def __init__(self, config: ...) -> None:
        ...
```

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -93,48 +92,21 @@ def from_fortran_dict(
...
-
-class NoMicrophysicsError(Exception):
-    """Raise this if microphysics was not configured but the configuration for it is accessed anyway."""
-
-    ...
+        nwp_phy_nml = atmo_dict["nwp_phy_nml"]
+        nwp_tuning_nml = atmo_dict["nwp_tuning_nml"]
+        return cls(
+            do_latent_heat_nudging=run_nml["ldass_lhn"],
+            use_constant_latent_heat=fortran_config.list_to_value(nwp_phy_nml["ithermo_water"])
```
jcanton, line 99: from a friend: `from_fortran_dict` now uses direct dict access (`atmo_dict["nwp_phy_nml"]`) instead of `.get()`, so it will raise `KeyError` if called without the guard from `read_config`. All current callers check `"nwp_phy_nml" in atm_dict` first, so this is safe today but makes the function less reusable if someone calls it directly without the guard. Consider an explicit guard or keeping the defensive `.get()` pattern with a clear error message:
```python
nwp_phy_nml = atmo_dict.get("nwp_phy_nml")
if nwp_phy_nml is None:
    raise KeyError("nwp_phy_nml required — caller must check before calling from_fortran_dict")
```
(Minor — mostly about self-defending API boundaries.)

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -392,7 +390,9 @@ def _adjust_ndyn_substeps_var(
 
     def _update_spinup_second_order_divergence_damping(self) -> ta.wpfloat:
         if self.config.driver.apply_extra_second_order_divdamp:
-            fourth_order_divdamp_factor = self.solve_nonhydro._config.fourth_order_divdamp_factor
+            fourth_order_divdamp_factor = (
+                self.granules.solve_nonhydro._config.fourth_order_divdamp_factor
+            )
```
jcanton: from a friend: Accessing `self.granules.solve_nonhydro._config` reaches into a private attribute (pre-existing, not introduced here). Consider exposing it as a public property on `SolveNonhydro` or storing the config reference on `Icon4pyDriver` directly.

jcanton (commented):
from a friend — inline comments below

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

jcanton (commented):
from a friend:

Overall clean refactor. The `None`-based optional config pattern is much nicer than the `NoMicrophysics` sentinel hack. A few things worth addressing:

---

`advection.py:L18-21` — `AdvectionConfig` lost all field defaults. Was previously `NO_ADVECTION`/`NO_LIMITER` for all four. If any external code constructs `AdvectionConfig()` bare, it breaks. Confirm no remaining callers outside this PR scope.

`graupel.py:L102` — `enabled: bool = True` added to `SingleMomentSixClassIconGraupelConfig` but never checked. If graupel stays optional-without-enabled-check, remove the field to avoid confusion. If intent is consistency, wire it into `ExperimentConfig` lookup path.

`graupel.py:L150-151` — `KeyError` raised when namelists are absent. This path is dead (callers guard with `do_physics`) but a future caller hitting this gets a bare `KeyError: "nwp_phy_nml and nwp_tuning_nml required."` with no traceback hint. Use `ValueError` with the same message, or `KeyError("nwp_phy_nml")` from the dict access itself so it naturally includes the missing key.

`driver_utils.py:L426-428` — `_is_enabled` uses `getattr(cfg, name)` keyed by config field name strings. No type-checker coverage if a field is renamed. Extract a typed helper like `_diffusion_enabled(cfg) -> bool` per granule, or make the check inline. Same fragility applies to `validate_granule_state_consistency` calling convention — config field `"nonhydrostatic"` does not match Granules field `solve_nonhydro`, making the cross-reference non-obvious.

`driver_utils.py:L430-456` — `validate_granule_state_consistency` checks 3/4 optional configs (skips graupel). Add a brief docstring note explaining why graupel is excluded (it is not a time-integration granule).

`standalone_driver.py:L891, L930-931, L946, L975, L993, L1012` — Assert + comment inside already-guarded body is redundant. The `if self.granules.X is not None` guard already narrows the type for mypy; the assert adds no value at runtime. Either drop the `if` and keep only the assert (trusting the validator), or drop the assert and comment (trusting the guard). Six instances.

`config.py:L268-270` (and `datatest_utils.py:L1175-1177` — same pattern) — `ntracer=0` when advection is disabled. Previously read unconditionally from namelist. With `ntracer=0`, `prognostic_state.tracer` will have zero slots. If an initial condition tries to write tracers with advection off, this silently stores nothing rather than erroring. Confirm that all disabled-advection experiments truly have zero tracers in their initial condition.

`driver_states.py:L320-388` — `ModelTimeVariables` de-dataclassed (manual `__init__` + `_init_from_config`). Lose `__repr__`, `__eq__`, `__hash__`. If any code relies on these (e.g., comparing instances in tests), it silently falls back to object identity. Consider adding a `__repr__` or keeping `@dataclass` with `__init_hook__` pattern.

`test_parallel_standalone_driver.py:L1108` — Trailing period removed from log message. Inconsistent with the log message style in `standalone_driver.py:L879-882` which does have a trailing period. Either both or neither.

model/atmosphere/advection/src/icon4py/model/atmosphere/advection/advection.py
```diff
@@ -97,15 +97,16 @@ class VerticalAdvectionLimiter(Enum):
...
-    horizontal_advection_type: HorizontalAdvectionType = HorizontalAdvectionType.NO_ADVECTION
-    horizontal_advection_limiter: HorizontalAdvectionLimiter = HorizontalAdvectionLimiter.NO_LIMITER
-    vertical_advection_type: VerticalAdvectionType = VerticalAdvectionType.NO_ADVECTION
-    vertical_advection_limiter: VerticalAdvectionLimiter = VerticalAdvectionLimiter.NO_LIMITER
-
     """
     Contains necessary parameters to configure an advection run.
     """
 
+    horizontal_advection_type: HorizontalAdvectionType
```
jcanton, line 104: from a friend: 🟡 risk: AdvectionConfig lost all field defaults (were =NO_ADVECTION/NO_LIMITER, now required). Any external code calling AdvectionConfig() bare will crash. Confirm no callers outside this PR.

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -60,6 +59,8 @@ class SingleMomentSixClassIconGraupelConfig:
     The COSMO microphysics documentation "A Description of the Nonhydrostatic Regional COSMO-Model Part II Physical Parameterizations" can be downloaded via this link: https://www.cosmo-model.org/content/model/cosmo/coreDocumentation/cosmo_physics_6.00.pdf
     """
 
+    #: Master switch that enables/disables the graupel granule.
+    enabled: bool = True
```
jcanton: from a friend: 🔵 nit: enabled: bool = True added but never checked outside the config. No caller references config.graupel.enabled. Either wire it up or drop it.

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
```diff
@@ -95,46 +96,21 @@ def from_fortran_dict(
...
-
-    def __deepcopy__(self, memo: dict) -> NoMicrophysics:
-        return NoMicrophysics()
-
-
-class NoMicrophysicsError(Exception):
-    """Raise this if microphysics was not configured but the configuration for it is accessed anyway."""
-
-    ...
+        if nwp_phy_nml is None or nwp_tuning_nml is None:
```
jcanton: from a friend: 🔵 nit: KeyError raised when namelists absent. This path is dead (callers guard with do_physics) but if hit, KeyError("nwp_phy_nml…") is harder to debug than ValueError or letting the dict access itself raise naturally.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -61,6 +62,58 @@
...
 
 
+@dataclasses.dataclass
+class Granules:
+    diffusion: diffusion.Diffusion | None = None
+    solve_nonhydro: solve_nh.SolveNonhydro | None = None
+    tracer_advection: advection.Advection | None = None
+
+
+def validate_granule_state_consistency(
```
jcanton, line 72: from a friend: 🔵 nit: validate_granule_state_consistency checks 3/4 optional configs but skips graupel. Add docstring note explaining graupel is excluded (not a time-integration granule).

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -61,6 +62,58 @@
...
+    states: driver_states.DriverStates,
+) -> None:
+    """
+    Validate that enabled granules have their required states allocated.
+
+    Raises:
+        ValueError: if a granule is enabled but a state it requires is None.
+    """
+
+    def _is_enabled(cfg: driver_config.ExperimentConfig, name: str) -> bool:
```
jcanton: from a friend: 🟡 risk: _is_enabled uses getattr(cfg, name) keyed by raw strings. No type-checker coverage if a config field is renamed. Extract typed helpers (_diffusion_enabled, _nh_enabled, ...) or inline.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -148,7 +144,9 @@ def time_integration(
 
             self.model_time_variables.is_first_step_in_simulation = False
 
-            self._adjust_ndyn_substeps_var(solve_nonhydro_diagnostic_state)
+            if self.granules.solve_nonhydro is not None:
+                assert solve_nonhydro_diagnostic_state is not None
```
jcanton, line 148: from a friend: 🔵 nit: assert solve_nonhydro_diagnostic_state is not None inside an already-guarded body (if self.granules.solve_nonhydro is not None). The guard narrows the type for mypy; the assert adds no runtime value. Drop both the assert and the "Guaranteed by validate…" comment. Same pattern at L177, L193, L222, L269, L291, L416.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -181,6 +188,9 @@ def read_config(
         atm_dict=atm_dict,
         master_dict=master_dict,
         profiling_stats=profiling_stats,
+        ntracer=fortran_config.list_to_value(atm_dict["run_nml"]["ntracer"])
```
jcanton, line 193: from a friend: ❓ q: ntracer=0 when tracer_advection disabled. Previously read unconditionally from namelist. With ntracer=0, prognostic_state.tracer has zero slots. If an initial condition writes tracers with advection off, data silently disappears. Confirm all disabled-advection experiments have zero tracers in their initial condition.

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -59,52 +58,54 @@ class DriverStates(NamedTuple):
...
+    solve_nonhydro_diagnostic: dycore_states.DiagnosticStateNonHydro | None
+    diffusion_diagnostic: diffusion_states.DiffusionDiagnosticState | None
+    tracer_advection_diagnostic: advection_states.AdvectionDiagnosticState | None
+    prep_tracer_advection_prognostic: advection_states.AdvectionPrepAdvState | None
     prognostics: common_utils.TimeStepPair[prognostics.PrognosticState]
     diagnostic: diagnostics.DiagnosticState
 
 
-@dataclasses.dataclass
 class ModelTimeVariables:
```
jcanton, line 76: from a friend: 🔵 nit: ModelTimeVariables de-dataclassed (manual __init__ + _init_from_config). Lose __repr__, __eq__, __hash__. If anything compares instances or prints them in tests/debug logs, it silently falls back to object identity. Add __repr__ or keep @dataclass with __init_hook__.

jcanton:
@msimberg why can I squash and merge without approvals?

model/atmosphere/advection/src/icon4py/model/atmosphere/advection/advection.py
jcanton: that's intentional and restores the previous behaviour: we don't want defaults because they risk hiding bugs

model/atmosphere/subgrid_scale_physics/microphysics/src/icon4py/model/atmosphere/subgrid_scale_physics/microphysics/single_moment_six_class_gscp_graupel.py
jcanton: @friend: graupel in its entirety is not used

maybe we should make a granule Protocol of some sort where to move these options?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
jcanton: this is being fixed by https://github.com/C2SM/icon4py/pull/1317

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: added repr

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
jcanton: done

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
jcanton: good point, inlining

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton: fixed and removed unnecessary `enabled`s

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/atmosphere/advection/src/icon4py/model/atmosphere/advection/advection.py
```diff
@@ -97,15 +97,15 @@ class VerticalAdvectionLimiter(Enum):
...
+    horizontal_advection_limiter: HorizontalAdvectionLimiter
```
msimberg, line 105: Note from earlier discussion with @jcanton: we could consider bundling these into `HorizontalAdvectionConfig`, likewise for vertical config. So that `AdvectionConfig` is just composed from `HorizontalAdvectionConfig` and `VerticalAdvectionConfig`. Not for this PR.

model/atmosphere/dycore/tests/dycore/mpi_tests/test_parallel_solve_nonhydro.py
```diff
@@ -92,6 +92,8 @@ def test_run_solve_nonhydro_single_step(  # noqa: PLR0917 [too-many-positional-a
     )
 
     config = experiment.config.nonhydrostatic
+    assert config is not None
```
msimberg: Maybe I missed it: why can experiment.config be None now?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -153,24 +152,32 @@ def read_config(
...
         max_nudging_coefficient=interpolation_config.max_nudging_coefficient,
     )
 
-    graupel_config = graupel.SingleMomentSixClassIconGraupelConfig.from_fortran_dict(atm_dict)
+    do_tracer_advection = not (
+        "exclaim_ch_r04b09_dsl" in config_file_path.name
+        or "exclaim_ape_R02B04" in config_file_path.name
+    )
+    # The experiments above were run in fortran with a tracer advection scheme
+    # that has not been ported to ICON4Py and can not be used for testing.
```
msimberg, line 165: I think this needs a TODO. For a special case this isn't the right place to keep it, so the TODO is either for fixing this or moving this to a better place.

msimberg (commented):
Just a few last comments, possibly nothing to change. Looks good to me overall.

model/atmosphere/dycore/tests/dycore/mpi_tests/test_parallel_solve_nonhydro.py
jcanton: it's `nonhydrostatic` that can be None, or?

jcanton: maybe not consistent with how `config.diffusion` is handled just below, I'll make it consistent

jcanton: and actually the `config.diffusion` was used inappropriately, `config.driver` is more appropriate for getting `ndyn_substeps` (though they're actually duplicates so it was technically ok)

msimberg: I misread the line above as `config = experiment.config` and wondered why the whole config can be None, but of course the dycore config can be None.

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

msimberg (approved):

PR #1315, opened 2026-06-12
fix(testing): auto-download datatest data via autouse fixture

### LINKED ISSUE #1314: experiment/experiment_description fixtures skip download_ser_data when overridden in tests
## Issue

The `experiment` and `experiment_description` fixtures only trigger `download_ser_data` when they are **not** overridden by re-defined fixtures in individual tests. Tests that redefine these fixtures silently skip downloading serialized data, which can lead to false positives or incorrect test behavior.

jcanton:
All datatest tests now automatically download serialized data via an autouse fixture in pytest_hooks, regardless of how experiment_description or process_props are parametrized.

Previously, tests that parametrized experiment_description directly (bypassing the experiment fixture) would skip data download because the fixture function body with the download_ser_data dependency was never called when overridden via pytest.mark.parametrize.

Closes #1314

### DIFF
--- a/model/testing/src/icon4py/model/testing/pytest_hooks.py
+++ b/model/testing/src/icon4py/model/testing/pytest_hooks.py
@@ -13,7 +13,7 @@ import numpy as np
 import pytest
 
 from icon4py.model.common import model_backends
-from icon4py.model.testing import filters
+from icon4py.model.testing import datatest_utils, filters
 
 
 __all__ = [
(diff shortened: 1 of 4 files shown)


jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -256,3 +256,19 @@ def pytest_terminal_summary(terminalreporter, exitstatus, config):
                 f"{benchmark_name:<{max_name_len}} | {np.mean(gtx_metrics):>10.8f} | {np.median(gtx_metrics):>10.8f} | {np.std(gtx_metrics):>10.8f} | {len(gtx_metrics):>4}"
             )
         terminalreporter.line("-" * len(header), blue=True)
+
+
+@pytest.fixture(autouse=True)
+def auto_download_ser_data(request: pytest.FixtureRequest) -> None:
```
havogt, line 262: doesn't this imply that download is even triggered if you don't do datatests?

jcanton: should't the `if () -> return`s below avoid that?

jcanton:
not the right way of going about this

PR #1317, opened 2026-06-12
Use the tracer state that was written time ago

jcanton:
and isolate the fortran-inherited list approach.
`TracerState` was implemented but not used, and a list-based approach was instead used in standalone driver, with index-based identification of tracers.
This fixes that, hopefully before the physics interface goes through so it can be directly used there as well


model/common/src/icon4py/model/common/states/data.py
```diff
@@ -50,43 +50,57 @@
...
 
+#: index of tracer variables in the serialized fortran list
+QV: Final[int] = 0
+QC: Final[int] = 1
+QI: Final[int] = 2
+QR: Final[int] = 3
+QS: Final[int] = 4
+QG: Final[int] = 5
+
 #: CF attributes of common tracer variables
```
jcanton, line 62: from a friend: renaming the keys of `COMMON_TRACER_CF_ATTRIBUTES` from descriptive names (`specific_humidity`, etc.) to tracer codes (`qv`, etc.) changes a public API. if anything downstream indexes this dict by the old names, it will break. worth calling out explicitly in the PR description or keeping aliases.

model/common/src/icon4py/model/common/states/tracer_state.py
```diff
@@ -7,9 +7,74 @@
...
+
+@dataclasses.dataclass(frozen=True)
+class TracerConfig:
+    """Which tracers are active in the model configuration.
+
+    Each boolean field indicates whether the corresponding tracer is active.
+    Used instead of a raw ``ntracer: int`` to provide type-safe tracer selection.
+    """
+
+    qv: bool = False
```
jcanton, line 43: from a friend: the dataclass field order is `qv, qc, qr, qi, qs, qg`, but `_TRACER_FIELDS` sorts by Fortran index to `qv, qc, qi, qr, qs, qg`. this makes `__repr__` / `asdict` order differ from the canonical tracer order. align the field order with the index order to avoid confusion.

model/common/src/icon4py/model/common/states/tracer_state.py
```diff
@@ -7,9 +7,74 @@
...
+    @classmethod
+    def all(cls) -> TracerConfig:
+        return cls(qv=True, qc=True, qr=True, qi=True, qs=True, qg=True)
+
+    @classmethod
+    def none(cls) -> TracerConfig:
+        return cls()
+
+    @classmethod
+    def from_ntracer(cls, ntracer: int) -> TracerConfig:
```
jcanton, line 59: from a friend: `from_ntracer` silently treats negative values as `none` and values > 6 as `all`. consider validating the input or documenting that this is intentional.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -65,14 +66,9 @@ class DriverConfig:
     vertical_cfl_threshold: ta.wpfloat = dataclasses.field(default_factory=lambda: ta.wpfloat(0.85))
     ndyn_substeps: int = 5
     enable_statistics_output: bool = False
-    ntracer: int = 0
     do_tracer_advection: bool = True
```
jcanton: from a friend: the old `DriverConfig.__post_init__` forced `ntracer = 0` when `do_tracer_advection = False`. that guard is gone now, so a run with advection disabled but `ntracer > 0` will still allocate and read tracer fields. was dropping this behavior intentional?

model/common/src/icon4py/model/common/states/tracer_state.py
```diff
@@ -20,14 +85,31 @@ class TracerState:
...
+    qg: fa.CellKField[ta.wpfloat] | None = None
+
+    def active_fields(self) -> Iterator[tuple[str, fa.CellKField[ta.wpfloat]]]:
+        """Yield ``(name, field)`` for each non-``None`` tracer field."""
+        for name in _TRACER_FIELDS:
+            field = getattr(self, name)
+            if field is not None:
+                yield name, field
+
+    def active_pairs(
```
jcanton: from a friend: `active_pairs` silently skips tracers that are active in only one of the two states. since `current` and `next` should always share the same `TracerConfig`, a mismatch is almost certainly a bug; consider raising instead of silently dropping.

jcanton (commented):
from a friend: overall this is a nice cleanup. a few questions/notes below, mostly around behavior changes from dropping the old `ntracer` guard and the public API surface of `COMMON_TRACER_CF_ATTRIBUTES`.

model/common/src/icon4py/model/common/states/data.py
jcanton: the dict is not used anywhere yet, so we're free to do as we please with the keys. however, what is the best approach?
a. use descriptive names and add a "icon4py_var_name" to the entries and then we index from there?
b. keep the current version which is easier for indexing but uses a non-descriptive key?

jcanton: from a friend:
go with b, but drop the redundant icon_var_name field.
The dict key should be the canonical identifier used in code, because that is what TracerConfig, TracerState, and the Fortran variable names all agree on. Using qv/qc/etc. as keys lets you do things like:
```
for name in COMMON_TRACER_CF_ATTRIBUTES:
    setattr(tracer_state, name, field)
```
without any extra mapping layer. The descriptive information already lives where it belongs—in standard_name and long_name.
Option a forces every consumer to carry two names for the same thing: a descriptive dict key and a code-facing variable name. That adds indirection for no real gain, especially since the dict is not user-facing metadata; it is code-facing configuration.
One small cleanup while you are at it: since the key is now the variable name, icon_var_name in each entry duplicates the key. Remove it unless something else relies on FieldMetaData.icon_var_name generically. If it is used generically, rename it to icon_var_name and keep it, but the key should still be the code-facing name.
So my recommendation:
```
COMMON_TRACER_CF_ATTRIBUTES: Final[dict[str, model.FieldMetaData]] = dict(
    qv=dict(
        standard_name="specific_humidity",
        long_name="ratio of water vapor mass to total moist air parcel mass",
        units="1",
        icon_var_list_index=QV,
    ),
    ...
)
```

jcanton: we keep `icon_var_name` because that doesn't always correspond to the `icon4py_var_name`

I'd suggest we use `icon4py_var_name` as dict key, then have `standard_name` as descriptive name (plus `long_name`) and `icon_var_name` for compatibility with fortran (which should be used by serialbox now that I think about it)

model/common/src/icon4py/model/common/states/data.py
```diff
@@ -50,43 +50,57 @@
...
 COMMON_TRACER_CF_ATTRIBUTES: Final[dict[str, model.FieldMetaData]] = dict(
-    specific_humidity=dict(
+    qv=dict(
```
havogt, line 64: Why move away from standard name?

model/common/src/icon4py/model/common/states/data.py
```diff
@@ -50,43 +50,57 @@
     ),
 )
 
+#: index of tracer variables in the serialized fortran list
+QV: Final[int] = 0
+QC: Final[int] = 1
+QI: Final[int] = 2
+QR: Final[int] = 3
+QS: Final[int] = 4
+QG: Final[int] = 5
```
havogt, line 60: As you know from the Tuesday discussion. To me the better location is where it was before...

model/common/src/icon4py/model/common/states/data.py
jcanton: because we discussed on tuesday that we use the icon4py names as keys?
then the dicts need only specify the standard name in FieldMetaData and you avoid duplication

model/common/src/icon4py/model/common/states/data.py
jcanton: I can move this, but I'd prefer it stays in the same location where the fortran names are defined, not separate.
should I find a way to move both?
maybe out-of-scope for this PR and better left for a separate one where we collect all fortran stuff somewhere?

model/common/src/icon4py/model/common/states/data.py
jcanton: see also this discussion (above) https://github.com/C2SM/icon4py/pull/1317#discussion_r3418591896

havogt: But we control icon4py names, so why not change it?

jcanton: are you suggesting we change all icon4py names to standard names?

model/common/src/icon4py/model/common/states/tracer_state.py
```diff
@@ -7,9 +7,74 @@
...
+from collections.abc import Iterator
 
 from icon4py.model.common import field_type_aliases as fa, type_alias as ta
+from icon4py.model.common.states.data import COMMON_TRACER_CF_ATTRIBUTES
+
+
+_TRACER_FIELDS: tuple[str, ...] = tuple(
+    sorted(
+        COMMON_TRACER_CF_ATTRIBUTES,
+        key=lambda k: COMMON_TRACER_CF_ATTRIBUTES[k].get("icon_var_list_index", 0),  # type: ignore[arg-type]
```
msimberg: Why fall back to `0` in `get`? Is it intentional for tracers that are not defined in icon fortran so that they get sorted first (or together with the zeroth icon fortran tracer)?

model/common/src/icon4py/model/common/states/prognostic_state.py
```diff
@@ -47,9 +48,11 @@ def w_1(self) -> fa.CellField[ta.wpfloat]:
 def initialize_prognostic_state(
     grid: icon_grid.IconGrid,
     allocator: gtx_typing.Allocator,
-    ntracer: int = 0,
+    tracer_config: TracerConfig | None = None,
 ) -> PrognosticState:
     """Initialize the prognostic state with zero fields."""
+    if tracer_config is None:
+        tracer_config = TracerConfig.none()
```
msimberg, line 55: I think I get why you do this, but why not just check if it's `None` further down and skip the `TracerState` construction (let it be `None` like for other granules)? In isolation I like this and perhaps this is the nicer solution because it avoids branching later on, just want to check if this was a conscious choice.

model/common/src/icon4py/model/common/states/tracer_state.py
```diff
@@ -20,14 +85,31 @@ class TracerState:
...
+
+    def active_pairs(
+        self, new: TracerState
+    ) -> Iterator[tuple[str, fa.CellKField[ta.wpfloat], fa.CellKField[ta.wpfloat]]]:
+        """Yield ``(name, now_field, new_field)`` for each non-``None`` tracer in both states."""
+        for name in _TRACER_FIELDS:
+            self_field = getattr(self, name)
+            new_field = getattr(new, name)
+            if self_field is not None and new_field is not None:
+                yield name, self_field, new_field
```
msimberg: I think this would benefit from a little helper dataclass with named fields (or even a little dict), instead of returning a tuple. I think same above for `active_fields`.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/from_file.py
```diff
@@ -90,8 +94,8 @@ def read_edge_k(name: str) -> data_alloc.NDArray:
 
     if ntracer > 0:
         tracers_raw = array_ns.squeeze(ser.read("tracers_now", sp).astype(float))
-        for i in range(ntracer):
-            state.tracer[i].ndarray[:, :] = array_ns.asarray(tracers_raw[:num_cells, :, i])
+        for i, (_, field) in enumerate(state.tracer.active_fields()):
```
msimberg: Related to https://github.com/C2SM/icon4py/pull/1317/changes#r3452241895. If `active_fields` returned a dict this could instead be:

```suggestion
        for i, field in enumerate(state.tracer.active_fields().values()):
```
which in my opinion would be clearer.

model/common/src/icon4py/model/common/states/tracer_state.py
msimberg: Possibly this is even a `TimeStepPair`: https://github.com/C2SM/icon4py/blob/fd4e2406a8d9546949c5af1ec0df1a7a7c3ee070/model/common/src/icon4py/model/common/utils/_common.py#L236-L238?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -110,6 +111,7 @@ class ExperimentConfig:
     driver: DriverConfig
     nonhydrostatic: solve_nh.NonHydrostaticConfig | None = None
     diffusion: diffusion.DiffusionConfig | None = None
+    tracer_config: TracerConfig = dataclasses.field(default_factory=TracerConfig.none)
```
msimberg: I think I see what you're going for here, but I find this looks out of place next to the other configs with `None`. Are you now assuming that `TracerConfig` is never `None` after setup? Why use a different mechanism here for "tracers disabled" than for other configs?

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -533,7 +537,7 @@ def display_driver_setup_in_log_file(
     log.info(f"Vertical CFL threshold : {config.vertical_cfl_threshold}")
     log.info(f"Second-order divdamp   : {config.apply_extra_second_order_divdamp}")
     log.info(f"Statistics enabled     : {config.enable_statistics_output}")
-    log.info(f"Number of tracers      : {config.ntracer}")
+    log.info(f"Active tracers         : {tracer_config.active_names or 'none'}")
```
msimberg: Can the output here instead be put into `__str__`?

model/common/src/icon4py/model/common/states/tracer_state.py
jcanton: bad design, erroring out instead of falling back

model/common/src/icon4py/model/common/states/prognostic_state.py
jcanton: I think you caught friend, who answers:
> I prefer the current approach. Making the TracerState construction unconditional (after normalizing None at the entry) is simpler — one path through the function, no hidden field default to reason about. The None in the parameter type is just "caller doesn't care about tracers", and normalizing it immediately makes the rest of the function uniformly typed. It's a minor style choice either way, so if the reviewer prefers the conditional-skip, not worth fighting over.

I'm not sure I understand what's your preference

model/common/src/icon4py/model/common/states/tracer_state.py
jcanton: good point

jcanton: actually this was a bad nested thing: PrognosticState already takes care of .current and .next
no need for TracerState to duplicate this functionality, since TracerState is a member of PrognosticState

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/from_file.py
jcanton: should be better now?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
jcanton: caught again, changed ;-)

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/common/src/icon4py/model/common/states/data.py
msimberg: I don't know @havogt's intention, but my 1 cent: I don't care much if the icon4py name is _the_ standard name or not, but I definitely like a longer name like `specific_humidity` more than `qv`.

model/common/src/icon4py/model/common/states/prognostic_state.py
msimberg: > I'm not sure I understand what's your preference

I'm not sure I know my preference ;)

But I agree with your friend about the reasoning, I wanted to understand if that was indeed the motivation. Keep it like this, all good. Thanks for the clarification.

model/testing/src/icon4py/model/testing/datatest_utils.py
```diff
@@ -215,9 +216,10 @@ def create_experiment_configuration(
         master_dict=master_dict,
         profiling_stats=None,
         enable_statistics_output=False,
-        ntracer=fortran_config.list_to_value(atm_dict["run_nml"]["ntracer"])
-        if do_tracer_advection
-        else 0,
+    )
+
+    tracer_config = TracerConfig.from_ntracer(
+        fortran_config.list_to_value(atm_dict["run_nml"].get("ntracer", 0))
```
msimberg: I might have missed this on the first round, but is this equivalent to what it was before? If `ntracer` is now accidentally missing from `run_nml` it'll silently disable tracers? Perhaps correct, just checking if intentional?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -186,22 +187,25 @@ def read_config(
     )
 
     profiling_stats = ProfilingStats() if enable_profiling else None
+    ntracer = (
+        fortran_config.list_to_value(atm_dict["run_nml"]["ntracer"]) if do_tracer_advection else 0
```
msimberg: Related, because it's done differently to here: https://github.com/C2SM/icon4py/pull/1317/changes#r3457885942. Perhaps I'm missing a good reason why it's done differently?

model/common/src/icon4py/model/common/states/tracer_state.py
msimberg: If this is actually a requirement (i.e. that the members are sorted the same way they are in fortran) then this really needs at the very least a comment, though ideally it should be made so it doesn't matter...

model/testing/src/icon4py/model/testing/datatest_utils.py
jcanton: possibly I should not use `.get` but `[]` so it fails: `ntracer` should never be missing from `run_nml`

jcanton: yeah, exactly as below: I think I changed it in one place but not both

model/common/src/icon4py/model/common/states/tracer_state.py
jcanton: I fixed and added a comment about it, it's not technically required, just a nice to have to use one less helper so we sort by fortran ordering

model/common/src/icon4py/model/common/states/data.py
jcanton: I have (internally) conflicting opinions about this: descriptive names are nice but variable names used on papers/maths/equations also have their purpose...

msimberg: Is something like `q_humidity`, `q_ice` etc. a good or a bad compromise name? If we can't agree, let's leave it. We can bikeshed about this anytime.

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/common/src/icon4py/model/common/states/tracer_state.py
jcanton: from a friend:
>The short answers:
Field order in TracerState the dataclass is not functionally important. No code constructs it positionally, and active_fields() always iterates _TRACER_FIELDS, not dataclasses.fields(). So swapping qi and qr in the class body would break nothing.
_TRACER_FIELDS ordering is important — it's derived from Fortran icon_var_list_index and governs all iteration: TracerConfig.from_ntracer(), active_names, active_fields, and __iter__. Any code that processes tracers in sequence (e.g. mapping ntracer → which tracers are active) relies on this order.
That said: the dataclass field order already matches _TRACER_FIELDS (qv, qc, qi, qr, qs, qg). Keeping them in sync is a good convention for maintainability, even though no runtime code requires it. If someone adds a new tracer, they should add it in the corresponding position in both TracerState and TracerConfig, matching the new entry in COMMON_TRACER_CF_ATTRIBUTES's icon_var_list_index.

jcanton: so I think it makes sense that the comment is close to where `_TRACER_FIELDS` is defined, not where it's used, right?

msimberg: Thanks for checking, this makes sense. No comment needed then.

msimberg (approved):

PR #1319, opened 2026-06-15
Declarative diffusion config

DropD:
Make DiffusionConfig into a data class and introduce a declarative way to describe config options in code.

The aim is to use config option annotations to code share mechanisms (between config classes) for:
- mapping ICON config options to Icon4Py
- (optionally, later) move ICON -> Icon4Py mapping to a separate translation layer
- easing the transition to consistent use of descriptive names for config options in Icon4Py

Potential extensions / future use cases could be:
- generating annotated config templates (with help comments)
- validating configs
- in general: tooling around authoring, inspecting and otherwise working with Icon4Py configurations

Still missing:
- [x] ADR about making config classes as declarative as possible


DropD:
cscs-ci run default

DropD:
cscs-ci run distributed

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion_utils.py
```diff
@@ -21,7 +21,7 @@ def _identity_c_k(field: fa.CellKField[float]) -> fa.CellKField[float]:
 
 
 @gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
-def copy_field(old_f: fa.CellKField[float], new_f: fa.CellKField[float]):
+def copy_field(old_f: fa.CellKField[float], new_f: fa.CellKField[float]) -> None:
```
jcanton, line 24: see https://github.com/C2SM/icon4py/issues/1332
these guys could/should be moved in a cleanup PR.
just noting here that we noticed and will do something about it ;-)

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+---
+
+# [Configuration Classes]
+
+- **Status**: valid
+- **Authors**: Rico Häuselmann (@DropD)
+- **Created**: 2026-06-17
+- **Updated**: 2026-06-17
+
+(Wh)Y-statement: In the context of [use case/user story u], facing [concern c] we decided for [option o] and neglected [other options], to achieve [system qualities/desired consequences], accepting [downside/undesired consequences].
```
jcanton: should this line be deleteed?

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+While making Icon4Py user-configurable, facing code duplication between modules, we decided to provide a standard way to declaratively encode common attributes of configuration options alongside utilities that can then be reused across modules.
+
+## Context
+
+We notice that currently, each module has a class, which contains it's configuration. Each of these classes knows how to construct itself from a dictionary that represents an ICON namelist. The mapping is contained implicitly in a classmethod. Usually the class attributes that have an equivalent in ICON are annotated with documentation comments (where to find in ICON namelists and what it does).
```
jcanton:
```suggestion
We notice that, currently, each module has a class which contains its configuration. Each of these classes knows how to construct itself from a dictionary that represents an ICON namelist. The mapping is contained implicitly in a classmethod. Usually the class attributes that have an equivalent in ICON are annotated with documentation comments (where to find in ICON namelists and what it does).
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Really, given the mapping for each configuration option in each of those classes, the logic is the same for all of them.
```
jcanton:
```suggestion
Given the mapping for each configuration option in each of those classes, the logic is the same for all of them.
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Extrapolating from this case of duplication we find the same for:
+
+- describing configuration options to users (currently only documentation comments exist for this)
+- validation (currently partially done in monolithic, imperative validation methods)
+
+## Decision
+
+The chosen pattern is for configuration classes to look as Follows
```
jcanton:
```suggestion
The chosen pattern is for configuration classes to look as follows
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+from icon4py.model.common.config import options as common_conf_opt
+
+
+@dataclasses.dataclass
+class MyModuleConfig:
+    some_configuration_option: typing.Annotated[
+        int,
+        common_conf_opt.ConfigOption(
+            description="This is help telling the user what `some_configuration_option` does.",
```
jcanton:
```suggestion
            description="Describe what `some_configuration_option` does.",
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+            icon_equivalent=common_conf_opt.IconOption(
+                name="isomcfgop",
+                path=("diffusion_nml",),  # for nested sections: ("toplevel_nml", "sublevel_nml")
+                read_from_icon=True,  # optional (default: True), whether to actually read this from ICON namelists
+                list_to_value=False,  # optional (default: False), whether to map a list of values in ICON namelists to a single value
+            ),
```
jcanton, line 47: I love this, especially the **very**, **very** realistic icon variable name 😍

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+            ...,
+            validators=(
+                DependsOn("some_flag", condition=validate_some_configuration_option_with_some_flag),
+                Min(0),
+                Max(59000),
+                ...,
+            ),
+        ),
+    ] = 4
+```
```
jcanton, line 81: This is also very nice

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Consequences
+
+Code for mapping from ICON inputs to Icon4Py configurations can be shared between modules.
+
+It becomes possible to automatically provide helpful (for users) information for all available configuration options (for a given set of modules).
```
jcanton:
```suggestion
It becomes possible to automatically provide helpful information for all available configuration options.
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Different user facing utilities for configuring / inspecting Icon4Py runs will not have to duplicate logic for
```
jcanton:
```suggestion
Different user facing utilities for configuring / inspecting ICON4Py runs will not have to duplicate logic for
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+- mapping to / from ICON (fortran)
+- validation
+- documentation
+- etc
+
+The current solution will not achieve maximum separation isolation between Icon4Py internals and their mapping to and from ICON concepts. However it provides a structured path towards it.
```
jcanton:
```suggestion
The current solution will not achieve maximum separation between ICON4Py internals and their mapping to and from ICON concepts. However, it provides a structured path towards it.
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+Another downside is a certain loss of type checking when using the provided utilities. This can be made up for in tests.
```
jcanton: can you specify what you mean with "a certain loss of type checking" and / or why?

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+
+## Alternatives Considered
+
+### Separating (fortran-) ICON mapping information from the rest
+
+- Pro: allow free evolution of Icon4Py internals, decoupled from ICON
```
jcanton:
```suggestion
- Pro: allows for free evolution of ICON4Py internals, decoupled from ICON
```

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,126 @@
...
+- Con: more complicated, less (for now) helpful information immediately available
+
+This could be achieved, however, by simply moving all the instances of `IconOption` out of `ConfigOption` annotations and into a separate (per-module) dictionary, keyed by the name (or class AND name) of the option it belongs to.
```
jcanton:
```suggestion
This could be achieved, however, by moving all the instances of `IconOption` out of `ConfigOption` annotations and into a separate (per-module) dictionary, keyed by the name (or class AND name) of the option it belongs to.
```

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
+        common_conf_opt.ConfigOption(
+            description="If True, apply diffusion to the vertical wind field.",
+            icon_equivalent=common_conf_opt.IconOption("lhdiff_w", ("diffusion_nml",)),
+        ),
+    ] = True
+
+    apply_to_horizontal_wind: typing.Annotated[
+        bool,
+        common_conf_opt.ConfigOption(
+            description="If true, apply diffusion on the horizontal wind field, is ONLY used in mo_nh_stepping.f90.",
```
jcanton:
```suggestion
            description="If true, apply diffusion on the horizontal wind field.",
```

jcanton: I don't know who added that,  but I find it not helpful.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
+            icon_equivalent=common_conf_opt.IconOption(
+                "lsmag_3d", ("diffusion_nml",), list_to_value=True
+            ),
+        ),
+    ] = False
+
+    type_vn_diffu: typing.Annotated[
+        SmagorinskyStencilType,
+        common_conf_opt.ConfigOption(
+            description="Options for discretizing the Smagorinsky momentum diffusion.",
```
jcanton:
```suggestion
            description="Reconstruction method used for Smagorinsky diffusion.",
```

I took this from `icon-nwp/doc/www/documentation/buildrun/buildrun_namelist_overview.md` and I find it more helpful than the previous comment/description.
Can you take all from there? (or send an LLM to do so?)

jcanton: I think the one-liner description is the appropriate.
The allowed values and their respective descriptions, i.e.,
```
    - 1: u/v reconstruction at vertices only
    - 2: u/v reconstruction at cells and vertices
```
should not belong here but rather in `SmagorinskyStencilType`, right?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
+    max_nudging_coefficient: typing.Annotated[
+        float,
+        common_conf_opt.ConfigOption(
+            description=(
+                "Parameter describing the lateral boundary nudging in limited area mode."
+                "Maximal value of the nudging coefficients used cell row bordering the boundary interpolation zone,"
+                "from there nudging coefficients decay exponentially with `nudge_efold_width` in units of cell rows."
+            ),
+            icon_equivalent=common_conf_opt.IconOption(
+                name="nudge_max_coeff", path=("interpol_nml",), read_from_icon=False
```
jcanton, line 356: I assume you set `read_from_icon=False` on this one because of my `from_fortran_dict`.

This one is "tricky" and may benefit from voice discussion, but the jist is:
- it's triplicated beween `InterpolationConfig`, `NonHydrostaticConfig` and `DiffusionConfig`
- the fortran namelist value is internally multiplicated by `DEFAULT_DYNAMICS_TO_PHYSICS_TIMESTEP_RATIO`, so it would need some kind of `scale_icon_equivalent_value_by()` method or something like that if enabling `read_from_icon` here

jcanton: for now this is taken care of by the duplicated config readers in `testing/datatest_utils.py :: create_experiment_configuration` and `standalone_driver/config.py :: read_config` which read `InterpolationConfig` first, which compute the derived `max_nudging_coefficient` value internally in `__post_init__` and then pass it to `NonHydrostaticConfig` and `DiffusionConfig` so this is safe and can probably be left for a followup cleanup?

jcanton (commented):
This looks great.
Mostly cosmetic (some optional?) comments on my side

bindings/src/icon4py/bindings/diffusion_wrapper.py
```diff
@@ -117,13 +120,13 @@ def diffusion_init(  # noqa: PLR0917 [too-many-positional-arguments]
...
     # Diffusion parameters
     config = DiffusionConfig(
-        diffusion_type=diffusion_type,
-        hdiff_w=hdiff_w,
-        hdiff_vn=hdiff_vn,
-        hdiff_smag_w=hdiff_smag_w,
-        zdiffu_t=zdiffu_t,
-        type_t_diffu=type_t_diffu,
-        type_vn_diffu=type_vn_diffu,
+        diffusion_type=DiffusionType(diffusion_type),
```
msimberg, line 123: Do we need to consider the following:

In principle this is a fortran -> python mapping, and we are currently tying the icon4py enums to be exactly the same (values) as those in fortran. This means if we add new types etc. they always have to take the fortran values into account. Not particularly difficult, but also not particularly pretty in the context of the discussion that we should separate the fortran <-> python mappings better.

This may be a big separate follow up task, but perhaps you can at least have a look if it would be useful to do now, or how feasible/invasive such a change would be?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
+        ),
+    ] = False
 
-        #: Output flag for horizontal shear
-        #: Called 'loutshs' in mo_turbdiff_nml.f90
-        #: not a namelist parameter: its default is FALSE and only set to true in fortran `IF (.NOT. ldynamics)`
-        self.loutshs: bool = loutshs
+    def __post_init__(self) -> None:
 
         self._validate()
```
msimberg, line 399: Do you think calling the validation would make sense to put in a config base class? Such that every config dataclass has to implement `_validate`, but one can't forget to call `_validate`?

model/common/src/icon4py/model/common/config/options.py
```diff
@@ -0,0 +1,93 @@
...
+            data = icon_config
+            for subsection in opt.icon_equivalent.path:
+                data = data[subsection]
+                raw_value = data[opt.icon_equivalent.name]
+                de_listified = (
+                    fortran_config.list_to_value(raw_value)
+                    if opt.icon_equivalent.list_to_value
+                    else raw_value
+                )
+                yield name, annotations[name](de_listified)
```
msimberg: Should these maybe also be private (`_foo`) for now?

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
-        #: Scaling factor for horizontal shear production term
-        #: Called 'a_hshr' in mo_turbdiff_nml.f90
-        self.a_hshr: float = a_hshr
+    diffusion_type: typing.Annotated[
+        DiffusionType,
+        common_conf_opt.ConfigOption(
+            description="Order of Nabla operator for diffusion.",
+            icon_equivalent=common_conf_opt.IconOption("hdiff_order", ("diffusion_nml",)),
+        ),
+    ] = DiffusionType.SMAGORINSKY_4TH_ORDER
```
msimberg, line 157: Out of ignorance, what's the technical difference to using `typing.Annotated` here instead of making `diffusion_type` itself have something like a `ConfigOption` type?

msimberg (commented):
Minor comments and questions. PR description already notes the follow ups to this and I'm happy if we can move towards that instead.

bindings/src/icon4py/bindings/diffusion_wrapper.py
DropD: Could you clarify a bit more in which direction you would like this change to go? I can see multiple ways of doing this separation, for more or less effort (and I would like to add a general direction to the ADR):

- encode the possible values in an annotation instead (something like 
  ```python
   Annotated[
        int, 
        ImplentedIconChoices((1, "meaning of the value 1"), ...), 
        NotImplementedIconChoices(...), 
        Icon4PyOnlyChoices(...))
   ] 
   ```
  I realize these could also be represented by enums, but for clarity I chose something more explicit above.

- keep valid values in an enum (without any information as to whether they relate to (fortran) ICON and encode the relationships in annotations (again, for now, to be moved to a translation layer at a later point), optionally with annotations for valid ICON values which do not have an equivalent in Icon4Py. Example:
  ```python
  class DiffusionType(int, enum.Enum):
      implemented_in_icon4py = 1

  Annotated[
      DiffusionType,
      IconEnumMapping(("implemented_in_icon4py", "ICON help text for value '1'"), ..., (42, "Not implemented in Icon4Py"))
  ]
  ```

- use a general validation mechanism to catch ICON values not mapped to Icon4Py (less to build but harder to later split out into translation layer), something like `BlacklistedValue(42, "Not implemented.")`

docs/development/ADRs/0002_configuration_classes.md
DropD: Yes, thanks!

docs/development/ADRs/0002_configuration_classes.md
DropD: Thanks, I will correct the `it's` typo (I checked and none of the commas in either of our versions is actually appropriate here).

docs/development/ADRs/0002_configuration_classes.md
DropD: This is basically copied from Pydantic / Django / SqlAlchemy etc, no need to reinvent the wheel.

docs/development/ADRs/0002_configuration_classes.md
DropD: Replaced throughout

bindings/src/icon4py/bindings/diffusion_wrapper.py
DropD: Or more along the lines of using enums that don't care about fortran ICON (as in they can use `enum.auto()` values)? That could looks like

```python
class DiffusionType(int, enum.Enum):
    SMAGORINKSY_4TH_ORDER = enum.auto()


def unkown_icon_diffusion_type(icon_option: IconOption, value: int) -> typing.Never:
    raise NotImplementedValue(f"value {icon_diffusion_type} for "{icon_option.name} is not supported in ICON4Py.")

Annotated[
    DiffusionType,
    IconEnumMapping({
        5: {"value": DiffusionType.SMAGORINSKY_4TH_ORDER}, 
        "_": {"factory": unknown_icon_diffusion_type}
    }),
    ConfigOption(...)
```

This could be quite simple. If the reaction to an unaccounted-for value is always the same, then some of the above could be moved into `common` as well.

As with the options above, I envision a separate constructor that applies the mapping, which removes the concern from the call site in the bindings wrapper:

```python
DiffusionConfig.map_from_icon_fortran(
    diffusion_type=diffusion_type,
    ...
)
```

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: 1. no special casing required for the `common.config.options` functionality (treating builtin types vs enums)
2. all the mapping data in one class vs distributed over config class and associated types
3. more separation between internals and the mapping `ICON -> ICON4Py`

For the first I can imagine design solutions (also, I am not convinced the special casing would be significant).

For the second one I can imagine automating the collection of the mapping data, but the visual separation when reading the code base will be there for sure.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: I would like to keep this PR mostly focused on the structural changes, would you mind if I apply this one but put updating all of them into an issue on the board? I don't mind doing it in a follow-up PR in the cooldown but would prioritize the rest of the config project during the cycle.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: My understanding is that this is part of the wider config project (porting the ideas from #936 ) I am working on, so a followup is already planned.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: I do. However, won't it remain possible to override `__post_init__` without calling `super().__post_init__()`? In that case it would still help for config classes that do not require their own `__post_init__` logic though.

As the ADR outlines, I think it would be even better to not have to implement `_validate`.

model/common/src/icon4py/model/common/config/options.py
DropD: I lean more towards making `_opt_from_type_hint` "public" as well, since it covers all the cases anyway and is not unsafe to use. The names could be better and docstrings would help though.

bindings/src/icon4py/bindings/diffusion_wrapper.py
DropD: I added the above idea to the ADR and created #1334 to track future work on this.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: I created #1335 to track the follow-up

DropD:
cscs-ci run default

DropD:
cscs-ci run distributed

DropD:
cscs-ci run default

DropD:
cscs-ci run distributed

DropD:
cscs-ci run default

DropD:
cscs-ci run distributed

docs/development/ADRs/0002_configuration_classes.md
DropD: Added an example code snippet

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,219 @@
...
+```python
+class DiffusionConfig:
+    diffusion_type: Annotated[DiffusionType, ConfigOption(description="...", validation=...)]
+    ...
+```
+
+And in `icon4py.model.atmosphere.diffusion.fortran_icon_mappings`:
+
+```python
+CONFIC_OPTIONS: {DiffusionConfig: {"diffusion_type": ...}}
```
msimberg, line 202:
```suggestion
CONFIG_OPTIONS: {DiffusionConfig: {"diffusion_type": ...}}
```

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -145,206 +148,261 @@ class DiffusionConfig:
...
+
+    a_hshr: typing.Annotated[
+        float,
+        common_conf_opt.ConfigOption(
+            description="Scaling factor for horizontal shear production term.",
+            icon_equivalent=common_conf_opt.IconOption("a_hshr", ("turbdiff_nml",)),
+        ),
+    ] = 1.0
+
+    loutshs: typing.Annotated[
```
msimberg, line 385: I think this is a prime example of where we could do a bit better than icon fortran with naming... Just highlighting it, please don't change it in this PR.

I'm guessing stuff like this was inspiration for the realistic fortran names in the tests as well 😛

docs/development/ADRs/0002_configuration_classes.md
```diff
@@ -0,0 +1,219 @@
...
+```
+
+To arrive at the same usability, `icon4py.common.fortan_icon_mappings` could contain some infrastructure to find the relevant mapping for each configuration class (via module path heuristics or entry points or similar), so that a different version of `icon4py.common.config.options.construct_config_from_icon` could be provided.
+
```
msimberg, line 206: I like this section. I think the long term goal should still be to fully separate the python <-> fortran mapping and you might want to explicitly mention that here or together with

> The current solution will not achieve maximum separation between internals and their mapping to and from ICON concepts. However it provides a structured path towards it.

to make it clear to someone reading this that this is not the final shape of the python <-> fortran mapping.

msimberg (approved):
This looks good to me with some minor optional on clarifying the plan for fortran <-> python mappings.

I think this doesn't (significantly) block us from separating the mapping completely in the future, but this allows making progress on having our own config file infrastructure.

docs/development/ADRs/0002_configuration_classes.md
DropD: I was going to put this into a separate ADR, so it can be more broadly reviewed without blocking any code changes.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: Indeed. I will need help to better name this at a later point. The current description does not give me enough information about what it does for me to even suggest a better name.

docs/development/ADRs/0002_configuration_classes.md
DropD: I rephrased the section to make it apply more specifically to configuration classes and to state that the decision was only on whether to separate more now or later.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
jcanton: I was writing more to document rather than request anything different

jcanton (approved):
very nice!
now onwards to all other `Config`s ;-)

PR #1324, opened 2026-06-16
Compute analytical mean geometry quantities

msimberg:
Helper functionality for avoiding numeric instabilities in global reductions for e.g. mean cell area etc. For torus grids all triangles are exactly the same, so we don't really need a reduction. For icosahedral grids we can derive "global" quantities based on the sphere radius and root/level subdivision. For torus grids both methods yield almost the same result within floating point rounding errors (which is what I'm trying to avoid here). For icosahedral grids the difference can be a lot bigger, especially for limited area grids.

There's also now a GeometryConfig that can be used to select whether to use the analytical means (default) or to do the global reduction.


model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,57 @@ def __init__(
...
+    def _compute_analytical_means(self) -> dict[str, float] | None:
+        """Compute mean geometry values analytically from grid parameters.
+
+        For regular grids (global icosahedron and torus) the mean cell area,
+        edge length and their dual counterparts can be computed directly from
+        the grid parameters, avoiding non-deterministic global reductions.
+
+        Returns:
+            A dictionary of scalar mean values, or None if analytical
+            computation is not possible (e.g. limited-area grids).
```
msimberg: For limited area grids analytical computation is "possible" (with the global values). The question is rather if that's the correct value to use, or if it should really be based on the triangles in the local area.

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,57 @@ def __init__(
...
+
+        if geometry_type == icon.GeometryType.ICOSAHEDRON:
+            if self._grid.config.limited_area:
+                return None
+            subdivision = grid_params.subdivision
+            if subdivision is None:
+                return None
+            radius = grid_params.radius
+            if radius is None:
+                return None
```
msimberg: One question here is if this should be "automatic" (as currently) or through a config option. This will degrade gracefully but silently. We have to make sure there aren't cases where R/B accidentally aren't available, but we'd actually like to use the global analytical values.

Have to compare with ICON fortran.

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,57 @@ def __init__(
...
+            root = subdivision.root
+            level = subdivision.level
+            num_cells = 20 * root**2 * 4**level
+            num_vertices = num_cells // 2 + 2
```
msimberg: Check this.

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,57 @@ def __init__(
...
+            mean_dual_area = 4.0 * math.pi * radius**2 / num_vertices
+            mean_edge_length = math.sqrt(4.0 * mean_cell_area / math.sqrt(3.0))
+            mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
+        elif geometry_type == icon.GeometryType.TORUS:
+            domain_length = grid_params.domain_length
+            domain_height = grid_params.domain_height
+            if domain_length is None or domain_height is None:
+                return None
+            num_cells = self._grid.num_cells
+            mean_cell_area = domain_length * domain_height / num_cells
```
msimberg: Check this. Should it rather be based on the area of an equilateral triangle (is the result guaranteed to be identical?)

jcanton:
thoughts looking at fortran:

- `mean_edge_length` is not used anywhere so it doesn't matter

- `mean_cell_area` is used for:
  - divergence damping (where the exactness of the value *probably* does not matter)
  - `to_geometry_info%mean_characteristic_length = SQRT(to_geometry_info%mean_cell_area)` (in `mo_grid_geometry_info.f90`) which is then used in RBF to select the expression for the coefficients (and other parts that we didn't port yet).
     **possibly important side-note**: why do we compute it as `sqrt(mean_dual_area)` in `geometry.py`??
   
  as far as I can see it is always computed as (`mo_model_domimp_patches.f90`):
  ```
      ! divide the sphere surface by the number of cells
      ! Note: this works only for old grids
      patch%geometry_info%mean_cell_area = &
        & (4._wp * pi * patch%geometry_info%sphere_radius**2) &
        & / (REAL(20*nroot**2,wp)*4._wp**patch%level)
  ```
  or read from file (`mo_grid_geometry_info.f90`), which would explain the comment working only for old grids

model/common/src/icon4py/model/common/grid/geometry.py
jcanton: in the grid generator https://gitlab.dkrz.de/mpim-sw/grid-generator
it is assigned to triangle area: `torus_grid%geometry_info%mean_cell_area     = triangle_area` in `mo_create_torus_grid`
but your formula should also be correct because the torus is essentially a parallelogram with triangles re-shuffled on the side, and `length * height / num_cells` should be exactly correct

model/common/src/icon4py/model/common/grid/geometry.py
jcanton: `num_cells` is correct, see long comment below with code snippet and reference.
for `num_vertices` I had to ask a friend:

> The formulas match the code. Here's the evidence:
> estimate.py:135-141 — explicit computation:
> def compute_icon_standard_size(grid_refine_root, grid_refine_level):
>     cells = (20 * grid_refine_root * grid_refine_root) * (4**grid_refine_level)
>     verts = (10 * grid_refine_root * grid_refine_root) * 4**grid_refine_level + 2
>     edges = (30 * grid_refine_root * grid_refine_level) * 4**grid_refine_level
> Which is algebraically identical to:
> - num_cells = 20 * root² * 4^level ✓
> - num_vertices = num_cells // 2 + 2 ✓  (since 10 * root² * 4^level + 2 = (20 * root² * 4^level)/2 + 2)
> The same formula also appears in mo_model_domimp_patches.f90:973 as the denominator of the mean cell area calculation:
> & / (REAL(20*nroot**2,wp)*4._wp**patch%level)
> For a closed triangular mesh on a sphere, Euler's formula (V - E + C = 2) with E = 3C/2 gives V = C/2 + 2, so the vertex formula is a consequence of the cell formula.
> 

so if I understand correctly num_vertices has that value only for global grids, not limited area

model/common/src/icon4py/model/common/grid/geometry.py
jcanton: after online discussion we agree that we prefer analytical computations.
1. they are the fortran way and we like that
2. they produce the same values on a global and corresponding limited-area grids, which makes sense because the limited-area grid is a subset of the parent global, so it should be equivalent numerically.

model/common/src/icon4py/model/common/grid/geometry.py
jcanton: should we rather read from file (as fortran does) and only compute / error out if the file does not have it?

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,57 @@ def __init__(
...
+            mean_cell_area = 4.0 * math.pi * radius**2 / num_cells
+            mean_dual_area = 4.0 * math.pi * radius**2 / num_vertices
+            mean_edge_length = math.sqrt(4.0 * mean_cell_area / math.sqrt(3.0))
+            mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
+        elif geometry_type == icon.GeometryType.TORUS:
+            domain_length = grid_params.domain_length
+            domain_height = grid_params.domain_height
+            if domain_length is None or domain_height is None:
+                return None
+            num_cells = self._grid.num_cells
```
msimberg: This is actually likely the wrong num_cells (it's the local patch size, not th eglobal size).

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -314,45 +367,53 @@ def _register_computed_fields(self) -> None:
...
-            deps={
-                "buffer": attrs.DUAL_AREA,
-            },
-            fields=(attrs.MEAN_DUAL_AREA,),
-        )
-        self.register_provider(mean_dual_cell_area_np)
+        # TODO(msimberg): Should this be a config option, or is analytical
+        # computation always the right choice?
+        use_analytical_means = True
+        if use_analytical_means:
```
msimberg: This is the big TODO. Figure out if we can always use the analytical means, or if we ever have situations where we don't have this information.

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,58 @@ def __init__(
...
+                mean_dual_area = 4.0 * math.pi * radius**2 / num_vertices
+                mean_edge_length = math.sqrt(4.0 * mean_cell_area / math.sqrt(3.0))
+                mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
+            case icon.GeometryType.TORUS:
+                # For a uniform torus grid all cells are identical equilateral
+                # triangles. Read the common edge length directly from the edge
+                # length field (the grid file stores it on every edge).
+                # TODO(msimberg): Check if we can should get it from the grid
+                # file directly instead (e.g. via
+                # MPIMPropertyName.MEAN_EDGE_LENGTH).
```
msimberg: I think this would be nice to investigate, but is not a blocker in my opinion. We can easily get by with the current setup (of ignoring MEAN_EDGE_LENGTH etc. completely from the grid file).

msimberg:
cscs-ci run default;LEVELS=unit:integration

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +183,58 @@ def __init__(
...
+                mean_edge_length = float(self.get(attrs.EDGE_LENGTH).ndarray[0])
+                mean_cell_area = mean_edge_length**2 * math.sqrt(3.0) / 4.0
+                mean_dual_area = 2.0 * mean_cell_area
+                mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
```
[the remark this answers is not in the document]
msimberg: This is a very good point and would be an argument for reading quantities from the grid file (would be same on all ranks, guaranteed). I'll have to explore a bit if the edge lengths ever differ (they could if we compute them, they might be identical if computed in the grid generator).

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +190,58 @@ def __init__(
...
+                subdivision = grid_params.subdivision
+                root = subdivision.root
+                level = subdivision.level
+                num_cells = 20 * root**2 * 4**level
+                num_vertices = num_cells // 2 + 2
+                mean_cell_area = 4.0 * math.pi * radius**2 / num_cells
+                mean_dual_area = 4.0 * math.pi * radius**2 / num_vertices
+                mean_edge_length = math.sqrt(4.0 * mean_cell_area / math.sqrt(3.0))
+                mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
+            case icon.GeometryType.TORUS:
```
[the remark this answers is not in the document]
jcanton: but if this is how icon computes it, then that's what we'll do...

model/common/src/icon4py/model/common/grid/geometry.py
jcanton: true, but getting it from the grid file could have the positive side-effect that if anybody ever uses a different torus grid generator (of which there are none) this would error out and possibly point to a useful part of the code that should be looked at?

jcanton: (remove the TODO after deciding)

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +190,58 @@ def __init__(
...
+                # For a uniform torus grid all cells are identical equilateral
+                # triangles. Read the common edge length directly from the edge
+                # length field (the grid file stores it on every edge).
+                # TODO(msimberg): Check if we can should get it from the grid
+                # file directly instead (e.g. via
+                # MPIMPropertyName.MEAN_EDGE_LENGTH).
+                mean_edge_length = float(self.get(attrs.EDGE_LENGTH).ndarray[0])
+                mean_cell_area = mean_edge_length**2 * math.sqrt(3.0) / 4.0
```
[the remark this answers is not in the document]
jcanton: `bcast(ndarray[0])` or get from grid file

model/common/src/icon4py/model/common/grid/geometry.py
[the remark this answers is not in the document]
jcanton: see

https://gitlab.dkrz.de/mpim-sw/grid-generator/-/blob/master/src/grid_generator/mo_create_torus_grid.f90?ref_type=heads#L260 `SUBROUTINE fit_resolution()`

and lower down 

https://gitlab.dkrz.de/mpim-sw/grid-generator/-/blob/master/src/grid_generator/mo_create_torus_grid.f90?ref_type=heads#L791 `SUBROUTINE create_planar_torus_geometry()`

there is only one `edge_length` in the grid generator.

it's also used for
```
    half_edge_length =  edge_length * 0.5_wp
    dual_edge_length  = edge_length * (1.0_wp / SQRT(3.0_wp))
    triangle_area     = edge_length * edge_length * SQRT(0.1875_wp)
    hexagon_area      = dual_edge_length * edge_length * 1.5_wp
```
and
```
    torus_grid%geometry_info%mean_edge_length   = edge_length
    torus_grid%geometry_info%mean_dual_edge_length  = dual_edge_length
```

model/common/tests/common/grid/unit_tests/test_geometry.py
```diff
@@ -518,7 +557,41 @@ def test_geometry_mean_fields(
...
     assert value == pytest.approx(value_ref)
+
+
+@pytest.mark.datatest
+@pytest.mark.parametrize(
+    "attr_name", ["mean_edge_length", "mean_dual_edge_length", "mean_cell_area", "mean_dual_area"]
+)
+def test_analytical_and_global_reduction_mean_fields_agree(
+    backend: gtx_typing.Backend,
+    experiment: definitions.Experiment,
```
jcanton, line 538: not for now, but this should be parametrized by `grid_description` instead of `experiment`
only issue is that `get_grid_geometry` expects `Experiment` just to use `experiment.grid`...

jcanton: ah no, it uses `num_levels` as well...
please ignore

jcanton (commented):
just needs the TODOs handled as discussed, then looks good to go

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -314,45 +367,53 @@ def _register_computed_fields(self) -> None:
...
-                "buffer": attrs.DUAL_AREA,
-            },
-            fields=(attrs.MEAN_DUAL_AREA,),
-        )
-        self.register_provider(mean_dual_cell_area_np)
+        # TODO(msimberg): Should this be a config option, or is analytical
+        # computation always the right choice?
+        use_analytical_means = True
+        if use_analytical_means:
+            analytical_means = self._compute_analytical_means()
```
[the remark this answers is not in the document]
msimberg: I've added a GeometryConfig for this.

model/common/src/icon4py/model/common/grid/geometry.py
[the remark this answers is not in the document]
msimberg: I've added a broadcast from the first rank to all others so we're guaranteed to use the same mean edge length on all ranks. It may be slight overkill even with the knowledge that the grid generator computes the edge length exactly the same for all edges, but if we ever have another grid generator that computes the distances directly of edges without assuming that they're all the same, we have a bit of protection against that.

model/common/src/icon4py/model/common/grid/geometry.py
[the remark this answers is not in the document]
msimberg: Indeed. And now we have the option of choosing whether to do it like this or doing the global reduction (defaulting to "analytical" means, as in icon fortran).

model/common/tests/common/grid/unit_tests/test_geometry.py
msimberg: Grid manager needs num levels, but I think the grid geometry itself doesn't really need num_levels. I'd leave this as a cleanup for the fixture cleanup. `get_grid_geometry` shouldn't even exist and should just be a fixture as well.

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/grid/geometry.py
msimberg: I'd like to keep the todo becase 1. I still think it'd be nice to get it from the grid file but 2. I would like to avoid doing that in this PR. However, I'm not opposed to doing it now if you think it would be very useful to have already now (I'm wary to do it right now because of the ugliness of having to pass it on from the grid file via grid manager to the grid geometry).

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +199,69 @@ def __init__(
...
+                if self._process_props.comm is not None:
+                    array_ns = data_alloc.array_namespace(edge_length)
+                    send_buffer = array_ns.empty(1, dtype=edge_length.dtype)
+                    send_buffer[0] = edge_length[0]
+                    if hasattr(array_ns, "cuda"):
+                        array_ns.cuda.runtime.deviceSynchronize()
+                    self._process_props.comm.Bcast(send_buffer, root=0)
+                    mean_edge_length = float(send_buffer[0])
+                else:
+                    mean_edge_length = float(edge_length[0])
```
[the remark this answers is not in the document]
msimberg: Changed to a numpy buffer broadcast (still using Bcast, but avoiding the GPU -> CPU -> GPU transfers).

model/testing/src/icon4py/model/testing/grid_utils.py
```diff
@@ -96,9 +97,13 @@ def _download_grid_file(grid: definitions.GridDescription) -> pathlib.Path:
 
 
 def get_grid_geometry(
-    backend: gtx_typing.Backend | None, experiment: definitions.Experiment
+    backend: gtx_typing.Backend | None,
+    experiment: definitions.Experiment,
+    config: geometry_config.GeometryConfig,
 ) -> geometry.GridGeometry:
```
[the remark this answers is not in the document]
msimberg: I changed get_grid_geometry to take an ExperimentConfig instead and an explicit Grid, since those two are the only members accessed from Experiment. The GeometryConfig can be overridden via ExperimentConfig instead.

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -182,6 +200,66 @@ def __init__(
...
+                num_vertices = num_cells // 2 + 2
+                mean_cell_area = 4.0 * math.pi * radius**2 / num_cells
+                mean_dual_area = 4.0 * math.pi * radius**2 / num_vertices
+                mean_edge_length = math.sqrt(4.0 * mean_cell_area / math.sqrt(3.0))
+                mean_dual_edge_length = mean_edge_length / math.sqrt(3.0)
+            case icon.GeometryType.TORUS:
+                # For a uniform torus grid all cells are identical equilateral
+                # triangles. Read the common edge length directly from the edge
+                # length field (the grid file stores it on every edge).
+                # TODO(msimberg): Check if we can should get it from the grid
```
msimberg:
```suggestion
                # TODO(msimberg): Check if we can/should get it from the grid
```

msimberg:
cscs-ci run default

jcanton (approved):
me like

msimberg:
cscs-ci run merge

msimberg:
cscs-ci run merge

msimberg:
cscs-ci run merge

jcanton:
cscs-ci run merge

msimberg:
cscs-ci run merge

PR #1325, opened 2026-06-16
Assemble driver states and initial conditions cleanup

jcanton:


jcanton (commented):
from a friend — good cleanup, nice separation of concerns. A few observations:

### 1. Unnecessary reassignment pattern
In `standalone_driver.py:run_driver` and all test files, `initial_condition.create()` mutates `prognostic_state_now` in-place and returns the same reference. The reassignment like:
```
prognostic_state_now = initial_condition.create(..., prognostic_state_now=prognostic_state_now, ...)
```
is misleading — it suggests `create` might return a different object. Either remove the reassignment or have the IC functions NOT mutate in-place (i.e., make them pure functions that return a new state). Same pattern in the individual IC functions (`gauss3d`, `jablonowski_williamson`, `_fill_prognostics_from_serialbox`) where they both mutate the passed-in state AND return it.

**A:** reverted to mutating in place

### 2. Return values ignored in `config.py:create`
All three `match` arms in `config.py:create` call IC functions that return `prognostic_state_now`, but the return values are discarded. The function falls through to `return prognostic_state_now` at the bottom. This works (same reference) but is inconsistent with the return type annotation. If the IC functions are always mutating in-place, consider having them return `None` instead.

**A:** reverted to mutating in place

### 3. Diagnostic `temperature`/`pressure` fields no longer initialized
`jablonowski_williamson` used to fill `diagnostic_state.temperature`, `diagnostic_state.pressure`, and `diagnostic_state.pressure_ifc` during IC computation. These lines are removed — the diagnostic state now gets zero-initialized for these fields. The Newton iteration still runs (needed for `exner`/`rho`/`theta_v`), but the intermediate results aren't stored. Please verify that nothing reads `ds.diagnostic.temperature` or `ds.diagnostic.pressure` before the first timestep recomputes them.

**A:** intentional, they were dead code. they'll be handled properly by IO

### 4. Pre-existing: private attribute access
`standalone_driver.py:78` accesses `self.static_field_factories.metrics._vertical_grid`. This was already there before (just renamed from `metrics_field_source`), but worth noting as a follow-up cleanup.

**A:** leave it for follow-up

jcanton (commented):
from a friend — one more: in `standalone_driver.py:run_driver` and the two test files, the `prognostic_state_now` is initialized with `ntracer=icon4py_driver.config.driver.ntracer`, but `_fill_prognostics_from_serialbox` then iterates using `config.ntracer` from `FromFileConfig`. If these two `ntracer` values ever diverge, you'd get mismatched tracer array sizes or uninitialized tracers. The old code allocated inside the IC function with its own `ntracer`, so this was self-consistent. Worth a comment or assertion that they match.

**A:** getting fixed by https://github.com/C2SM/icon4py/pull/1317

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/gauss3d.py
```diff
@@ -100,31 +103,27 @@ def gauss3d(
...
+        exner_ndarray[:, num_levels - 1] = 1.0 - geopot[:, num_levels - 1] / phy_const.CPD / t0
 
     testcases_utils.hydrostatic_adjustment_constant_thetav_ndarray(
-        wgtfac_c=metrics["wgtfac_c"],
-        ddqz_z_half=metrics["ddqz_z_half"],
-        exner_ref_mc=metrics["exner_ref_mc"],
-        d_exner_dz_ref_ic=metrics["d_exner_dz_ref_ic"],
-        theta_ref_mc=metrics["theta_ref_mc"],
-        theta_ref_ic=metrics["theta_ref_ic"],
+        wgtfac_c=wgtfac_c,
```
msimberg, line 121: Just checking: was `metrics["wgtfac_c"]` previously passed in incorrectly as a `Field`? It's now explicitly passed in as an ndarray, and I'm wondering if it was wrong before or is wrong now?

msimberg (commented):
Looks straightforward enough. I'm very slightly disliking the change to modifying fields in place, but this is something I think we can reconsider later (is e.g. an initial condition a Component with inputs and outputs?

Minor question on types, and one test is failing which still needs some attention.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/gauss3d.py
jcanton: they were all ndarrays before as well, just coming from this now-defunct helper https://github.com/C2SM/icon4py/pull/1325/changes/BASE..80d092d689541eefb902e6f8988e4ced7990ccd5#diff-8e7d93c45dfd7ef32c32afcffe4e0424f9da44e67d101a85e80ac41f1dc4f878L260

jcanton:
cscs-ci run default

jcanton:
cscs-ci run distributed

msimberg (approved):

PR #1294, opened 2026-05-28
Unified CI pipelines

msimberg:
**Requested review from a lot of people. Not expecting everyone to look, but feel free to start the bikeshedding...**

This PR unifies the default, extra, distributed, dace pipelines into one pipeline that can be parameterized to run different subsets of tests.

The test jobs are generated by a new script `generate_ci_pipeline.py` because we can't use variables to dynamically filter gitlab matrix jobs. The test jobs are triggered as a gitlab child pipeline (https://docs.gitlab.com/ci/pipelines/downstream_pipelines/#parent-child-pipelines).

There are now three main CI pipelines (in addition to the benchmark pipelines which are mostly unchanged, except to adapt to the changes in `base.yml`):
- `default` still exists, and should be the "default" pipeline to run when you want to check if the PR is roughly in shape. This runs serial and MPI tests with one GPU backend. Bikeshedding on what belongs here welcome, but in my opinion this should be relatively small. `default` can and should be customized to run a subset of tests that are relevant for your changes (see below). A future extension would be to automatically detect which tests actually need to run based on the changes in the PR (see e.g. #1291). This will not be implemented in this PR.
- `merge` is new, and will be made gating for PR merges. `default` will not be required for merges since it's a subset of `merge`. `merge` is roughly like the current `default` pipeline in that it runs most tests with all backends. This can be triggered when a PR has been approved and auto-merge has been enabled. A future extension is to make this pipeline run using merge queues (https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/managing-a-merge-queue).
- `all` is also new, and runs _all_ tests, including future "extended" or "validation" tests (#1303). This is meant to be run e.g. once a week on `main`. I've currently set up a run that runs daily at 2 AM on this branch (first run here: https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504677/-/pipelines/2624464232). I'll update this to `main` and weekly when this is merged.

As mentioned above, the `default` pipeline is intended to be customized with CI variables, e.g.

> cscs-ci run default;SESSIONS=model;MODEL_SUBPACKAGES=dycore

All variables can be overwritten. Otherwise they take the default values from the `default` pipeline definition. The test reminder has been updated with some hints and reminders about which pipelines to run and how to customize them.

The `dace`, `extra`, and `distributed` pipelines have been removed since they're covered by the other pipelines now.

[...]

PR #1337, opened 2026-06-23
Make unmarked tests implicitly unit tests

msimberg:
Currently selecting a test level marker with `pytest --level=unit` will run all tests marked `unit` and unmarked tests. Likewise `--level=integration` will run all tests marked `integration` and unmarked tests. This means that `pytest --level=unit` and `pytest --level=integration` aren't disjoint subsets as one would expect.

This makes all unmarked tests implicitly unit tests as I think that's a reasonable default level for tests.

We can separately consider removing the level marker and rely fully on `foo_tests` directories and `-k foo_tests`. With this PR I just want to correct what to me seems like a bug, to help with #1294 where I don't want the same tests to run in multiple CI jobs.

### DIFF
--- a/model/testing/src/icon4py/model/testing/pytest_hooks.py
+++ b/model/testing/src/icon4py/model/testing/pytest_hooks.py
@@ -34,7 +34,7 @@ def pytest_configure(config):
     )
     config.addinivalue_line(
         "markers",
-        "level(name): marks test as unit or integration tests. tests without this marker are implicitly treated as 'unit'.",
+        "level(name): marks test as unit or integration tests, tests without this marker are implicitly treated as 'unit'",
     )
 
     # Check if the --enable-mixed-precision option is set and set the environment variable accordingly
--- a/model/testing/src/icon4py/model/testing/pytest_hooks.py
+++ b/model/testing/src/icon4py/model/testing/pytest_hooks.py
@@ -134,8 +134,7 @@ def pytest_collection_modifyitems(config, items):
     if test_level == "any":
         return
     for item in items:
-        marker = item.get_closest_marker("level")
-        if marker is not None:
+        if (marker := item.get_closest_marker("level")) is not None:
             assert all(level in _TEST_LEVELS for level in marker.args), (
                 f"Invalid test level argument on function '{item.name}' - possible values are {_TEST_LEVELS}"
             )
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run default

msimberg:
This will reduce the number of tests run in CI, though I might increase it again in #1294 (or at least change when we run what tests, all tests don't need to run for quick tests at least).

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -134,7 +134,8 @@ def pytest_collection_modifyitems(config, items):
     if test_level == "any":
         return
```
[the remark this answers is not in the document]
msimberg: Keeping this since it's the same behaviour as before.

msimberg:
#1294 will run both `--level=unit` and `--level=integration` as disjoint sets of tests before merge.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -111,31 +127,15 @@ def pytest_addoption(parser: pytest.Parser):
...
-        if not items:
-            pytest.exit("No tests assigned to this MPI subcomm group", returncode=0)
-
     test_level = config.getoption("--level")
     if test_level == "any":
         return
     for item in items:
-        if (marker := item.get_closest_marker("level")) is not None:
+        marker = item.get_closest_marker("level")
+        if marker is not None:
```
edopao: I liked the old syntax, but it's just a matter of taste.

edopao (approved):
This change seems good to me.

model/testing/src/icon4py/model/testing/pytest_hooks.py
msimberg: Me too, thanks for noticing (you can guess that an LLM made this change, and I didn't notice 😞). I'll change it back.

msimberg:
```suggestion
        if (marker := item.get_closest_marker("level")) is not None:
```

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1339, opened 2026-06-24
Add ruff rule name to noqa in `io.py`

msimberg:
See discussion in https://github.com/C2SM/icon4py/commit/878bb18f695520f7c85ea7a16b66f34bd203440b#r189834412.

Purely cosmetic, just to give the name of the RUF009 rule in the comment, as suggested in https://github.com/C2SM/icon4py/blob/744c5463d62ed3b35a5ecf36ce902b6857bf27de/CODING_GUIDELINES.md?plain=1#L121.


kotsaloscv (approved):
Thanks @msimberg !

PR #1303, opened 2026-06-10
Standalone driver validation test

msimberg:
Adds a new testing level (`validation`) and one test that uses it. The test is the parallel standalone driver test, with a seven-day simulation time. With CPU backends, and the newly added testing option ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE we expect bitwise identical results between single and multi-rank runs. The option is so far only used for the one parallel standalone driver test, but could in theory be used in other comparison tests as well. I'm leaving that as a future possible extension. To actually get bitwise identical results, CXXFLAGS=-ffp-contract=off has to be set.

PR #1343, opened 2026-06-25
Set nonzero tolerance for `RHO_REF_ME` with `dace_gpu` in `test_parallel_grid_manager.py`

msimberg:
I don't think this is a regression. This field is just not tested by default in the usual default pipeline (one of many "extra" fields that are skipped because they take a long time to test for little added value).

### DIFF
--- a/model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
+++ b/model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
@@ -542,7 +542,11 @@ def _compare_metrics_fields_single_multi_rank(
             model_backends.is_cpu_backend(backend)
             or (
                 model_backends.is_gpu_backend(backend)
-                and attrs_name == metrics_attributes.DDQZ_Z_FULL_E
+                and attrs_name
+                in {
+                    metrics_attributes.DDQZ_Z_FULL_E,
+                    metrics_attributes.RHO_REF_ME,
+                }
             )
         ):
             # TODO (jcanton,phimuell): figure out dace undeterministic behaviour


model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
```diff
@@ -542,7 +542,11 @@ def _compare_metrics_fields_single_multi_rank(
             model_backends.is_cpu_backend(backend)
             or (
                 model_backends.is_gpu_backend(backend)
-                and attrs_name == metrics_attributes.DDQZ_Z_FULL_E
+                and attrs_name
+                in {
+                    metrics_attributes.DDQZ_Z_FULL_E,
+                    metrics_attributes.RHO_REF_ME,
+                }
```
edopao, line 549: I am a bit suspicious about the CPU backend being much more permissive, since it does not filter the test cases, although it is unrelated to this PR.

edopao (approved):

model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
msimberg: Yeah, you're right to be suspcicious, but I think it's just a consequence of the special cases being added in separate PRs. For GPU I started adding whitelists for individual fields while CPU was at some point just set to non-zero for all fields without checking which ones need it.

I'm hoping https://github.com/C2SM/icon4py/pull/1303 will improve this a bit. Then dace_cpu should always be expected to have zero diff.

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

jcanton (approved):

msimberg:
Restarted https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504675/-/jobs/15031642841 manually. CI-related failure, not because of icon4py.

PR #1344, opened 2026-06-25
Fix SyntaxWarnings from invalid escape sequences in strings

msimberg:
A lot of strings (docstrings included) use `\` not to actually escape anything, but just as a plain backslash. This PR fixes warnings by invalid escape sequences by either using raw strings where appropriate. This also adds W605 (https://docs.astral.sh/ruff/rules/invalid-escape-sequence/) to the ruff checks.

Without this, running icon4py results in warnings such as
```
model/common/src/icon4py/model/common/grid/grid_refinement.py:53: SyntaxWarning: invalid escape sequence '\ '
  /\ 1  /\ 1  /\ 1  /    ← Row 1
```
at runtime which is pure noise.

### DIFF
--- a/model/atmosphere/dycore/docs/ext/icon4py_sphinx.py
+++ b/model/atmosphere/dycore/docs/ext/icon4py_sphinx.py
@@ -254,7 +254,7 @@ class ScidocMethodDocumenter(autodoc.MethodDocumenter):
                 continue
 
             # Collect offset providers
-            if "\offProv" in line:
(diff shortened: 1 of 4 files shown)


jcanton (approved):
thanks, I was annoyed by the runtime warnings :-)

msimberg:
cscs-ci run default

msimberg:
cscs-ci run distributed

PR #1345, opened 2026-06-25
NonHydrostaticConfig -> dataclass

DropD:
Turn NonHydrostaticConfig into a dataclass in declarative style (as described in [ADR 0002](https://github.com/C2SM/icon4py/blob/main/docs/development/ADRs/0002_configuration_classes.md)).


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/dycore_states.py
```diff
@@ -90,7 +90,7 @@ class DiagnosticStateNonHydro:
     """Data class containing diagnostic fields that are calculated in the dynamical core (SolveNonHydro)."""
 
     # `max_vertical_cfl` stored as 0-d array of type ta.wpfloat (to be able to avoid cupy synchronization)
-    max_vertical_cfl: data_alloc.ScalarLikeArray[ta.wpfloat]
+    max_vertical_cfl: data_alloc.ScalarLikeArray[ta.wpfloat]  # type: ignore[type-var] # TODO(ricoh): find out what this is about
```
jcanton, line 93: you mean the variable itself or the reason behind the type ignore?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -82,7 +84,7 @@ class IntermediateFields:
     contain state that is built up over the predictor and corrector part in a timestep.
     """
 
-    horizontal_pressure_gradient: fa.EdgeKField[ta.vpfloat]
+    horizontal_pressure_gradient: fa.EdgeKField[float]
```
jcanton: ignorance / question for myself: why these changed type?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -138,6 +142,7 @@ def allocate(cls, grid: grid_def.Grid, allocator: gtx_typing.Allocator | None):
         )
 
 
+@dataclasses.dataclass
```
jcanton: `kw_only`?

jcanton (approved):
LGTM
just two minor comments / questions more for myself than anything

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -347,11 +347,7 @@ class DiffusionConfig:
     max_nudging_coefficient: typing.Annotated[
         float,
         common_conf_opt.ConfigOption(
-            description=(
-                "Parameter describing the lateral boundary nudging in limited area mode."
-                "Maximal value of the nudging coefficients used cell row bordering the boundary interpolation zone,"
-                "from there nudging coefficients decay exponentially with `nudge_efold_width` in units of cell rows."
-            ),
+            description="Maximum relaxation coefficient for lateral boundary nudging",
```
msimberg, line 350: I like the shortening, but maybe a scientist can comment whether the extra verbosity in the old version was useful or not?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/vertically_implicit_dycore_solver.py
```diff
@@ -439,7 +439,7 @@ def vertically_implicit_solver_at_predictor_step(
     end_cell_index_halo_lvl1: gtx.int32,
     vertical_start_index_model_top: gtx.int32,
     vertical_end_index_model_surface: gtx.int32,
-):
+) -> None:
```
msimberg, line 442: Just for me to understand: all these new type hints and type ignores come from you enabling mypy for these files and are not strictly required for the config changes?

Keep the changes (and thank you for fixing them), I'm just asking to understand that there isn't another reason (actually related to the config) for changing these now.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -495,6 +595,8 @@ def __init__(
             offset_provider=self._grid.connectivities,
         )
 
+        assert self._cell_params.mean_cell_area is not None
```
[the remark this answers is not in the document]
msimberg: I think you can ignore copilot for these asserts.

Better long term would be to investigate if the members of `CellParams` etc. can be made non-nullable, but given that they are nullable I find the asserts ok.

msimberg (approved):
No blocking comments, thanks!

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
jcanton: we thought we'd use the short / first line commetn in `icon-nwp/doc/www/documentation/buildrun/buildrun_namelist_overview.md`
then it's consistent everywhere and if more verbose info is needed people can go look up the docs and we avoid duplicating here info that may become outdated

jcanton: should be found here too:
https://docs.icon-model.org/documentation/buildrun/buildrun_namelist_overview.html

do you think we should mention this and / or add the link for reference?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/dycore_states.py
DropD: Just the type ignore

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
DropD: Thanks, meant to add that!

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: I would like that. I had a conversation about this with @jcanton and we agreed that for now we would try to slowly converge toward the first sentence of the ICON namelist docs (where there is an ICON equivalent). The rest of the info is still always available there.

From there it will be easier to have discussions about where something needs to be added.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/vertically_implicit_dycore_solver.py
DropD: Indeed, these are orthogonal to the config changes (they are also all in a separate commit).

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
DropD: `vpfloat` is not (genuinely) a valid type for type hints. What it is trying to say is "this can be one of a number of floating point types".

Due to the GT4Py mypy plugin, all float types are treated as equivalent during static typing, so `float` is equivalent for now. However, possibly `ta.anyfloat` is an even better equivalent way to state the same.

DropD:
cscs-ci run merge

PR #1347, opened 2026-06-25
Add new APE experiment with AES physics

jcanton:
add aquaplanet experiment with AES physics.
sync with https://gitlab.dkrz.de/icon/icon-nwp/-/merge_requests/2200


yiluchen1066 (approved):

jcanton:
cscs-ci run merge

jcanton (commented):
from a friend:
Thanks for this PR! The moist IC initialization and the pressure-diagnosis refactoring look well-structured. The physics is faithful to the ICON Fortran (Tetens constants, qsat_rho formula, RH profile, moisture iteration, global rescaling). I have a few minor comments below.

One broader question: the `baroclinic_amplitude` default changed from `0.0` to `1.0`. The existing `JW` experiment uses `jabw_s` which is explicitly reset to `0.0`, so it's safe. But are there any other callers that rely on the old default without going through `from_fortran_dict`?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -77,14 +100,21 @@ def from_fortran_dict(
         run_nml = atm_dict["run_nml"]
         master_time_control_nml = master_dict["master_time_control_nml"]
         master_model_nml = master_dict["master_model_nml"]
-        dtime = run_nml["dtime"]
+        # Both 'modeltimestep' (an ISO 8601 duration) and 'dtime' (seconds) are
+        # always present; a non-empty 'modeltimestep' takes priority over 'dtime'.
```
jcanton, line 104: from a friend: The comment says "always present" but the code uses `run_nml.get("modeltimestep", "")` with a fallback, and `test_missing_modeltimestep_falls_back_to_dtime` explicitly tests the missing case. Consider updating to something like "'modeltimestep' may be absent; a non-empty value takes priority over 'dtime'."

model/common/src/icon4py/model/common/thermodynamic_functions.py
```diff
@@ -0,0 +1,75 @@
...
+    array_ns = data_alloc.array_namespace(rho)
+    saturation_pressure = array_ns.where(
+        temperature <= phy_const.MELTING_TEMPERATURE,
+        sat_pres_ice(array_ns.maximum(temperature, 180.0)),
+        sat_pres_water(temperature),
+    )
+    # avoid water vapour pressure > total pressure
+    vapour_pressure = array_ns.minimum(saturation_pressure, pressure / (relative_humidity + 1.0e-6))
+    saturation_qv = vapour_pressure / (rho * phy_const.RV * temperature)
+    return array_ns.minimum(saturation_qv, relative_humidity * saturation_qv)
```
jcanton: from a friend: The `min(saturation_qv, relative_humidity * saturation_qv)` is slightly confusing. When `rh < 1` (the normal case), this always returns `rh * saturation_qv`. The `min` only matters when `rh > 1` (a safety cap at saturation). Consider adding a brief comment explaining the `min` is a supersaturation guard, or restructuring as `rh * saturation_qv` with an explicit `min(..., saturation_qv)` cap.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -253,3 +275,46 @@ def jablonowski_williamson(  # noqa: PLR0915 [too-many-statements]
...
+            grid, dims.CellDim, dims.KDim, allocator=allocator, dtype=ta.wpfloat
+        )
+        virtual_temperature.ndarray[:, :] = theta_v_ndarray * exner_ndarray
+        pressure_ndarray = diagnose_pressure_ndarray(
+            grid=grid,
+            backend=backend,
+            allocator=allocator,
+            exner=prognostic_state_now.exner,
+            virtual_temperature=virtual_temperature,
+            ddqz_z_full=metrics.get(metrics_attributes.DDQZ_Z_FULL),
```
jcanton: from a friend: `metrics.get(metrics_attributes.DDQZ_Z_FULL)` is fetched again here as a field, but the ndarray was already extracted at line 128 (`ddqz_z_full = metrics.get(...).ndarray`). Consider storing the field in a variable at line 128 and reusing it here to avoid the redundant lookup, e.g. `ddqz_z_full_field = metrics.get(...)` then `.ndarray` for the local variable.

model/common/src/icon4py/model/common/diagnostic_calculations/pressure.py
```diff
@@ -0,0 +1,134 @@
...
+    # surface pressure lives at the bottom interface; extract it as a cell field
+    surface_pressure.ndarray[:] = surface_pressure_k.ndarray[:, num_levels]
+    pressure_ifc.ndarray[:, -1] = surface_pressure.ndarray
+
+    diagnose_pressure_stencil.diagnose_pressure.with_backend(backend)(
+        ddqz_z_full,
+        virtual_temperature,
+        surface_pressure,
+        pressure,
+        pressure_ifc,
```
jcanton: from a friend: The `diagnose_pressure` call uses positional arguments while the `diagnose_surface_pressure` call above (line 68) uses keyword arguments. For consistency, consider using kwargs for both stencil calls.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
```diff
@@ -221,6 +226,66 @@ def init_w(
...
+        # hydrometeors are zero here, so only qv enters (see diagnose_temperature).
+        temperature = virtual_temperature / (1.0 + phy_const.RV_O_RD_MINUS_1 * qv_values)
+        qv_values = _qv(temperature)
+
+    if normalize_global_moisture:
+        # rescale qv so the global mean column-integrated moisture matches the
+        # prescribed value (Fortran opt_global_moist / ztmc_ape).
+        column_moisture = global_reductions.sum(
+            ddqz_z_full * rho * qv_values * cell_area[:, array_ns.newaxis]
+        )
```
jcanton, line 290: from a friend: The `rho` in the column moisture integration is the dry-air density from the hydrostatic adjustment (no moisture feedback). This is consistent with ICON where `opt_global_moist` is a multiplicative rescaling, but it might be worth a one-line comment noting that `rho` here is the pre-moisture density.

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -49,6 +50,28 @@ class ProfilingStats:
...
+#: ISO 8601 duration, restricted to the fixed-length components (weeks, days,
+#: hours, minutes, seconds). Years and months are intentionally not matched since
+#: their length is not fixed, and this is currently only used for dtime.
+_ISO8601_DURATION = re.compile(
+    r"P(?:(?P<weeks>\d+)W)?(?:(?P<days>\d+)D)?"
+    r"(?:T(?:(?P<hours>\d+)H)?(?:(?P<minutes>\d+)M)?(?:(?P<seconds>\d+(?:\.\d+)?)S)?)?"
+)
+
+
+def _timedelta_from_iso8601(duration: str) -> datetime.timedelta:
```
jcanton, line 62: from a friend: The regex accepts strings like `"P1DT"` (trailing T with no time components) which is technically invalid ISO 8601 — it would silently parse as `timedelta(days=1)`. Low priority since dtime values are typically `"PT300S"` format, but you could require at least one time component after the T by restructuring the time group.

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

model/common/src/icon4py/model/common/diagnostic_calculations/pressure.py
```diff
@@ -0,0 +1,134 @@
...
+        ddqz_z_full=ddqz_z_full,
+        surface_pressure=surface_pressure_k,
+        horizontal_start=0,
+        horizontal_end=horizontal_end,
+        vertical_start=num_levels,
+        vertical_end=num_levels + 1,
+        offset_provider={},
+    )
+    # surface pressure lives at the bottom interface; extract it as a cell field
+    surface_pressure.ndarray[:] = surface_pressure_k.ndarray[:, num_levels]
```
yiluchen1066: ```surface_pressure_k``` seems redundant. I think we can pass ```pressure_ifc``` directly and remove ```surface_pressure_k``` entirely: 

```suggestion
    diagnose_surface_pressure_stencil.diagnose_surface_pressure.with_backend(backend)(
        exner=exner,
        virtual_temperature=virtual_temperature,
        ddqz_z_full=ddqz_z_full,
        surface_pressure=pressure_ifc,
        horizontal_start=0,
        horizontal_end=horizontal_end,
        vertical_start=num_levels,
        vertical_end=num_levels + 1,
        offset_provider={},
    )
    # surface pressure lives at the bottom interface; extract it as a cell field
    surface_pressure.ndarray[:] = pressure_ifc.ndarray[:, num_levels]
```

model/common/src/icon4py/model/common/diagnostic_calculations/pressure.py
```diff
@@ -0,0 +1,134 @@
...
+    pressure_ifc.ndarray[:, -1] = surface_pressure.ndarray
+
+    diagnose_pressure_stencil.diagnose_pressure.with_backend(backend)(
+        ddqz_z_full=ddqz_z_full,
+        virtual_temperature=virtual_temperature,
+        surface_pressure=surface_pressure,
+        pressure=pressure,
+        pressure_ifc=pressure_ifc,
```
yiluchen1066: could we rename ```pressure_ifc``` to ```pressure_on_cells_half_levels```, It matches with the naming we just standardized on in ```muphys/state.py```

model/common/src/icon4py/model/common/thermodynamic_functions.py
```diff
@@ -0,0 +1,76 @@
...
+    saturation_pressure = array_ns.where(
+        temperature <= phy_const.MELTING_TEMPERATURE,
+        sat_pres_ice(array_ns.maximum(temperature, 180.0)),
+        sat_pres_water(temperature),
+    )
+    # avoid water vapour pressure > total pressure
+    vapour_pressure = array_ns.minimum(saturation_pressure, pressure / (relative_humidity + 1.0e-6))
+    saturation_qv = vapour_pressure / (rho * phy_const.RV * temperature)
+    # avoid supersaturation: if rh > 1 return saturation_qv
+    return array_ns.minimum(saturation_qv, relative_humidity * saturation_qv)
```
yiluchen1066:
```suggestion
    return array_ns.minimum(relative_humidity, 1.0) * staturation_qv
```

model/common/src/icon4py/model/common/thermodynamic_functions.py
```diff
@@ -0,0 +1,76 @@
...
+    capped so the vapour pressure cannot exceed the total pressure, converted to a
+    saturation specific humidity, and scaled by the relative humidity.
+
+    This is the general computation only; the relative-humidity *profile* and any
+    test-case specific caps are the caller's responsibility.
+    """
+    array_ns = data_alloc.array_namespace(rho)
+    saturation_pressure = array_ns.where(
+        temperature <= phy_const.MELTING_TEMPERATURE,
+        sat_pres_ice(array_ns.maximum(temperature, 180.0)),
```
yiluchen1066: make ```180.0``` into a named constant

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -56,12 +60,28 @@ class JablonowskiWilliamsonConfig:
     dtemp: float = 4.8e5
     lon_perturbation_center: float = math.pi / 9.0
     lat_perturbation_center: float = 2.0 * math.pi / 9.0
+    # Moist tracer initialization (inwp tracers, see init_nh_inwp_tracers in
+    # mo_nh_jabw_exp.f90). Relevant only when transport is active.
+    rh_at_1000hpa: float = 0.7
+    qv_max: float = 20e-3
+    #: number of iterations to converge qv against the moisture-dependent
+    #: temperature (Fortran l_rediag=.TRUE. => 10 iterations).
```
yiluchen1066:
```suggestion
    # number of iterations to converge qv against the moisture-dependent
    # temperature (Fortran l_rediag=.TRUE. => 10 iterations).
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -56,12 +60,28 @@ class JablonowskiWilliamsonConfig:
...
+    moisture_init_iterations: int = 10
+    #: target column-integrated moisture [kg/m**2] (ztmc_ape). Only used by the
+    #: APE test cases, which rescale qv to match this global value.
+    global_moisture_content: float = 25.006
+    #: whether to rescale qv to ``global_moisture_content``; enabled for the APE
+    #: cases (mirrors the optional ``opt_global_moist`` argument in Fortran).
```
yiluchen1066: I agree, our friend generated very nice documentation here. Maybe we can make it a bit shorter to stay consistent with the rest of the code?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -56,12 +60,28 @@ class JablonowskiWilliamsonConfig:
...
+    # Moist tracer initialization (inwp tracers, see init_nh_inwp_tracers in
+    # mo_nh_jabw_exp.f90). Relevant only when transport is active.
+    rh_at_1000hpa: float = 0.7
+    qv_max: float = 20e-3
+    #: number of iterations to converge qv against the moisture-dependent
+    #: temperature (Fortran l_rediag=.TRUE. => 10 iterations).
+    moisture_init_iterations: int = 10
+    #: target column-integrated moisture [kg/m**2] (ztmc_ape). Only used by the
+    #: APE test cases, which rescale qv to match this global value.
+    global_moisture_content: float = 25.006
```
yiluchen1066, line 71: if this one is only used by the APE test case, shall we add a suffix ```ape``` with the name?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -104,7 +127,9 @@ def jablonowski_williamson(  # noqa: PLR0915 [too-many-statements]
     theta_ref_ic = metrics.get(metrics_attributes.THETA_REF_IC).ndarray
     wgtfac_c = metrics.get(metrics_attributes.WGTFAC_C).ndarray
     ddqz_z_half = metrics.get(metrics_attributes.DDQZ_Z_HALF).ndarray
-    c_lin_e = static_fields.interpolation.get(interpolation_attributes.C_LIN_E)
+    ddqz_z_full_field = metrics.get(metrics_attributes.DDQZ_Z_FULL)
+    ddqz_z_full = ddqz_z_full_field.ndarray
```
yiluchen1066: we can inline this here

model/common/src/icon4py/model/common/constants.py
```diff
@@ -72,6 +72,16 @@
 #: Triple point of water at 611hPa [K]
 WATER_TRIPLE_POINT_TEMPERATURE: Final[ta.wpfloat] = 273.16
 
+#: Tetens formula constants for the saturation vapour pressure, called c1es, c3les,
+#: c4les, c3ies and c4ies in ICON (mo_lookup_tables_constants.f90).
+#: e_sat = TETENS_P0 * exp(A * (T - tmelt) / (T - B)), with the *_WATER coefficients
+#: over liquid water and the *_ICE coefficients over ice.
```
yiluchen1066:
```suggestion
# Tetens formula constants for the saturation vapour pressure, called c1es, c3les,
# c4les, c3ies and c4ies in ICON (mo_lookup_tables_constants.f90).
# e_sat = TETENS_P0 * exp(A * (T - tmelt) / (T - B)), with the *_WATER coefficients
# over liquid water and the *_ICE coefficients over ice.
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -253,3 +278,46 @@ def jablonowski_williamson(  # noqa: PLR0915 [too-many-statements]
...
+        log.info("Initializing inwp tracers (qv from RH profile, others zero).")
+
+        # virtual temperature theta_v * exner is the post-hydro_adjust base (not the
+        # pre-adjust temperature_jw); it is independent of qv and feeds both the
+        # hydrostatic pressure diagnosis and the moist-iteration first guess, so the
+        # iteration converges to the same fixed point as Fortran.
+        virtual_temperature = data_alloc.zero_field(
+            grid, dims.CellDim, dims.KDim, allocator=allocator, dtype=ta.wpfloat
+        )
+        virtual_temperature.ndarray[:, :] = theta_v_ndarray * exner_ndarray
```
yiluchen1066:
```suggestion
        virtual_temperature = gtx.as_field(
    (dims.CellDim, dims.KDim), theta_v_ndarray * exner_ndarray, allocator=allocator
)
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
yiluchen1066: sorry, I think to remove the duplication: 
we can remove ```ddqz_z_full = ddqz_z_full_field.ndarray``` here

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -253,3 +278,46 @@ def jablonowski_williamson(  # noqa: PLR0915 [too-many-statements]
...
+            exner=prognostic_state_now.exner,
+            virtual_temperature=virtual_temperature,
+            ddqz_z_full=ddqz_z_full_field,
+        )
+        testcases_utils.init_inwp_tracers(
+            rho=rho_ndarray,
+            virtual_temperature=virtual_temperature.ndarray,
+            pressure=pressure_ndarray,
+            cell_area=cell_area,
+            ddqz_z_full=ddqz_z_full,
```
yiluchen1066:
```suggestion
            ddqz_z_full=ddqz_z_full_field.ndarray
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -253,3 +278,46 @@ def jablonowski_williamson(  # noqa: PLR0915 [too-many-statements]
         num_levels=num_levels,
     )
     log.info("Hydrostatic adjustment computation completed.")
+
+    # Moist initialization only runs when transport is active. The only tracer we
+    # need to set is qv; the hydrometeors (qc, qi, ...) keep their zero-initialized
+    # value, so we don't touch them.
+    active_tracers = {name for name, _ in prognostic_state_now.tracer.active_fields()}
```
yiluchen1066, line 282: from a friend: 

Question rather than a change: the gate here is "any active tracer ⟹ run moist init and require qv." That's correct if every tracer that can be active in this configuration is a moisture species. But if a non-moisture tracer could ever be active without qv (now or later), this would wrongly raise "requires the 'qv' tracer to be active." Is "active tracers" guaranteed to imply "moisture transport" here, or should the gate key specifically on qv/the moisture species rather than "any tracer"?

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
```diff
@@ -221,6 +226,68 @@ def init_w(
...
+    array_ns = data_alloc.array_namespace(rho)
+
+    # linearly decreasing relative humidity with height (independent of qv)
+    relative_humidity = array_ns.maximum(rh_at_1000hpa - 0.5 + pressure / 200000.0, 0.0)
+
+    def _qv(temperature: data_alloc.NDArray) -> data_alloc.NDArray:
+        q = thermo.qv_from_relative_humidity(temperature, pressure, rho, relative_humidity)
+        # stratosphere and tropics caps (init_nh_inwp_tracers)
+        q = array_ns.where(pressure <= 10000.0, array_ns.minimum(q, 5.0e-6), q)
+        return array_ns.minimum(q, qv_max)
```
yiluchen1066: shall we also define these numbers as named constants in ```phy_const``` and import them? like

```
    _STRATOSPHERE_PRESSURE = 10000.0     
    _STRATOSPHERE_QV_CAP = 5.0e-6
    _RH_REFERENCE_PRESSURE = 200000.0    
```

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
```diff
@@ -221,6 +226,68 @@ def init_w(
...
+    if normalize_global_moisture:
+        # rescale qv so the global mean column-integrated moisture matches the
+        # prescribed value (Fortran opt_global_moist / ztmc_ape).
+        column_moisture = global_reductions.sum(
+            ddqz_z_full * rho * qv_values * cell_area[:, array_ns.newaxis]
+        )
+        total_area = global_reductions.sum(cell_area)
+        mean_column_moisture = column_moisture / total_area
+        if mean_column_moisture > 1.0e-25:
+            qv_values = qv_values * (global_moisture_content / mean_column_moisture)
```
yiluchen1066, line 294: from a friend: 

This uniform rescale runs after the _qv caps (qv_max, the 5e-6 stratosphere cap) are applied in the iteration, so a cell can end up above those caps post-rescale. I assume that's intentional and matches Fortran's end-of-routine ztmc_ape rescale — could you confirm ICON doesn't re-apply the caps after the global normalization? Just want to be sure we land on the same values as the serialized state.

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/config.py
```diff
@@ -56,15 +56,24 @@ def from_fortran_dict(
...
-        match testcase_nml.get("nh_test_name"):
-            case "jabw" | "jabw_s" | "APE_nwp":
+        match test_name:
+            case "jabw" | "jabw_s" | "APE_nwp" | "APE_aes":
                 log.info("Analytical initial condition for Jablonowski-Williamson test case")
                 config = fortran_config.config_dataclass_from_dict(
                     jw_ic.JablonowskiWilliamsonConfig, testcase_nml
                 )
+                # The APE cases rescale qv to a prescribed global moisture content;
+                # the jabw cases do not (Fortran passes opt_global_moist only for APE).
```
yiluchen1066:
```suggestion
                # Only the APE cases rescale qv to a prescribed global moisture content
```
to make it shorter

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/config.py
```diff
@@ -56,15 +56,24 @@ def from_fortran_dict(
...
+                config.normalize_global_moisture = test_name in ("APE_nwp", "APE_aes")
+                # Fortran resets the u-perturbation amplitude jw_up to 0 for the
+                # jabw_s/jabw_m cases only; the others keep the namelist default
+                # (1.0), see mo_nh_testcases.f90. (jabw_m is not handled above.)
```
yiluchen1066: not sure we should keep the documents in so detailed

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -49,6 +50,28 @@ class ProfilingStats:
     skip_first_timestep: bool = True
 
 
+#: ISO 8601 duration, restricted to the fixed-length components (weeks, days,
+#: hours, minutes, seconds). Years and months are intentionally not matched since
+#: their length is not fixed, and this is currently only used for dtime.
```
yiluchen1066:
```suggestion
# ISO 8601 duration, restricted to the fixed-length components (weeks, days,
# hours, minutes, seconds). Years and months are intentionally not matched since
# their length is not fixed, and this is currently only used for dtime.
```

model/standalone_driver/src/icon4py/model/standalone_driver/driver_io.py
```diff
@@ -213,33 +210,16 @@ def compute(
...
-            self._virtual_temperature,
-            self._surface_pressure,
-            self._pressure,
-            self._pressure_ifc,
-            horizontal_start=0,
-            horizontal_end=end_cell_end,
-            vertical_start=0,
-            vertical_end=num_levels,
-            offset_provider={},
+            surface_pressure_k=self._surface_pressure_k,
```
yiluchen1066: Note for later, here we still pass ```surface_pressure_k```, so it is still allocated. It is the same redundant buffer I pointed on the wrapper ```diagnose_surface_pressure_k```. I think we can remove the ```self._surface_pressure_k``` allocation after we the simplifcation in the wrapper

yiluchen1066 (commented):
mostly minor stuffs. So I would merge the current state into the physics interface pr and proceed from there. Thanks!

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
jcanton: it would be the only one with that suffix, and you/Chia Rui removed the `jw_` suffixes so I think it's more consistent without (and we have no suffixes in icon4py)

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/jablonowski_williamson.py
jcanton: friend is not very intelligent:
1. all tracers are moisture tracers
2. there is an explicit gate on `qv` already in place at line 284
`        if prognostic_state_now.tracer.qv is None:`

model/standalone_driver/src/icon4py/model/standalone_driver/initial_condition/analytical/utils.py
jcanton: from another friend inspecting fortran to make sure I did correctly:

>      Confirmed — caps are not re-applied after the global rescale.
> 
>      In init_nh_inwp_tracers:
> 
>      - Caps (qv_max, 5.e-6 stratosphere limit) are applied inside the iteration loop (lines 510–516).
>      - Global rescale happens after the loop ends (lines 571–572):
>        ptr_nh_prog%tracer(:,:,:,iqv) = ptr_nh_prog%tracer(:,:,:,iqv) &
>          &                           * opt_global_moist / z_moist
>      - No second cap pass follows. It only calls diagnose_pres_temp again when l_rediag.
> 
>      So post-rescale values may exceed the earlier caps, exactly as your Python code does. I added a comment in utils.py documenting this.
> 
>      Also checked mo_nh_ape_exp.f90 line 270: same uniform rescale, no cap re-application there either.

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

jcanton:
cscs-ci run merge

PR #1354, opened 2026-06-26
Add Weisman-Klemp (warm bubble) initial condition

jcanton:
replaces https://github.com/C2SM/icon4py/pull/1161

from a friend:

Adds the Weisman-Klemp (WK82) idealized warm-bubble convection test case to the `standalone_driver`, reconstructing the work from #1161 on top of the refactored `initial_condition/analytical` architecture (instead of rebasing that outdated branch). Based on `add_exp_ape_aes`.

## What's in here
- **`initial_condition/analytical/weisman_klemp.py`** — `WeismanKlempConfig` + `weisman_klemp()`, a host port of `mo_nh_wk_exp.f90`: isothermal layer above the tropopause, prescribed θ/RH profiles below, two-pass downward hydrostatic integration, sheared wind, hydrostatic adjustment, and the warm bubble. The hardcoded WK constants from #1161 are now config fields; the 8 `nh_testcase_nml` params (`qv_max_wk`, `u_infty_wk`, `bub_*`, `bubctr_*`) are mapped via `fortran_name_map`. **Fails hard if the `qv` tracer is inactive** — the moist base state requires it.
- **`common.thermodynamic_functions.specific_humidity`** (`spec_humi`) and the `RD_O_RV` constant; reuses the existing `sat_pres_water`.
- **`common.math.distance_array_ns.horizontal_distance_to_point`** — host cell-to-point distance with an opt-in periodic torus `wrap` (default off). `init_bubble` (torus + icosahedron) lives in the analytical utils, and the **gaussian-hill topography** is refactored to reuse the same helper.
- Wires `wk82` through the dispatcher and adds `WEISMAN_KLEMP_TORUS` to the unified initial-condition datatest, plus a unit test for the distance helper.

## ⚠️ Note on the torus distance / wrapping
ICON's `plane_torus_distance` has a units bug: it compares width-normalized bubble coordinates against the dimensional half-domain, so the periodic-wrap branch is never taken. Combined with non-origin-centered torus coords (`x∈[0,49621]`, `y∈[218,5030]`) and a bubble center at `(0,0)`, the reference data is a **non-periodic, corner-clipped** bubble. The helper therefore defaults to **no wrap** to bit-reproduce ICON; a "correct" minimum-image wrap diverges ~2.5% in the bubble footprint and fails the datatest. The `wrap=True` path is kept for future use. Worth a separate upstream bug report.

## Verification
- Pre-commit (ruff/tach/mypy/license) ✓
- Datatests (`gtfn_cpu`): `WEISMAN_KLEMP_TORUS` ✓ (rho/exner/θv/vn/w/qv within tolerance); `GAUSS3D` ✓ (no regression from the gaussian-hill refactor)
- Distance helper unit tests ✓

The WK case is marked `embedded_remap_error` (xfails under the default `embedded` backend, like the other torus experiments; verified under `gtfn_cpu`).


model/common/src/icon4py/model/common/math/distance_array_ns.py
```diff
@@ -0,0 +1,51 @@
...
+arrays, e.g. during initial-condition and topography setup.
+"""
+
+from __future__ import annotations
+
+from icon4py.model.common.utils import data_allocation as data_alloc
+
+
+def horizontal_distance_to_point(
+    *,
```
OngChia, line 59: There is a Python-loop version of this function `diff_on_edges_torus_numpy` in `projection.py`, which is used in computing the least-squares coefficients. Either I can move it to `common` while refactoring the least-squares computation (https://github.com/C2SM/icon4py/pull/1379) or you can move it now.

model/common/src/icon4py/model/common/math/distance_array_ns.py
```diff
@@ -0,0 +1,51 @@
...
+    wrap: bool = False,
+) -> data_alloc.NDArray:
+    """Horizontal distance from each point ``(x, y)`` to a fixed ``(point_x, point_y)``.
+
+    With ``wrap=False`` (the default) this is the plain Euclidean distance on the
+    plane. With ``wrap=True`` the distance is computed on a doubly-periodic torus
+    using the minimum-image convention, where ``domain_length`` and ``domain_height``
+    are the periodic extents in the x and y directions (both required in that case).
+
+    The default reproduces ICON: its ``plane_torus_distance`` (``mo_grid_utilities.f90``)
```
OngChia:
```suggestion
    The default reproduces ICON: its ``plane_torus_distance`` (``mo_math_utilities.f90``)
```

model/common/src/icon4py/model/common/math/distance_array_ns.py
```diff
@@ -0,0 +1,51 @@
...
+    y: data_alloc.NDArray,
+    point_x: float,
+    point_y: float,
+    domain_length: float | None = None,
+    domain_height: float | None = None,
+    wrap: bool = False,
+) -> data_alloc.NDArray:
+    """Horizontal distance from each point ``(x, y)`` to a fixed ``(point_x, point_y)``.
+
+    With ``wrap=False`` (the default) this is the plain Euclidean distance on the
```
OngChia: Although it is nice to consider the case where the torus domain is non-periodic and non-wrapped version of the distance function, but do we support this non-periodicity in dycore? ICON also does not support non-periodicity. I wonder whether we need this functionality if it will never be used.

model/common/src/icon4py/model/common/math/distance_array_ns.py
```diff
@@ -0,0 +1,51 @@
...
+    plane. With ``wrap=True`` the distance is computed on a doubly-periodic torus
+    using the minimum-image convention, where ``domain_length`` and ``domain_height``
+    are the periodic extents in the x and y directions (both required in that case).
+
+    The default reproduces ICON: its ``plane_torus_distance`` (``mo_grid_utilities.f90``)
+    is fed coordinates normalized by the feature width while the wrap threshold uses the
+    dimensional domain size, so the periodic branch is never taken and the effective
+    distance is non-periodic. Idealized torus test cases (e.g. the Weisman-Klemp warm
+    bubble and the Gaussian-hill topography) rely on this non-wrapping behaviour.
```
OngChia: I did not realize that Weisman-Klemp uses non-wrapping distance function. Is there a recent update in ICON related to this issue? Is it True for Gaussian hill experiment? I thought we always have periodic boundary condition with Torus.

model/common/tests/common/math/unit_tests/test_distance_array_ns.py
```diff
@@ -0,0 +1,40 @@
...
+    # periodic image, so its distance is small.
+    x = np.array([1.0, 9.0])
+    y = np.array([0.0, 0.0])
+    dist = distance_array_ns.horizontal_distance_to_point(
+        x=x, y=y, point_x=0.0, point_y=0.0, domain_length=10.0, domain_height=10.0, wrap=True
+    )
+    np.testing.assert_allclose(dist, [1.0, 1.0])
+
+
+def test_horizontal_distance_to_point_wrap_requires_domain_extents() -> None:
```
OngChia, line 54: Do we need this test? Isn't it trivial?

model/common/src/icon4py/model/common/initial_condition/analytical/utils.py
```diff
@@ -301,6 +309,84 @@ def _qv(temperature: data_alloc.NDArray) -> data_alloc.NDArray:
...
+            cell_lon = geometry.get(geometry_meta.CELL_LON).ndarray
+            center_lon = math.radians(center_x)
+            center_lat = math.radians(center_y)
+            central_angle = array_ns.arccos(
+                array_ns.sin(center_lat) * array_ns.sin(cell_lat)
+                + array_ns.cos(center_lat)
+                * array_ns.cos(cell_lat)
+                * array_ns.cos(cell_lon - center_lon)
+            )
+            horizontal_distance = phy_const.EARTH_RADIUS * central_angle
```
OngChia: There is a function (`gnomonic_proj`) doing this in `projection.py` (the docstring is not very good. I modify a bit in my PR https://github.com/C2SM/icon4py/pull/1379)

model/common/src/icon4py/model/common/initial_condition/analytical/weisman_klemp.py
```diff
@@ -0,0 +1,321 @@
...
+    # Sheared horizontal wind, projected onto the edge-normal direction.
+    wind_speed = config.max_wind_speed * (
+        array_ns.tanh((height - config.h_min) / (config.wind_scale_height - config.h_min)) - 0.45
+    )
+    interior_edge = array_ns.ones(num_edges, dtype=bool)
+    interior_edge[0 : zone_idx["end_edge_lateral_boundary_level_2"]] = False
+    prognostic_state_now.vn.ndarray[:, :] = (
+        array_ns.where(interior_edge[:, array_ns.newaxis], wind_speed[array_ns.newaxis, :], 0.0)
+        * primal_normal_x[:, array_ns.newaxis]
+    )
```
OngChia:
```suggestion
    boundary_lvl2 = zone_idx["end_edge_lateral_boundary_level_2"]
    prognostic_state_now.vn.ndarray[:boundary_lvl2, :] = 0.0
    prognostic_state_now.vn.ndarray[boundary_lvl2:, :] = wind_speed[array_ns.newaxis, :] * primal_normal_x[boundary_lvl2:, array_ns.newaxis]
```
Just a personal taste. Leave the decision to you whether to change it or not.

model/common/src/icon4py/model/common/initial_condition/analytical/weisman_klemp.py
```diff
@@ -0,0 +1,321 @@
...
+    h_tropopause: float = 12000.0
+    #: potential temperature at the surface [K] (theta_0_wk)
+    theta_surface: float = 300.0
+    #: potential temperature at the tropopause [K] (theta_tropo_wk)
+    theta_tropopause: float = 343.0
+    #: exponent of the potential-temperature profile below the tropopause (expo_theta_wk)
+    exponent_theta: float = 1.25
+    #: exponent of the relative-humidity profile below the tropopause (expo_relhum_wk)
+    exponent_relative_humidity: float = 1.25
+    #: temperature of the (isothermal) tropopause [K] (t_tropo_wk)
```
OngChia:
```suggestion
    #: temperature of the tropopause [K] (t_tropo_wk)
```

model/common/src/icon4py/model/common/math/distance_array_ns.py
jcanton: ? 

it is in `mo_grid_utilities`, line 529

jcanton:
cscs-ci run default;MODEL_SUBPACKAGES=standalone_driver;LEVELS=unit;BACKENDS=all;SESSIONS=model;NUM_PROCESSES=0;ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=true

model/common/src/icon4py/model/common/math/distance_array_ns.py
jcanton: not entirely sure but I moved `diff_on_edges_torus_numpy` into here

model/common/src/icon4py/model/common/math/distance_array_ns.py
jcanton: I think you answered your own question in the comment below?

model/common/src/icon4py/model/common/math/distance_array_ns.py
jcanton: also true for gaussian hill topography. no updates about this, you should just place the bubble and hill in places where they make sense and are not cut by the boundary ;-)
that's why I added the periodicity fix here as option

model/common/src/icon4py/model/common/math/distance_array_ns.py
OngChia: Looks like it was not there few months ago before I left, including the functionality to compute distance without wrapping. Interesting!

model/common/tests/common/math/unit_tests/test_distance_array_ns.py
jcanton: for sure it doesn't hurt. more tests is better than fewer tests. especially with the periodicity / non-periodicity option

model/common/src/icon4py/model/common/math/distance_array_ns.py
OngChia: I just realized this was merged two months ago https://gitlab.dkrz.de/icon/icon/-/commit/28c93c4c89275134edbdb913c8ae6fc088c24c09

jcanton:
cscs-ci run default;MODEL_SUBPACKAGES=standalone_driver;LEVELS=unit;BACKENDS=dace_cpu:gtfn_cpu:gtfn_gpu;SESSIONS=model;NUM_PROCESSES=0;ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=true;PYTEST_ADDOPTS=-k test_initial_conditions

model/common/src/icon4py/model/common/initial_condition/analytical/utils.py
jcanton: I merged the common functionality of `central_angle` and `cos_` into `distance_array_ns`

OngChia (commented):
Thanks! Looks really good overall. Before I approve it, I still think we can improve the torus distance calculation because the torus domain is doubly periodic, and the idealized test cases also assume periodic bc.

Although the bubble/mountain distance is normalized, so users should be aware not to place the bubble or mountain too close to the boundaries in ICON, I think we should avoid carrying over this somewhat unintuitive behavior into icon4py. In particular, the default value of `wrap` is `False`, which may lead to unexpected behavior if users later reuse this function to construct initial conditions.

My suggestion is that we reuse `diff_on_edges_torus_numpy` and pass `wrap` through there. Personally, I would prefer the default value of `wrap` to be `True`, and it may be better to make it a configurable parameter in the WK and Gauss3D experiment configs. Then, `horizontal_distance_to_point` could be generalized to a 3D `euclidean_distance_to_point` by using the x- and y-distance returned by `diff_on_edges_torus_numpy` together with the mountain/bubble radius in the x/y/z directions for normalization.

jcanton:
> Thanks! Looks really good overall. Before I approve it, I still think we can improve the torus distance calculation because the torus domain is doubly periodic, and the idealized test cases also assume periodic bc.
> 
> Although the bubble/mountain distance is normalized, so users should be aware not to place the bubble or mountain too close to the boundaries in ICON, I think we should avoid carrying over this somewhat unintuitive behavior into icon4py. In particular, the default value of wrap is False, which may lead to unexpected behavior if users later reuse this function to construct initial conditions.
> 
> My suggestion is that we reuse diff_on_edges_torus_numpy and pass wrap through there. Personally, I would prefer the default value of wrap to be True, and it may be better to make it a configurable parameter in the WK and Gauss3D experiment configs. Then, horizontal_distance_to_point could be generalized to a 3D euclidean_distance_to_point by using the x- and y-distance returned by diff_on_edges_torus_numpy together with the mountain/bubble radius in the x/y/z directions for normalization.

---

1. wrap is now a required keyword (no default) on horizontal_distance_to_point. That's the part of your concern I agree with: an implicit non-periodic default on a doubly-periodic torus was a bad idea. Now every call site has to state its choice, and both existing ones (init_bubble, gaussian_hill) pass `wrap=False` right next to a comment explaining why.
While you were away we decided default values are bad in general, so the more required the better. the risk with defaults is that the code runs without errors and you don't realize it's doing something differently from what you thought.

2. Reuse of `diff_on_edges_torus_numpy` I went the other way round: It was scalar Python (float in/out), driven by a per-cell loop in `compute_lsq_coeffs`; calling it from the bubble would have replaced a vectorized round() over all cells with a Python loop. Its name was also misleading: it returned the wrapped neighbour coordinate, not a difference.
So instead now there is a `minimum_image_offset(delta, extent)` helper, plus a vectorized `nearest_periodic_image` built on it. `horizontal_distance_to_point` uses the former, `compute_lsq_coeffs` the latter, and the per-cell loop is gone.

3. Config parameter: I'd skip it. Now that wrap is a required argument, adding it to the WK/Gauss3D configs would make it look like a physical property of the test case, which it isn't: it's a coding decision about which distance metric to use (wrong ICON or correct).

4. 3D `euclidean_distance_to_point`: not yet. nothing would use it now (and would make the two current usages require some dummy z field), we will implement it when needed.

jcanton:
cscs-ci run default;MODEL_SUBPACKAGES=standalone_driver;LEVELS=unit;BACKENDS=all;SESSIONS=model;PYTEST_ADDOPTS=-k test_initial_conditions

OngChia:
> > Thanks! Looks really good overall. Before I approve it, I still think we can improve the torus distance calculation because the torus domain is doubly periodic, and the idealized test cases also assume periodic bc.
> > Although the bubble/mountain distance is normalized, so users should be aware not to place the bubble or mountain too close to the boundaries in ICON, I think we should avoid carrying over this somewhat unintuitive behavior into icon4py. In particular, the default value of wrap is False, which may lead to unexpected behavior if users later reuse this function to construct initial conditions.
> > My suggestion is that we reuse diff_on_edges_torus_numpy and pass wrap through there. Personally, I would prefer the default value of wrap to be True, and it may be better to make it a configurable parameter in the WK and Gauss3D experiment configs. Then, horizontal_distance_to_point could be generalized to a 3D euclidean_distance_to_point by using the x- and y-distance returned by diff_on_edges_torus_numpy together with the mountain/bubble radius in the x/y/z directions for normalization.
> 
> 1. wrap is now a required keyword (no default) on horizontal_distance_to_point. That's the part of your concern I agree with: an implicit non-periodic default on a doubly-periodic torus was a bad idea. Now every call site has to state its choice, and both existing ones (init_bubble, gaussian_hill) pass `wrap=False` right next to a comment explaining why.
>    While you were away we decided default values are bad in general, so the more required the better. the risk with defaults is that the code runs without errors and you don't realize it's doing something differently from what you thought.
> 2. Reuse of `diff_on_edges_torus_numpy` I went the other way round: It was scalar Python (float in/out), driven by a per-cell loop in `compute_lsq_coeffs`; calling it from the bubble would have replaced a vectorized round() over all cells with a Python loop. Its name was also misleading: it returned the wrapped neighbour coordinate, not a difference.
>    So instead now there is a `minimum_image_offset(delta, extent)` helper, plus a vectorized `nearest_periodic_image` built on it. `horizontal_distance_to_point` uses the former, `compute_lsq_coeffs` the latter, and the per-cell loop is gone.
> 3. Config parameter: I'd skip it. Now that wrap is a required argument, adding it to the WK/Gauss3D configs would make it look like a physical property of the test case, which it isn't: it's a coding decision about which distance metric to use (wrong ICON or correct).
> 4. 3D `euclidean_distance_to_point`: not yet. nothing would use it now (and would make the two current usages require some dummy z field), we will implement it when needed.

I totally agree with point 1. I do not like default in this case. It is a good direction without default. I thought you must have the default and I prefer the default to be True. 
For point 2, sorry that I did not make it clear. I meant that you could move this line `dx = dx - domain_length * array_ns.round(dx / domain_length)` to replace the python loop in `diff_on_edges_torus_numpy` and `horizontal_distance_to_point` can then use `diff_on_edges_torus_numpy` (rename if possible) to compute the distance in x and y directions. Anyway, your current version does kind of what I intended to say.
For point 3, it just feels strange to me that wrap is hardcoded False in the initialization. I guess that the real solution is to fix ICON side. Anyway, no objection from me.
For point 4, I actually meant that init_bubble uses 3D distance with normalization factors. It is true that a dummy zero value for z is needed in Gaussian hill. 
So, this is how I was thinking in details inside `init_bubble` function based on your most updated functions' name:
```
normalized_distance = euclidean_3d_distance_to_point(
    cell_center_x,
    cell_center_y,
    cell_center_z,
    bubble_center_x,
    bubble_center_y,
    bubble_center_z,
    horizontal_width,
    vertical_width,
    domain_length,
    domain_height,
    False,
)

## in distance_array_ns.py
def euclidean_3d_distance_to_point(cell_center_x, cell_center_y, cell_center_z, reference_x, reference_y, reference_z, horizontal_width, vertical_width, domain_length, domain_height, wrap):
    dx, dy =  nearest_periodic_image(
        cell_center_x,
        cell_center_y,
        reference_x,
        reference_y,
        domain_length,
        domain_height) if wrap else cell_center_x - reference_x, cell_center_y - reference_y # I omit the exception statement here
    dz = cell_center_z - reference_z
    return sqrt(
        (dx / horizontal_width)**2 +
        (dy / horizontal_width)**2 +
        (dz / vertical_width)**2
    ) # or use array_ns.norm, I omit broadcast
```
Or, directly use `nearest_periodic_image` in init_bubble to compute dx and dy depending on wrap, and compute the `normalized_distance` from the output of `nearest_periodic_image`.

Anyway, I am okay with your current version.

OngChia (approved):

jcanton:
answer to above:

point 2: I re-structured going in your direction, now clearer
point 3: feel free do open a bug-fix in fortran 😜 then we open a corresponding change here

point 4:
- Sphere branch can't use it: `init_bubble` handles TORUS and ICOSAHEDRON. On the sphere the horizontal part is the great-circle distance * EARTH_RADIUS. A 3D-cartesian helper only covers the torus, so `init_bubble` keeps the `sqrt((h/horizontal_width)**2 + (dz/vertical_width)**2)` line for the sphere anyway -> normalization logic duplicated in two places instead of one.
- Shape mismatch: `x`, `y` are (cell,); `z_mc` is (cell, k). You say "I omit broadcast", but that is the actual work, and putting it in a generic math helper means having an IC-specific assumption into common/math.

I renamed domain_height/width to extent_x/y, may be better names and less confusing (your vertical_width made me think about this)

PR #1333, opened 2026-06-19
Add more experiments to standalone driver tests

msimberg:
Open questions:
- tracers in the r4b9 mch experiment?
- tolerance for r4b9 experiment compared to serialized data? it's very far off because of rbf coefficients (same order of magnitude as rbf coefficients, i.e. 1e-2)

Issue #967, opened 2025-12-04
Stencil tests: decrease all_close tolerance


havogt:
Currently the all_close tolerance is using the default of 1e-7, although our computations are in float64. We should decrease it to the same as in data tests and make it configurable per test/output. This might require generating smoother input data instead of fully random.

PR #1357, opened 2026-07-01
Automatic test-tolerance measurement and tightening (RBF pilot)

jcanton:
Related issue: #967 
from a friend:

## Automatic test-tolerance measurement and tightening (RBF pilot)

Numeric test tolerances are currently chosen by hand: run with
`ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=true`, read the printed max diffs, and paste values into
the tests. Nothing revisits a tolerance once a fix makes it too loose, so stale tolerances can hide
regressions.

This PR adds machinery to **measure** tolerances automatically and **tighten** them, and pilots it
on the RBF interpolation tolerances. The bulk of the ~300 inline `atol`/`rtol` literals elsewhere
are intentionally left for a follow-up.

### How it works
- **Record mode** (`ICON4PY_RECORD_TOLERANCES=<file>`): `dallclose`/`assert_dallclose` measure the
  max absolute/relative difference instead of asserting, tagged with `{backend, experiment, field}`
  via the pytest plugin, and dump to JSON lines.
- **Drift warning** (`ICON4PY_TOLERANCE_DRIFT_WARN`): a non-failing terminal summary flags
  tolerances that are much looser than the measured difference.
- **Updater** (`scripts/python/update_tolerances.py`, `nox -s update_tolerances`): aggregates
  measurements (**max over the deterministic CPU backends**, scaled by a safety factor, rounded up)
  and **only tightens** a stored tolerance — never loosens it.
- **CI (report-only):** a job in the weekly `all` pipeline runs the nox session on `gtfn_cpu`/
  `embedded` and publishes the proposed tightenings as a `tolerance-updates.patch` artifact. It
  never commits or pushes — a human reviews and applies the patch. (No GitHub token needed in CSCS
  CI; the `all` pipeline already has the serialized data, the backends and the weekly cadence.)

### Design notes
- **Backend spread:** deltas differ per backend; the updater takes the **highest** across the
  auto-managed backends, matching current practice. GPU/dace backends are excluded because they are
  documented as non-deterministic; their per-backend tolerances stay hand-maintained. The recorder
  keeps per-backend data in the JSONL, so a future per-backend store schema is a compatible
  extension.
- **Tighten-only** means the automated output can only make tests stricter, so it cannot mask a
  regression.
- The RBF store keeps the same `{dimension: {experiment: atol}}` shape, so consumers are unchanged.

### Validation (locally, real serialized data)
- Normal assert mode unchanged: RBF datatests pass on `embedded` and `gtfn_cpu`.
- Record mode captures correctly-keyed measurements on both backends.
- Updater correctly **no-ops** on the already-tight RBF tolerances; when entries were artificially
  loosened it tightened them from the measured diffs (e.g. `Vertex/gauss3d 1e-6 -> 9e-15`); the
  `dace_gpu` sample was ignored and MCH was never loosened.
- pre-commit (ruff/format/tach/mypy/…) is green.

### Follow-ups
- Wire more test suites into the store (e.g. `standalone_driver` prognostic-field tolerances once
  #1333 lands).
- Add a notification (Slack) when the weekly job proposes changes — the artifact is the concrete
  deliverable; a webhook secret would let CI ping instead of relying on manual monitoring.

Draft: opening for review of the approach before wiring more test suites into the store.


jcanton:
may benefit from rel delta computation from https://github.com/C2SM/icon4py/pull/1333 as well

model/common/tests/common/interpolation/unit_tests/test_rbf_interpolation.py
```diff
@@ -41,23 +43,12 @@
...
-    },
-    dims.VertexDim: {
-        definitions.Experiments.EXCLAIM_APE.name: 3e-10,
-        definitions.Experiments.MCH_CH_R04B09.name: 3e-3,
-        definitions.Experiments.GAUSS3D.name: 1e-15,
-    },
-}
+# Absolute tolerances for the RBF interpolation coefficients, keyed by dimension and experiment
+# name. Maintained via the tolerance-tightening tooling (see 'scripts/python/update_tolerances.py');
+# edit 'rbf_tolerances.json' rather than hardcoding values here.
```
nfarabullini, line 48:
```suggestion
# Absolute tolerances for the RBF interpolation coefficients, maintained via the tolerance-tightening tooling.
```

model/testing/src/icon4py/model/testing/config.py
```diff
@@ -33,5 +33,15 @@ def _default_download_cache() -> pathlib.Path:
 DALLCLOSE_PRINT_INSTEAD_OF_FAIL: bool = env.flag_to_bool(
     "ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL", False
 )
+# When set to a file path, 'assert_dallclose' records the measured max absolute/relative
+# differences (instead of asserting) so that tolerances can be measured and tightened automatically.
```
nfarabullini, line 37:
```suggestion
```
You don't need this

model/testing/src/icon4py/model/testing/config.py
```diff
@@ -33,5 +33,15 @@ def _default_download_cache() -> pathlib.Path:
...
+RECORD_TOLERANCES_PATH: pathlib.Path | None = (
+    env.path("ICON4PY_RECORD_TOLERANCES", pathlib.Path())
+    if "ICON4PY_RECORD_TOLERANCES" in os.environ
+    else None
+)
+# When enabled, 'assert_dallclose' emits a non-failing warning whenever a passed tolerance is much
+# larger than the measured difference, flagging tolerances that have become too loose.
```
nfarabullini, line 44:
```suggestion
```
Neither this

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -51,6 +51,10 @@ def pytest_configure(config):
     if config.getoption("--datatest-skip"):
         config.option.markexpr = " and ".join(["not datatest", *m_option])
 
+    # Activate tolerance recording/drift detection when the corresponding options are enabled.
+    if testing_config.RECORD_TOLERANCES_PATH is not None or testing_config.TOLERANCE_DRIFT_WARN:
```
nfarabullini, line 55:
```suggestion
    if testing_config.RECORD_TOLERANCES_PATH or testing_config.TOLERANCE_DRIFT_WARN:
```
I think you can also just say this

model/testing/src/icon4py/model/testing/tolerances.py
```diff
@@ -0,0 +1,211 @@
...
+]
+
+# A stored tolerance is flagged as drifted (too loose) when it exceeds the measured difference by
+# more than this factor.
+DRIFT_FACTOR = 100.0
+# Proposed tolerances are the measured difference scaled by this factor to leave headroom.
+SAFETY_FACTOR = 4.0
+# Backends whose results are deterministic enough that measured differences may be used to tighten
+# tolerances automatically. GPU and dace backends are excluded (documented as non-deterministic).
+DETERMINISTIC_CPU_BACKENDS = frozenset({"gtfn_cpu", "embedded"})
```
nfarabullini, line 58: can you summarize comments?

model/testing/src/icon4py/model/testing/tolerances.py
```diff
@@ -0,0 +1,211 @@
...
+
+
+def max_differences(actual: npt.ArrayLike, desired: npt.ArrayLike) -> tuple[float, float]:
+    """Return the maximum absolute and relative difference between 'actual' and 'desired'."""
+    actual_array = np.asarray(actual, dtype=float)
+    desired_array = np.asarray(desired, dtype=float)
+    absolute = np.abs(actual_array - desired_array)
```
nfarabullini, line 65:
```suggestion
    absolute = np.abs(actual_array - desired_array)
```
Maybe you can dimply do this?

model/testing/src/icon4py/model/testing/tolerances.py
```diff
@@ -0,0 +1,211 @@
...
+    """Holds the process-wide active recorder (avoids a module-level 'global')."""
+
+    active: ToleranceRecorder | None = None
+
+
+def get_active_recorder() -> ToleranceRecorder | None:
+    return _RecorderState.active
+
+
+def activate_recorder() -> ToleranceRecorder:
```
nfarabullini, line 148:
```suggestion
def activate_recorder() -> ToleranceRecorder | None:
```

model/testing/src/icon4py/model/testing/tolerances.py
```diff
@@ -0,0 +1,211 @@
...
+
+    The measured difference is scaled by 'SAFETY_FACTOR' and rounded up to one significant figure.
+    A measured difference of zero yields an exact (zero) tolerance.
+    """
+    scaled = max_abs * SAFETY_FACTOR
+    if scaled <= 0.0:
+        return 0.0
+    exponent = math.floor(math.log10(scaled))
+    fraction = scaled / 10.0**exponent
+    return math.ceil(fraction) * 10.0**exponent
```
nfarabullini, line 211: why all of this is effectively this is the same as:
```
return math.ceil(scaled)
```

nfarabullini (commented):
some initial comments

PR #1359, opened 2026-07-02
Port tmx turbulence

jcanton:
gist with intial part of conversation 
https://gist.github.com/jcanton/463e411efeef3ad5963d554b14b38a89

PR #1360, opened 2026-07-07
Physics driver tmx

yiluchen1066:
Plugs the TMX turbulence granule (#1359) into the ```PhysicsDriver``` as the second physics process, **parallel-coupled** with muphys under the two-layer state design of #1436.

This branch includes: 

**tmx ComponentState adapter** which is the same as the muphys pattern: 
- In `state.py`:  `collect_inputs` binds the frozen `EntryState` , derives `air_mass`/`cv_air`, runs the surface-flux provider; `as_component_input` maps the 21-key component contract. 
- `data.py` defines a Field metadata for the TmxComponent input/output contract. Every output's metadata carries a kind - the 7 tendencies get ```kind="tendency"``` from the shared ```tendency_of()``` helper, and the diagnostics are explicitly tagged ```kind=diagnostic```
- In ```physics_driver.py```, after each process computes, the driver makes two calls with the outputs and their metadata:  accumulators.accumulate(...), which takes output with ```kind=tendency```, and ```diagnostics_of()``` stores the diagnostic output into ```driver.diagnostics[process.name]```
- `component.py`defines  the wrapper ```TmxComponent```  and ```static_fields.py``` for factory-built states.

Cherry picked @jcanton's `PrescribedFluxProvider`: fixed kinematic heat fluxes, ≈ −83 W/m² sensible, zero latent. In ```config.py``` parses the `nh_testcase_nml` flux members and the constant aquaplanet SST (303.15 K); `driver_utils` binds the provider to the entry state's live `pressure_ifc`. Other `isrfc_type` values fall back to the
  zero-flux seam default.
  
tmx auto-enable from the ```aes_vdf_nml``` namelist. Re-enables the ```EXCLAIM_APE_AES``` end-to-end - with the physics integration, the case validates again in assert mode with measured tolerances.

**Validation (v08, gtfn_cpu)** ```test_driver[EXCLAIM_APE_AES]``` passes in the assertion mode.


yiluchen1066:
cscs-ci run default

yiluchen1066:
cscs-ci run default

model/common/src/icon4py/model/common/physics/thermodynamics.py
```diff
@@ -0,0 +1,230 @@
...
+
+class ThermodynamicConstants(ta.wpfloat, enum.Enum):
+    """Thermodynamic constants of mo_aes_thermo.f90 (values as used by muphys)."""
+
+    cvd = 717.60  # [J/K/kg] specific heat of dry air at constant volume => cpd - rd
+    cvv = 1407.95  # [J/K/kg] specific heat of water vapor at constant volume => cpv - rv
+    clw = 4192.6641119999995  # specific heat capacity of liquid water => (rcpl + 1.0) * cpd
+    ci = 2108.0  # specific heat of ice
+    lvc = 3135383.2031928  # invariant part of vaporization enthalpy => alv - (cpv - clw) * tmelt
+    lsc = 2899657.201  # invariant part of sublimation enthalpy => als - (cpv - ci) * tmelt
```
OngChia, line 41: Most of these constants are already defined in `constants.py`. I think we can just use the constants or missing constants there (in ICON, the physical constants are all stored in `mo_physical_constants.f90`).

model/common/src/icon4py/model/common/physics/thermodynamics.py
```diff
@@ -0,0 +1,230 @@
...
+    dz: ta.wpfloat,  # Extent of grid cell
+    temperature: ta.wpfloat,  # output
+):
+    _T_from_internal_energy_scalar(
+        u=u, qv=qv, qliq=qliq, qice=qice, rho=rho, dz=dz, out=temperature
+    )
+
+
+@gtx.field_operator
+def _internal_energy(  # noqa: PLR0917 [too-many-positional-arguments] GT4Py operator parameters are positional
```
OngChia, line 151: We already have some diagnostic calculations in `common/diagnostic_calculations/stencils`. I think it may be better to keep all those diagnostic calculations in one place (either move those in `common/diagnostic_calculations/stencils` to here or vice versa)?

Note: energy or diagnostic calculations are always tricky. For example, there are different definitions for internal energy. So, there will be multiple ways to compute the same quantity. Usually, a whole physics suite sticks to one formulation (aes may use one calculation and nwp uses another). It depends on "philosophy" of the the parameterizations or the dycore algorithm.

OngChia (commented):
Looks good in a broad sense. I am still less than half way through. I left two comments first.

PR #1368, opened 2026-07-10
Enable bitwise-identical GPU results for standalone driver and grid manager MPI tests

msimberg:
This PR enables bitwise-identical single-rank vs multi-rank results for the standalone driver and grid manager MPI tests on GPU backends.

The main source of non-determinism was the RBF interpolation coefficient computation on GPU, where batched CuPy linear algebra produces results that depend on the rank-local batch size (and therefore on the domain decomposition). The PR adds a dispatcher in `rbf_interpolation.py` that routes the coefficient solve through NumPy on the host when `ICON4PY_DETERMINISTIC_RBF_COEFFS` is set, while keeping the GPU solve as the production default. The environment flag is now cached at module import.

GPU FMA contraction was the second source of divergence. The PR disables it via `NVCC_APPEND_FLAGS="--fmad=false"` for the MPI validation jobs in `ci/base.yml`, alongside `CXXFLAGS="-ffp-contract=off"` for the host code. The test harness now applies zero tolerance to both CPU and GPU `gtfn` and `dace` backends when `ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE=1`, and the grid manager MPI tests use this helper.

The PR also adds the `z_nabla2_e` edge halo exchange in the diffusion module before the second RBF interpolation, matching the ICON Fortran reference (`mo_nh_diffusion.f90:853`).


msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu:dace_gpu;LEVELS=integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu:dace_gpu;LEVELS=integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu:gtfn_cpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu,gtfn_cpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

ci/base.yml
```diff
@@ -179,7 +179,22 @@ variables:
...
+        # related tooling is unused). Disable FMA contraction (GPU
+        # counterpart of -ffp-contract=off) and pin IEEE-compliant
+        # division/sqrt. Only this flag is needed for bitwise identical
+        # GPU results; CUPY_ACCELERATORS, CUBLAS_WORKSPACE_CONFIG, and
+        # CUDAFLAGS were bisected out as unnecessary.
+        NVCC_APPEND_FLAGS: "--fmad=false -prec-div=true -prec-sqrt=true"
+        # Override the .test_template_aarch64 default ("persistent") so the
+        # cache is scoped to this session, ensuring kernels are freshly
+        # compiled.
+        GT4PY_BUILD_CACHE_LIFETIME: "session"
```
msimberg: Remove.

ci/base.yml
```diff
@@ -179,7 +179,22 @@ variables:
...
+        # while the reproducibility flags mature.
+        ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: "1"
         CXXFLAGS: "-ffp-contract=off"
+        # GPU reproducibility flag (no-op on CPU backends where the
+        # related tooling is unused). Disable FMA contraction (GPU
+        # counterpart of -ffp-contract=off) and pin IEEE-compliant
+        # division/sqrt. Only this flag is needed for bitwise identical
+        # GPU results; CUPY_ACCELERATORS, CUBLAS_WORKSPACE_CONFIG, and
+        # CUDAFLAGS were bisected out as unnecessary.
+        NVCC_APPEND_FLAGS: "--fmad=false -prec-div=true -prec-sqrt=true"
```
msimberg: To do: check if all are necessary.

ci/base.yml
```diff
@@ -179,7 +179,22 @@ variables:
         # compiler flags like -ffp-contract=off and others that may be required to
         # actually get bitwise identical results.
         ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE: "1"
+        # Print max diff per field instead of hard-failing on the first
+        # non-zero mismatch, so remaining CPU/GPU divergences are observable
+        # while the reproducibility flags mature.
+        ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: "1"
```
msimberg: To do: remove.

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu:dace_gpu;LEVELS=integration;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=dace_gpu;LEVELS=integration;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration:validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration:validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=common:standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu;LEVELS=unit:integration
from a friend: single- vs multi-rank comparison of the static (geometry/interpolation/metrics) fields via test_parallel_grid_manager, on both CPU and GPU with the reproducibility flags and print-instead-of-fail enabled. Purpose: pinpoint the seed of the residual ~1.3e-12/step vn diff. Suspect: the batched cupy.linalg.solve in rbf_interpolation.py (batch size = rank-local vertex/edge count); note the rbf-specific atol=3e-9 in test_parallel_grid_manager.py dates from the commit that added gtfn_gpu to the distributed pipeline (#1012). The standalone_driver integration run doubles as a regression check that the bisection revert restored the original diffusion behaviour.

jcanton:
from a friend: results of the static-field comparison run (pipeline 8896/8897, all 8 jobs green, print-instead-of-fail):

**gtfn_gpu, single- vs multi-rank, `test_parallel_grid_manager` — the only nonzero fields are the RBF coefficients:**

| field | internal max diff |
|---|---|
| rbf_interpolation_coefficient_vertex_1 | 1.51e-10 |
| rbf_interpolation_coefficient_vertex_2 | 1.45e-10 |
| rbf_interpolation_coefficient_cell_1 | 1.62e-09 |
| rbf_interpolation_coefficient_cell_2 | 1.88e-09 |
| rbf_interpolation_coefficient_edge | **0.0** |
| every other geometry/interpolation/metrics field | **0.0** |

**gtfn_cpu**: everything 0.0, including all RBF fields (also reproduced locally on macOS/arm64, 4 ranks).

This pinpoints the seed of the residual GPU diff: the batched `cupy.linalg.solve` in `rbf_interpolation.py` (`_compute_rbf_interpolation_coeffs`) is not bitwise batch-size-independent, and the batch is the rank-local vertex/cell count, so single- and multi-rank runs (and different rank counts) get different coefficients. The edge systems (4×4) come out identical, which neatly explains why the dycore (edge RBF only, for vt) has been bitwise clean all along while diffusion (vertex RBF → u_vert/v_vert → nabla2 → nabla4) is not. It is also consistent with the rbf-specific atol=3e-9 in test_parallel_grid_manager.py dating from #1012, the commit that added gtfn_gpu to the distributed pipeline.

The full causal chain, all measured in CI: Δrbf_coeff_v (~1e-10) → Δu_vert/v_vert → Δz_nabla2_e ~6e-22 → Δz_nabla4_e2 ~5.8e-31 → ×(diff_multfac_vn·area_edge²)≈2.3e18 → Δvn 1.3073986337985843e-12 per step → chaotic growth ×~100 over 7 days → 1.9e-10. The same 17-digit vn diff reappears after reverting all stencil experiments (job 15396211241), confirming the stencils were never the problem — no formula restructuring needed once the coefficients are deterministic.

Fix options, in order of confidence: (a) compute the RBF (and other solver-based) coefficients with numpy on the host at init and transfer — proven deterministic by the CPU runs; (b) keep the GPU solve but chunk it to a fixed batch size identical on all ranks — cheaper, but needs a small cupy probe to confirm per-matrix results depend only on chunk size; (c) per-point Cholesky as in ICON Fortran. After the fix the rbf/pos_on_tplane tolerance paddings in test_parallel_grid_manager.py should be removable.

Caveat for the end goal: in the JW setup kh_smag_e is clamped to zero everywhere, so the smag branch of diffusion is never exercised by this test — worth a config with active Smagorinsky diffusion before declaring GPU bit-reproducibility done.

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=common:standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu;LEVELS=unit:integration
from a friend: validation of the deterministic-RBF fix (ICON4PY_DETERMINISTIC_RBF_COEFFS=1 now set in the reproducibility CI block). Prediction: all static-field comparisons bitwise zero on both backends, and the standalone driver integration test vn diff drops from 1.3073986337985843e-12 to exactly 0.0 on gtfn_gpu.

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=common:standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu;LEVELS=unit:integration
from a friend: retrigger — previous pipeline (8905) died in build_baseimage_aarch64 with a runner infra error (pod start timeout), no test ran.

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=common:standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu;LEVELS=unit:integration
from a friend: retrigger 2 — pipelines 8905 and 8907 both died in build_baseimage_aarch64 with 'timed out waiting for pod to start' (runner infra, no test ran).

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=common:standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu;LEVELS=unit:integration
from a friend: fix v2 — pipeline 8908 showed the rbf diffs persisting with slightly changed values, i.e. the host solve took effect but the batched matrix assembly (cupy matmul) and normalization reductions are batch-extent-dependent too. ICON4PY_DETERMINISTIC_RBF_COEFFS now moves the whole coefficient computation to the host. Same prediction as before: all statics bitwise zero, driver gtfn_gpu integration vn 0.0.

jcanton:
from a friend: **fix confirmed — bitwise identical GPU results achieved** (pipeline 8918/8919, all 8 jobs green).

With `ICON4PY_DETERMINISTIC_RBF_COEFFS=1` on gtfn_gpu, single- vs multi-rank:

| | before fix | after fix |
|---|---|---|
| rbf_coeff_vertex_1/2 (statics) | 1.5e-10 | **0.0** |
| rbf_coeff_cell_1/2 (statics) | 1.6-1.9e-9 | **0.0** |
| all other static fields | 0.0 | 0.0 |
| driver integration `vn` internal | 1.3073986337985843e-12 | **0.0** |
| driver `w`/`exner`/`theta_v`/`rho` | 0.0 | 0.0 |

Every field, every halo level, bitwise zero. gtfn_cpu also all zero (was already). Default (flag off) keeps the batched cupy solve, so production is unchanged.

**What v2 corrected over v1:** the first attempt moved only `linalg.solve` to the host and the diffs *persisted with slightly different values* (vertex_1 1.5080215e-10 → 1.5080198e-10) — proving the batched matrix assembly (cupy `matmul` in the pairwise dot products) and the axis-1 reductions in the normalization are batch-extent-dependent too, not just the solve. v2 runs the whole `_compute_rbf_interpolation_coeffs` on numpy under the flag; only bit-exact copies cross the device boundary.

Implication for options (b)/(c) from earlier: since assembly + solve + normalization are each decomposition-sensitive on cupy, a fixed-chunk GPU approach would have to chunk the entire computation, not just the solve — so the host path is the cleaner validation mechanism unless someone wants that larger change.

Two follow-ups for whoever picks this up:
1. The reproducibility jobs still set `ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1`, so bitwise equality is *printed, not asserted*. Now that it's actually zero, that can be dropped so the tests fail on regression.
2. In this JW/APE setup `kh_smag_e` is clamped to zero everywhere, so the Smagorinsky branch of diffusion is never exercised — worth a config with active smag diffusion before declaring GPU bit-reproducibility complete across all of diffusion.

jcanton:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=validation
from a friend: 7-day validation run of the deterministic-RBF fix on gtfn_gpu. Integration (1 step) is already all-zero; this confirms the fix holds under accumulation over a full simulation, which is the PR's actual target. Prediction: all fields bitwise 0.0 at 7 days (was vn 1.91e-10). Still print-not-fail, so this only measures — if zero, the print-instead-of-fail flag can be dropped to make it a hard assertion.

jcanton:
from a friend: **7-day validation confirms the fix under accumulation** (pipeline 8932/8933, gtfn_gpu, LEVELS=validation, 40 min).

Full JW simulation, single- vs multi-rank, every field bitwise identical:

| field | before | after (7 days) |
|---|---|---|
| vn internal | 1.91e-10 | **0.0** |
| vn / w / exner / theta_v / rho, all halo levels | — | **0.0** |

So the fix eliminates the diff not just at step 1 (integration) but across the full ~2000-step run — the residual really was seeded entirely by the RBF coefficients, with nothing else in the model contributing once they're deterministic.

**Remaining item before this can become a hard gate:** the reproducibility jobs still set `ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1`, so equality is printed, not asserted. Dropping it would make CI fail on regression — but one caveat to decide first: the repro variable block is keyed on subpackage/level, not backend, and dace has its own known non-bitwise kernel behaviour (cf. the `atol=2e-13` "dace undeterministic" TODO in test_parallel_interpolation). `ICON4PY_DETERMINISTIC_RBF_COEFFS` only makes the RBF coefficients deterministic, not dace's kernels, so a global hard assertion is safe for gtfn but could break a dace_gpu repro job. Suggest scoping the bitwise gate to gtfn (or excluding dace) before removing print-instead-of-fail. Happy to make that change once you decide the backend scope.

Branch summary (4 commits on top of the bisection history): revert of the stencil experiments keeping only the z_nabla2_e halo exchange; repro flags extended to the common MPI tests; the opt-in `ICON4PY_DETERMINISTIC_RBF_COEFFS` host-solve for RBF coefficients; tolerances routed through `get_mpi_comparison_tolerance`.

msimberg:
> and dace has its own known non-bitwise kernel behaviour (cf. the atol=2e-13 "dace undeterministic" TODO in test_parallel_interpolation). ICON4PY_DETERMINISTIC_RBF_COEFFS only makes the RBF coefficients deterministic, not dace's kernels, so a global hard assertion is safe for gtfn but could break a dace_gpu repro job. Suggest scoping the bitwise gate to gtfn (or excluding dace) before removing print-instead-of-fail. Happy to make that change once you decide the backend scope.

I think by now we've concluded both have some indeterminism, whether from code generation (dace) or at runtime (dace and gtfn). Both should be able to go to zero diff now that we have RBF identified. I'd say we get this PR in for standalone driver validation tests with zero diffs, and then follow up with making more of `test_parallel_grid_manager.py` have zero diffs.

ci/base.yml
```diff
@@ -172,14 +172,48 @@ variables:
   script:
     - ci/scripts/ci-mpi-wrapper.sh nox -s "test_model_mpi-${PYVERSION_SHORT}($SELECTION, $MODEL_MPI_SUBPACKAGE)" -- --backend=$BACKEND --level=$LEVEL
   rules:
-    - if: $MODEL_MPI_SUBPACKAGE == 'standalone_driver' && $LEVEL == 'validation'
+    # The 'common' branch of this rule enables the same reproducibility
+    # setup for the static-field single- vs multi-rank comparison tests
+    # (test_parallel_grid_manager.py), which are level 'unit'/'integration'.
```
msimberg: Useless comment, let's remove.
```suggestion
```

ci/base.yml
```diff
@@ -172,14 +172,48 @@ variables:
...
       variables:
         # Signal to (selected) tests that they should expect bitwise identical
         # results between single- and multi-rank runs. This goes hand-in-hand with
         # compiler flags like -ffp-contract=off and others that may be required to
         # actually get bitwise identical results.
         ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE: "1"
+        # Print max diff per field instead of hard-failing on the first
+        # non-zero mismatch, so remaining CPU/GPU divergences are observable
+        # while the reproducibility flags mature.
+        ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL: "1"
```
msimberg:
```suggestion
```

ci/base.yml
```diff
@@ -172,14 +172,48 @@ variables:
...
+        # Compute RBF interpolation coefficients with a host (numpy) linear
+        # solve instead of batched cupy.linalg.solve, whose results depend on
+        # the batch extent (= rank-local element count) and therefore on the
+        # domain decomposition. Required for bitwise identical single- vs
+        # multi-rank results on GPU; production runs keep the GPU solve.
+        ICON4PY_DETERMINISTIC_RBF_COEFFS: "1"
         CXXFLAGS: "-ffp-contract=off"
+        # GPU reproducibility flags (no-op on CPU backends where the related
+        # tooling is unused):
```
msimberg:
```suggestion
        # GPU reproducibility flags:
```

ci/base.yml
msimberg:
```suggestion
```

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -892,6 +892,21 @@ def run(
...
+        # values from neighbouring ranks' interior, matching ICON Fortran
+        # (mo_nh_diffusion.f90: sync_patch_array(SYNC_E, ..., z_nabla2_e)).
+        log.debug("communication of z_nabla2_e - start")
+        self._exchange.exchange(
+            dims.EdgeDim,
+            self.z_nabla2_e,
+            stream=decomposition.DEFAULT_STREAM,
+        )
+        log.debug("communication of z_nabla2_e - end")
+
```
msimberg: Verify before merging: is this required right now? And if it's required, but didn't notice before, can we have a test that actually catches this being required?

SPEC.md
```diff
@@ -0,0 +1,133 @@
...
+| File                                                                                                                 | Type of change                                                              |
+| -------------------------------------------------------------------------------------------------------------------- | --------------------------------------------------------------------------- |
+| `model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py`                                     | Add `z_nabla4_e2` buffer, compile `calculate_nabla4` program, call sequence |
+| `model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/calculate_nabla4.py`                     | Revert bisection code, optionally convert Pattern B → A                     |
+| `model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/apply_diffusion_to_vn.py`                | Add `z_nabla4_e2` param, remove internal `_calculate_nabla4` call           |
+| `model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/stencils/apply_nabla2_and_nabla4_global_to_vn.py` | Possibly change z_nabla4_e2 type (Phase 2b-1)                               |
```
msimberg: Delete this file.

model/common/tests/common/grid/mpi_tests/test_parallel_grid_manager.py
```diff
@@ -351,18 +353,29 @@ def _compare_interpolation_fields_single_multi_rank(
     field = multi_rank_interpolation.get(attrs_name)
     dim = field_ref.domain.dims[0]
 
+    # Default tolerances cover the GPU runtime computation of the rbf (and
+    # pos_on_tplane) coefficients, which is not bitwise decomposition
+    # independent (batched cupy.linalg.solve). With
+    # ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE (plus reproducible compiler flags
+    # and ICON4PY_DETERMINISTIC_RBF_COEFFS) the comparison is bitwise.
```
msimberg:
```suggestion
```

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -295,6 +296,18 @@ def _cartesian_coordinates_from_zonal_and_meridional_components(
...
+    rhs_batch: data_alloc.NDArray,
+    array_ns: ModuleType,
+) -> data_alloc.NDArray:
+    # rhs_batch is expanded to 3D so both numpy and cupy treat it as a
+    # batched column vector (core dims (nv, 1)) rather than a matrix
+    # (core dims (B, nv)), which would mismatch m=nv from the LHS.
+    # This problem is well explained in https://github.com/numpy/numpy/issues/26598
+    return array_ns.linalg.solve(mat_batch, rhs_batch[..., array_ns.newaxis]).squeeze(-1)
+
+
```
msimberg: I don't think this needs to be moved out to a separate function, it was just fine where it was.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -319,6 +332,40 @@ def _compute_rbf_interpolation_coeffs(
...
+            rbf_offset=data_alloc.as_numpy(rbf_offset),
+            rbf_kernel=rbf_kernel,
+            geometry_type=geometry_type,
+            scale_factor=scale_factor,
+            horizontal_start=horizontal_start,
+            horizontal_end=horizontal_end,
+            domain_length=domain_length,
+            domain_height=domain_height,
+        )
+        return tuple(array_ns.asarray(coeff) for coeff in host_coeffs)
```
msimberg: It's a bit ugly (IMO) that `_compute_rbf_interpolation_coeffs` is entered twice and the namespace is checked twice in the case that deterministic RBF coeffs are enabled and the namespace is cupy. I'd rather see the dispatch logic move to a helper outside of `_compute_rbf_interpolation_coeffs` (or have `_compute_rbf_interpolation_coeffs` become the dispatcher).

ci/base.yml
```diff
@@ -172,27 +172,49 @@ variables:
   script:
     - ci/scripts/ci-mpi-wrapper.sh nox -s "test_model_mpi-${PYVERSION_SHORT}($SELECTION, $MODEL_MPI_SUBPACKAGE)" -- --backend=$BACKEND --level=$LEVEL
   rules:
-    - if: $MODEL_MPI_SUBPACKAGE == 'standalone_driver' && $LEVEL == 'validation'
+    - if: ($MODEL_MPI_SUBPACKAGE == 'standalone_driver' && ($LEVEL == 'validation' || $LEVEL == 'integration')) || ($MODEL_MPI_SUBPACKAGE == 'common' && ($LEVEL == 'unit' || $LEVEL == 'integration'))
```
msimberg:
```suggestion
    - if: $MODEL_MPI_SUBPACKAGE == 'standalone_driver' || $MODEL_MPI_SUBPACKAGE == 'common'
```

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=unit:integration:validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=unit:integration:validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=unit:integration:validation

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration;ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration;ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1;GT4PY_BUILD_CACHE_LIFETIME=session;SLURM_TIMELIMIT=06:00:00

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_gpu;LEVELS=integration;ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;MODEL_MPI_SUBSETS=datatest;BACKENDS=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=unit:integration:validation;GT4PY_BUILD_CACHE_LIFETIME=session

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -465,6 +465,89 @@ def index_offset(f):
...
+    horizontal_start: gtx.int32,
+    horizontal_end: gtx.int32,
+    domain_length: ta.wpfloat,
+    domain_height: ta.wpfloat,
+) -> tuple[data_alloc.NDArray, ...]:
+    array_ns = data_alloc.array_namespace(element_center_lat)
+    # array_namespace returns array_api_compat wrapper modules, so detect the
+    # device by name rather than by module identity.
+    on_gpu = "cupy" in getattr(array_ns, "__name__", "")
+    if on_gpu and env.flag_to_bool("ICON4PY_DETERMINISTIC_RBF_COEFFS", False):
```
msimberg: This should go in a global, read once.

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGE=standalone_driver;BACKEND=gtfn_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session;CUDAFLAGS=;NVCC_APPEND_FLAGS=--fmad=false;CUPY_ACCELERATORS=;CUBLAS_WORKSPACE_CONFIG=

🤖 *Posted by an agent on behalf of @msimberg*.

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGE=standalone_driver:common;BACKEND=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;BACKEND=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver:common;BACKENDS=gtfn_cpu:gtfn_gpu:dace_cpu:dace_gpu;LEVELS=validation;GT4PY_BUILD_CACHE_LIFETIME=session

🤖 *Posted by an agent on behalf of @msimberg*.

ci/base.yml
```diff
@@ -172,14 +172,24 @@ variables:
...
         # results between single- and multi-rank runs. This goes hand-in-hand with
         # compiler flags like -ffp-contract=off and others that may be required to
         # actually get bitwise identical results.
         ICON4PY_TEST_EXPECT_MPI_REPRODUCIBLE: "1"
+        # Compute RBF interpolation coefficients with a host (numpy) linear
+        # solve instead of batched cupy.linalg.solve, whose results depend on
+        # the batch extent (= rank-local element count) and therefore on the
+        # domain decomposition. Required for bitwise identical single- vs
+        # multi-rank results on GPU; production runs keep the GPU solve.
+        ICON4PY_DETERMINISTIC_RBF_COEFFS: "1"
```
msimberg, line 187: This name may deserve bikeshedding, though I'm relatively happy with the current name.

On one hand, I think this is a debug/validation option, not something I would like/hope for users to use all the time. I also considered adding something like "host RBF" to the name, but then it describes how it does it rather than what it's trying to achieve. We may find a way restructure the RBF computation so that it uses cupy and still gives bitwise identical results, and then I wouldn't want to change the name just for that.

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -465,6 +468,89 @@ def index_offset(f):
...
+    geometry_type: icon_grid.GeometryType,
+    scale_factor: ta.wpfloat,
+    horizontal_start: gtx.int32,
+    horizontal_end: gtx.int32,
+    domain_length: ta.wpfloat,
+    domain_height: ta.wpfloat,
+) -> tuple[data_alloc.NDArray, ...]:
+    array_ns = data_alloc.array_namespace(element_center_lat)
+    # array_namespace returns array_api_compat wrapper modules, so detect the
+    # device by name rather than by module identity.
```
nfarabullini:
```suggestion
```
I don't think we need this explanation here

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -465,6 +468,89 @@ def index_offset(f):
...
+    on_gpu = "cupy" in getattr(array_ns, "__name__", "")
+    if on_gpu and _DETERMINISTIC_RBF_COEFFS:
+        # cupy's batched kernels (matmul in the matrix assembly, linalg.solve,
+        # axis reductions in the normalization) are not bitwise
+        # batch-size-independent, and the batch extent is the rank-local
+        # element count: on GPU the coefficients depend on the domain
+        # decomposition. Compute them entirely on the host instead; input and
+        # output copies are bit-exact, so the result is decomposition
+        # independent at the cost of a device round trip at initialization.
```
nfarabullini: if this could be made smaller, that would be better. Otherwise no biggie

ci/base.yml
jcanton: I think the name is fine

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
```diff
@@ -465,6 +468,84 @@ def index_offset(f):
     return rbf_vec_coeff
 
 
+def _compute_rbf_interpolation_coeffs_dispatch(
```
jcanton, line 471: curiosity on my side, decision left to you:
why avoid the `_dispatch` name and handle the logic
```
    array_ns = data_alloc.array_namespace(element_center_lat)
    on_gpu = "cupy" in getattr(array_ns, "__name__", "")
    if on_gpu and _DETERMINISTIC_RBF_COEFFS:
         copy_to_host
```
directly inside `_compute_rbf_interpolation_coeffs`?
avoid the rename at call sites and the presence of a single `_dispatch` function?

jcanton (approved):
just one question more than a request for change — non-blocking

model/common/src/icon4py/model/common/interpolation/rbf_interpolation.py
msimberg: I guess
> why avoid the _dispatch

is meant to be

> why not avoid the _dispatch

? If yes, I find it easier to follow the logic wrapping the copying to and from the GPU around a small function call rather than the full RBF computation. It's not just `copy_to_host`, but they have to be copied back afterwards to the GPU as well. The actual RBF computation is purely doing the work based on whichever namespace the arrays are in, and the extra logic for dispatching is separated out. At least to me the logic is easier to follow this way.

jcanton: yes, there was a "not" missing :-)
as I wrote it was curiosity, as in "maybe there's a software-engineering reason that I don't know"
so go ahead and queue 🚀

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;BACKEND=gtfn_gpu:gtfn_cpu;LEVELS=integration

msimberg:
cscs-ci run default;SESSIONS=model_mpi;MODEL_MPI_SUBPACKAGES=standalone_driver;BACKENDS=gtfn_gpu:gtfn_cpu;LEVELS=integration

PR #1369, opened 2026-07-13
Standalone driver: derive do_prep_adv from ltransport

jcanton:
from a friend:

Part 1 of a stack that makes the standalone driver reproduce MCH_CH_R04B09. Stacked on `standalone-driver-more-tests` (#1333).

## Problem

`run_driver` hardcodes `time_integration(ds, do_prep_adv=False)`. ICON derives it (`mo_nh_stepping.f90`, `perform_dyn_substepping`):

```fortran
IF ( ltransport .OR. p_patch%n_childdom > 0 .AND. grf_intmethod_e == 6 ) THEN
  lprep_adv = .TRUE.
```

ICON4Py has no nested domains, so `lprep_adv == ltransport`. That is `true` for MCH_CH_R04B09, EXCLAIM_APE_AES and Weisman-Klemp — and the MCH savepoints record `prep_adv: true` in their metadata.

## Change

`DriverConfig.do_prep_adv`, read from `run_nml.ltransport`, threaded to `solve_nonhydro.time_step(lprep_adv=...)` from the config instead of through the `time_integration` → `_integrate_one_time_step` → `_do_dyn_substepping` parameter chain.

## Impact

`lprep_adv` only gates the tracer-advection quantities (`vn_traj`, `mass_flx_me`, `mass_flx_ic`), not the prognostic fields, so the dry testcases are unchanged. Verified: JW and GAUSS3D driver datatests still pass against the tolerances measured on `standalone-driver-more-tests` (gtfn_cpu).

PR #1370, opened 2026-07-13
Standalone driver: fix the spin-up second order divergence damping

jcanton:
from a friend:

Part 2 of the MCH_CH_R04B09 stack. Stacked on #1369.

## Problem

`update_spinup_damping` (`mo_nh_stepping.f90`):

```fortran
time1 = 1800._wp ; time2 = 7200._wp
IF (elapsed_time <= time1) THEN
  divdamp_fac_o2 = 8._wp*divdamp_fac
ELSE IF (elapsed_time <= time2) THEN
  divdamp_fac_o2 = 8._wp*divdamp_fac*(time2-elapsed_time)/(time2-time1)
ELSE
  divdamp_fac_o2 = 0._wp
ENDIF
```

called **once per model time step**, under `divdamp_order == 24` and `elapsed_time_global = (jstep-0.5)*dtime <= 7200 + 0.5*dtime .AND. .NOT. ltestcase`.

The driver had four defects:

1. `ADJUST_FACTOR_FOR_SECOND_ORDER_DIVDAMP = 0.8`, ICON uses `8.0`.
2. The ramp was inverted: `(elapsed - 1800)/(7200 - 1800)` instead of `(7200 - elapsed)/(7200 - 1800)`, i.e. it ramped *up* where ICON ramps *down*.
3. It was evaluated per dynamics substep rather than once per time step.
4. No `divdamp_order` guard.

Defect 1 is not just a magnitude error. `solve_nonhydro` gates the fourth-order damping on `second_order_divdamp_factor <= 4 * fourth_order_divdamp_factor` (mirroring ICON's `divdamp_fac_o2 <= 4._wp*divdamp_fac`), whose purpose is to *disable* fourth-order damping during spin-up. With `0.8` that condition was always true, so the driver applied fourth-order damping where ICON does not, together with a 10x too small second-order coefficient.

## Change

- The constant becomes `8.0`.
- `driver_utils.spinup_second_order_divdamp_factor` is a pure function mirroring `update_spinup_damping`, with the decreasing ramp.
- `ModelTimeVariables.elapsed_time_at_step_midpoint_in_seconds` reproduces ICON's `(jstep - 0.5)*dtime` (the driver advances the datetime before integrating the step).
- The factor is computed once per time step in `time_integration` and threaded down to `solve_nonhydro.time_step`, with ICON's guards.

For `divdamp_order != 24` the factor now stays at `divdamp_fac`, as in `mo_nonhydrostatic_nml.f90`, where `divdamp_fac_o2` is initialized to `divdamp_fac` and then never updated at runtime. No serialized experiment uses that path (all are `divdamp_order = 24`).

## Verification

New unit tests pin `8 * 0.004 = 0.032` — exactly the `divdamp_fac_o2` recorded in every MCH `solve-nonhydro-init` savepoint — and the decreasing ramp. The JW and GAUSS3D driver datatests are unchanged (`ltestcase = true`, so the factor is `0.0` before and after).

The dycore is untouched: it already implements ICON's `-0.25*divdamp_fac_o2` clip (`dycore_utils.py`) and the `<= 4*divdamp_fac` gate (`solve_nonhydro.py`).

### DIFF
--- a/model/standalone_driver/src/icon4py/model/standalone_driver/driver_constants.py
+++ b/model/standalone_driver/src/icon4py/model/standalone_driver/driver_constants.py
@@ -18,8 +18,8 @@ CFL_THRESHOLD_FACTOR = ta.wpfloat("0.9")
 #: Factor multiplied to the user-defined CFL number to determine the whether to leave watchmode
 CFL_LEAVE_WATCHMODE_FACTOR = ta.wpfloat("0.76")
 
-#: Adjustment factor for second order divergence damping
-ADJUST_FACTOR_FOR_SECOND_ORDER_DIVDAMP = ta.wpfloat("0.8")
+#: Adjustment factor for second order divergence damping (update_spinup_damping in mo_nh_stepping.f90)
+ADJUST_FACTOR_FOR_SECOND_ORDER_DIVDAMP = ta.wpfloat("8.0")
 
 #: Initial period for second order divergence damping
 INITIAL_PERIOD_FOR_SECOND_ORDER_DIVDAMP = ta.wpfloat("1800.0")
--- a/model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
+++ b/model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
@@ -464,17 +464,6 @@ class Icon4pyDriver:
     def _second_order_divdamp_factor(self) -> ta.wpfloat:
         """
         Second order divergence damping factor (divdamp_fac_o2) for the current time step.
-
-        mo_nh_stepping.f90, in the time loop, before integrate_nh:
-
-            IF (divdamp_order==24) THEN
-              elapsed_time_global = (REAL(jstep,wp)-0.5_wp)*dtime
-              IF (elapsed_time_global <= 7200._wp+0.5_wp*dtime .AND. .NOT. ltestcase) THEN
-                CALL update_spinup_damping(elapsed_time_global)
-              ELSE
-                divdamp_fac_o2 = 0._wp
-              ENDIF
-            ENDIF
         """
         assert self.config.nonhydrostatic is not None
         fourth_order_divdamp_factor = self.config.nonhydrostatic.fourth_order_divdamp_factor
--- a/model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
+++ b/model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
@@ -191,7 +191,6 @@ class Icon4pyDriver:
                     prognostic_states=prognostic_states,
                     prep_adv=prep_adv,
                     tracer_prep_adv=tracer_prep_adv,
-                    second_order_divdamp_factor=self._second_order_divdamp_factor(),
                 )
                 device_utils.sync(self.backend)
 
@@ -229,7 +228,6 @@ class Icon4pyDriver:
         prognostic_states: common_utils.TimeStepPair[prognostics.PrognosticState],
         prep_adv: dycore_states.PrepAdvection | None,
         tracer_prep_adv: advection_states.AdvectionPrepAdvState | None,
-        second_order_divdamp_factor: ta.wpfloat,
     ) -> None:
         if self.config.nonhydrostatic is not None:
             assert solve_nonhydro_diagnostic_state is not None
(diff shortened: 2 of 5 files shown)


model/standalone_driver/src/icon4py/model/standalone_driver/driver_constants.py
```diff
@@ -18,8 +18,8 @@
 #: Factor multiplied to the user-defined CFL number to determine the whether to leave watchmode
 CFL_LEAVE_WATCHMODE_FACTOR = ta.wpfloat("0.76")
 
-#: Adjustment factor for second order divergence damping
-ADJUST_FACTOR_FOR_SECOND_ORDER_DIVDAMP = ta.wpfloat("0.8")
+#: Adjustment factor for second order divergence damping (update_spinup_damping in mo_nh_stepping.f90)
+ADJUST_FACTOR_FOR_SECOND_ORDER_DIVDAMP = ta.wpfloat("8.0")
```
OngChia, line 23: Thanks for catching my mistake!

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -191,6 +191,8 @@ def time_integration(
                     prognostic_states=prognostic_states,
                     prep_adv=prep_adv,
                     tracer_prep_adv=tracer_prep_adv,
+                    # updated once per time step, as in mo_nh_stepping.f90
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -456,39 +461,46 @@ def _adjust_ndyn_substeps_var(
...
+        mo_nh_stepping.f90, in the time loop, before integrate_nh:
+
+            IF (divdamp_order==24) THEN
+              elapsed_time_global = (REAL(jstep,wp)-0.5_wp)*dtime
+              IF (elapsed_time_global <= 7200._wp+0.5_wp*dtime .AND. .NOT. ltestcase) THEN
+                CALL update_spinup_damping(elapsed_time_global)
+              ELSE
+                divdamp_fac_o2 = 0._wp
+              ENDIF
+            ENDIF
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -344,7 +348,7 @@ def _do_dyn_substepping(
                     diagnostic_state_nh=solve_nonhydro_diagnostic_state,
                     prognostic_states=prognostic_states,
                     prep_adv=prep_adv,
-                    second_order_divdamp_factor=self._update_spinup_second_order_divergence_damping(),
+                    second_order_divdamp_factor=second_order_divdamp_factor,
```
OngChia, line 350: Can I know what is the benefit of passing this function `second_order_divdamp_factor` through a chain of calls (from `integrate_one_time_step` to here) instead of putting `self._second_order_divdamp_factor()` here

jcanton: the function is called once per time-step, not every substep, here it's only passing the value computed when calling `_integrate_time_step` https://github.com/C2SM/icon4py/pull/1370/changes#diff-0b0c11c272113e20f2b29fbce83f44dafcb8220123e1961b51fc994013941a41R194

OngChia: In that case, is it better if moving it to the line before `self._do_dyn_substepping(` because it is only used in solve_nonhydro, or make it `cached_property`? (not sure it works automatically or not)

jcanton: yeah, maybe that's a better idea!

jcanton: btw: `cached_property` does not work: the factor is a function of elapsed time, and caching it would freeze step 1's value (8 * divdamp_fac) for the entire run

OngChia (approved):

OngChia:
Looks good to me!

jcanton:

PR #1371, opened 2026-07-13
Standalone driver: restart the time loop from a serialized state

jcanton:
from a friend:

Part 3 of the MCH_CH_R04B09 stack. Stacked on #1370.

## Problem

The from-file initial condition always reads the single un-dated `prognostics / initial-state` savepoint, and `ModelTimeVariables` hardcodes `is_first_step_in_simulation = True`. The second MCH parametrization of `test_standalone_driver` is meant to start at `12:00:10`, so today it runs **one step from the 12:00:00 state** and compares the result against the **two-step** reference at 12:00:20. It is off by a whole time step, and it also tells the dycore it is at the initial time step when it is not.

## Change

`DriverConfig.start_of_simulation` keeps its meaning — the beginning of the experiment — and a new `start_of_timestepping` marks the beginning of the time loop. It defaults to the start of the simulation; setting it later means a restart. From it, `ModelTimeVariables` derives:

- `simulation_current_datetime` and `n_time_steps` (the time loop starts there),
- `elapsed_time_in_seconds = start_of_timestepping - start_of_simulation`, because ICON measures `elapsed_time_global` from the beginning of the simulation — this is what keeps the spin-up divergence damping of #1370 correct across a restart,
- `is_first_step_in_simulation`, which is ICON's `linit_dyn`: false when restarting.

`FromFileConfig` gains the driver clock and owns the savepoint selection through an `is_restart` property. `read_restart_from_file` reads, from the savepoints of the time step that starts at `start_of_timestepping` (they are stamped with the date of their *end*, hence `start_of_timestepping + dtime`):

- the prognostic variables and `exner_pr` from `solve-nonhydro-init[istep=1, dyn_timestep=1]`,
- the advective tendencies of the previous corrector from `velocity-tendencies-init` — the dycore reuses them in the first substep of a non-initial step, so the vertical pair is stored swapped.

This mirrors ICON: on `isRestart()` it reads exactly these from the restart file and skips `compute_exner_pert`. The perturbed exner pressure therefore becomes the initial condition's responsibility: the `exner - exner_ref_mc` computation moves out of `assemble_driver_states` into `initial_condition.create`, which now receives the dycore fields it initializes. They are passed as a small `DycoreInitialFields` container because `icon4py.model.common` cannot depend on the dycore package.

Restarting with tracers raises `NotImplementedError`: the solve-nonhydro savepoints do not carry them. Reaching past the serialized window raises with the dates that are available.

## Verification

[...]

PR #1372, opened 2026-07-13
Standalone driver: diffuse the initial state of a real data run

jcanton:
from a friend:

Part 4 of the MCH_CH_R04B09 stack. Stacked on #1371.

## Problem

ICON makes one extra diffusion call, with `linit=.TRUE.`, on the initial state before the first time step of a real-data run (`mo_nh_stepping.f90`, `integrate_nh`):

```fortran
! For real-data runs, perform an extra diffusion call before the first time
! step because no other filtering of the interpolated velocity field is done
IF (ldynamics .AND. .NOT.ltestcase .AND. linit_dyn(jg) .AND. &
    diffusion_config(jg)%lhdiff_vn .AND. init_mode /= MODE_IAU) THEN
  CALL diffusion(p_nh_state(jg)%prog(nnow(jg)), ..., dt_loc, .TRUE.)
ENDIF
```

The standalone driver never made it. This is the single largest error in MCH's first time step: in the reference data, that call alone moves `vn` by **4.1 m/s** and `w` by **4.1 m/s** — the same scale as the driver's disagreement with the reference.

## Change

- `DriverConfig.diffuse_before_time_loop`, derived from `not ltestcase and ldynamics`.
- `Icon4pyDriver._diffuse_before_time_loop` runs the diffusion granule with `initial_run=True` (the granule already implements ICON's enhanced initial coefficients) on the **current** state with the **full model time step**, before the time loop. It is guarded by the flag, by `apply_to_horizontal_wind` (ICON's `lhdiff_vn`), and by `is_first_step_in_simulation` (ICON's `linit_dyn`), so a restart correctly skips it.
- `ltestcase` and `ldynamics` now fall back to ICON's namelist defaults (`.TRUE.`, `mo_run_nml.f90`) instead of treating a missing `ltestcase` as a real-data run. Latent today — every serialized namelist sets it — but the old default was inverted.

ICON's `init_mode /= MODE_IAU` term is not ported: the incremental analysis update does not exist in ICON4Py. Noted in the code.

## Verification

Unit tests for the derived flag, including the namelist defaults. All integration datatests except the MCH driver pair pass unchanged on gtfn_cpu (7 passed): the flag is `False` for every testcase experiment, so nothing moves.

The MCH driver cases segfault on macOS inside a gtfn program (`solve_nonhydro.run_predictor_step`), independently of this stack — they need a Linux run.


model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -86,6 +86,8 @@ class DriverConfig:
     apply_extra_second_order_divdamp: bool = False
     # lprep_adv in fortran. The default matches the namelist default of ltransport.
     do_prep_adv: bool = False
+    # Extra diffusion call before the time loop, for real data runs.
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -123,11 +128,18 @@ def from_fortran_dict(
...
             # lprep_adv = ltransport .OR. (n_childdom > 0 .AND. grf_intmethod_e == 6)
             # There are no nested domains in ICON4Py, so lprep_adv is ltransport.
             do_prep_adv=run_nml.get("ltransport", False),
+            # mo_nh_stepping.f90 (integrate_nh), where linit_dyn is only true at the
+            # first time step of a simulation that is not a restart, and lhdiff_vn is
+            # checked by the driver against the configuration of the diffusion granule:
+            # IF (ldynamics .AND. .NOT.ltestcase .AND. linit_dyn(jg) .AND.
+            #     diffusion_config(jg)%lhdiff_vn .AND. init_mode /= MODE_IAU)
+            # The incremental analysis update (MODE_IAU) is not implemented in ICON4Py.
+            diffuse_before_time_loop=is_real_data_run and runs_dynamics,
```
OngChia: Minor suggestion, personal flavor
```suggestion
            diffuse_before_time_loop=is_real_data_run and run_dynamics,
```
?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -109,6 +111,9 @@ def from_fortran_dict(
         )
         start_datetime_str = master_time_control_nml["experimentstartdate"]
         end_datetime_str = master_time_control_nml["experimentstopdate"]
+        # The fortran defaults of mo_run_nml.f90 are used for the missing variables.
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -109,6 +111,9 @@ def from_fortran_dict(
...
+        is_real_data_run = not run_nml.get("ltestcase", True)
```
jcanton:
```suggestion
        is_testcase = run_nml["ltestcase"]
```
the keys are always present, leftover from before.
and I prefer to keep the testcase logic

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -109,6 +111,9 @@ def from_fortran_dict(
...
+        runs_dynamics = run_nml.get("ldynamics", True)
```
jcanton:
```suggestion
```
not necessary, already gated by the Granule being not None

OngChia:
```suggestion
        run_dynamics = run_nml.get("ldynamics", True)
```
?

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -123,11 +128,18 @@ def from_fortran_dict(
             # apply_extra_second_order_divdamp does not have a namelist
             # variable in fortran. It is coded as follows in mo_nh_stepping.f90:
             # IF (elapsed_time_global <= 7200._wp+0.5_wp*dtime .AND. .NOT. ltestcase)
-            apply_extra_second_order_divdamp=not run_nml.get("ltestcase", False),
+            apply_extra_second_order_divdamp=is_real_data_run,
```
jcanton:
```suggestion
            apply_extra_second_order_divdamp= not is_testcase,
```

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
jcanton:
```suggestion
            diffuse_before_time_loop=not is_testcase,
```
and move up close to `apply_extra_second_order_divdamp`

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -461,6 +463,38 @@ def _adjust_ndyn_substeps_var(
             0.0, self._allocator
         )
 
+    def _diffuse_before_time_loop(
+        self,
+        diffusion_diagnostic_state: diffusion_states.DiffusionDiagnosticState | None,
+        prognostic_state: prognostics.PrognosticState,
+    ) -> None:
+        """
+        Extra diffusion call before the first time step of a real data run.
```
jcanton:
```suggestion
        Extra diffusion call before the first time step.
```

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -461,6 +463,38 @@ def _adjust_ndyn_substeps_var(
...
+
+        mo_nh_stepping.f90 (integrate_nh): 'For real-data runs, perform an extra
+        diffusion call before the first time step because no other filtering of the
+        interpolated velocity field is done'. It is called on the current state, with
+        the model time step, and not for a restart (linit_dyn).
```
jcanton:
```suggestion
        For real-data runs, perform an extra
        diffusion call before the first time step because no other filtering of the
        interpolated velocity field is done. It is called on the current state, with
        the model time step, and not for a restart.
```

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -461,6 +463,38 @@ def _adjust_ndyn_substeps_var(
...
+        """
+        if not self.config.driver.diffuse_before_time_loop:
+            return
+        if not self.model_time_variables.is_first_step_in_simulation:
+            return
+        if self.granules.diffusion is None:
+            return
+        # lhdiff_vn in fortran
+        if not self.granules.diffusion.config.apply_to_horizontal_wind:
+            return
```
OngChia, line 481: I do not really like a lot of `if`s here. I think that there should be only one if to check whether `diffuse_before_time_loop` this is `True`, and in the config setup we derive `diffuse_before_time_loop` based on whether users want to have this extra diffusion and other factors such as restart run, diffusion component is not called. And we can remove or move the comments here to config. But since configuration is not ready yet, I am okay with this.

jcanton: can't fully work: nonhydro granule presence and is_first_step_in_simulation are runtime facts, not config...
and all others are necessary now that granules can be optional.
the only alternative is that we move all these in a separate helper `do_diffusion_before_time_loop` but it would be identical to this and just move away the logic from the call site? I'm not sure I like it.

OngChia: I was thinking that we may have a hierarchy of configs in the future and there is a master config such as the `DriverConfig`. In he master config, we can have a parameter `diffuse_before_time_loop`, but it can be overwritten if dycore component is turned off or we run from restart or diffusion on `vn` is turned off, and then a warning/debug message is printed. So here, we can have one `if`. Although `is_first_step_in_simulation` is a runtime variable, this function is only called once and we can check during initialization whether we run from restart (`is_first_step_in_simulation` will be `False` in restart)

OngChia: By the way, I see some assertions such as `assert diffusion_diagnostic_state is not None` or `assert solve_nonhydro_diagnostic_state is not None` in standalone driver. Some of these assertions sit inside `if self.config.nonhydrostatic is not None:`. So I guess there will be a refactoring so that there is a config parameter to turn off component instead of checking the diagnostic state is existent or not.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -460,6 +462,37 @@ def _adjust_ndyn_substeps_var(
...
+        prognostic_state: prognostics.PrognosticState,
+    ) -> None:
+        """
+        Extra diffusion call before the first time step.
+
+        For real-data runs, perform an extra diffusion call before the first time step
+        because no other filtering of the interpolated velocity field is done. It is
+        called on the current state, with the model time step, and not for a restart.
+        """
+        # ldynamics and lhdiff_vn in fortran are the granules being configured
```
OngChia: Maybe we can remove this comment as it is unclear to me

model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -131,7 +133,14 @@ def from_fortran_dict(
...
             # IF (elapsed_time_global <= 7200._wp+0.5_wp*dtime .AND. .NOT. ltestcase)
-            apply_extra_second_order_divdamp=not run_nml["ltestcase"],
+            apply_extra_second_order_divdamp=not is_testcase,
+            # mo_nh_stepping.f90 (integrate_nh), where linit_dyn is only true at the first
+            # time step of a simulation that is not a restart, and both ldynamics and
+            # lhdiff_vn are checked by the driver against its granules:
+            # IF (ldynamics .AND. .NOT.ltestcase .AND. linit_dyn(jg) .AND.
+            #     diffusion_config(jg)%lhdiff_vn .AND. init_mode /= MODE_IAU)
+            # The incremental analysis update (MODE_IAU) is not implemented in ICON4Py.
+            diffuse_before_time_loop=not is_testcase,
```
OngChia, line 143: I like the comments here because we see ICON as the reference answer, although being worried that the code will be flooded with comments as we port more parameters..

OngChia (approved):

OngChia:
Looks good to me, except that the meaning of a comment is not clear to me and consider removing it or modifying it

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton: move 
```
            or self.granules.solve_nonhydro is None
            or self.granules.diffusion is None
            or not self.granules.diffusion.config.apply_to_horizontal_wind
```
into config and override `config.driver.diffuse_before_time_loop`

jcanton:

PR #1373, opened 2026-07-13
Standalone driver: prescribe the lateral boundary and slow physics tendencies

jcanton:
from a friend:

Part 5 of the MCH_CH_R04B09 stack. Stacked on #1372.

## Problem

Six fields are *inputs* of the dycore (`solve_nonhydro` reads them in the predictor and the corrector) but the driver leaves them at zero:

| field | ICON | who computes it |
|---|---|---|
| `grf_tend_rho`, `grf_tend_thv`, `grf_tend_w`, `grf_tend_vn` | lateral boundary of a limited-area domain | `compute_boundary_tendencies` (`mo_async_latbc_utils.f90`) — the time difference of two lateral-boundary slices, refreshed when a new slice is read |
| `ddt_exner_phy`, `ddt_vn_phy` | slow physics (radiation, convection, SSO, GWD) | `mo_nh_interface_nwp.f90`, once per time step, held constant across the dynamics substeps |

ICON4Py has neither a lateral-boundary reader nor NWP physics, so for a real-data run these cannot be computed — and they are not small: `ddt_vn_phy` alone is worth ~0.57 m/s of `vn` over one 10 s MCH step.

## Change

A new `prescribed_tendencies` module reads all six from the serialized reference data, once per model time step (the savepoints of a step are stamped with the date of its *end*, and the tendencies are constant over its substeps). This is what the old driver did (`model/driver/.../initialization_utils.py`), except it read them once instead of per step.

The testcases are neither limited-area nor forced by physics, so their config carries no data path and the reader is `None` — their tendencies stay zero and their results are unchanged.

Reaching past the serialized window raises, listing the dates that exist. Note this makes a *default* MCH run (whose namelist asks for 6 time steps, of which only 2 are serialized) fail at step 3 instead of silently integrating on zeroed physics — deliberate, but worth a look if you'd rather it degraded quietly.

## Verification

All integration datatests except the MCH pair pass unchanged on gtfn_cpu. With this PR and the `keep_skip_values` fix that follows it, MCH finally reproduces the reference:

| | before the stack | after |
|---|---|---|
| `vn` | 4.16 / 11.45 m/s | 3.4e-3 / 2.2e-3 |
| `w` | 2.15 / 5.54 m/s | 6.9e-4 / 7.8e-10 |
| `theta_v` | 0.48 / 10.0 K | 1.0e-3 / 1.2e-4 |
| `rho` | 2.1e-3 / 1.9e-2 | 3.1e-6 / 1.3e-12 |


model/standalone_driver/src/icon4py/model/standalone_driver/config.py
```diff
@@ -155,6 +156,9 @@ class ExperimentConfig:
     topography: topography.TopographyConfig
     initial_condition: initial_condition.InitialConditionConfig
     driver: DriverConfig
+    prescribed_tendencies: prescribed_tendencies.PrescribedTendenciesConfig = dataclasses.field(
+        default_factory=prescribed_tendencies.PrescribedTendenciesConfig
+    )
```
jcanton:
```suggestion
    prescribed_tendencies: prescribed_tendencies.PrescribedTendenciesConfig
```
no defaults, like the rest. assign explicitly. intent is better than hidden errors

model/common/src/icon4py/model/common/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+    mo_async_latbc_utils.f90) and refreshes them whenever a new slice is read.
+  - ddt_exner_phy and ddt_vn_phy are the tendencies of the slow physics (radiation,
+    convection, sub-grid scale orography, gravity wave drag), computed once per time
+    step by the NWP interface (mo_nh_interface_nwp.f90) and held constant over the
+    dynamics substeps.
+
+Without a lateral boundary reader and without physics, both sets can only be
+prescribed from the serialized data of the reference run, once per time step. They
+are zero for the testcases, which are neither limited area nor forced by physics.
+"""
```
jcanton, line 24:
```suggestion
"""
The dycore is forced with two sets of tendencies that come from parts of ICON which
ICON4Py does not have:

  - grf_tend_rho, grf_tend_thv, grf_tend_w and grf_tend_vn drive the lateral boundary
    of a limited area domain. ICON computes them as the time difference of two
    consecutive lateral boundary data slices (compute_boundary_tendencies in
    mo_async_latbc_utils.f90) and refreshes them whenever a new slice is read.
  - ddt_exner_phy and ddt_vn_phy are the tendencies of the slow physics (radiation,
    convection, sub-grid scale orography, gravity wave drag), computed once per time
    step by the NWP interface (mo_nh_interface_nwp.f90) and held constant over the
    dynamics substeps.

Without a lateral boundary reader and without physics, both sets can only be
prescribed from the serialized data of the reference run, once per time step.
"""
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+
+
+log = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class PrescribedTendenciesConfig:
+    """Parameters for the tendencies read from the serialized data."""
+
+    #: Path to the serialized data directory. None for the testcases, whose tendencies are zero.
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+    from icon4py.model.common.decomposition import definitions as decomposition_defs
+    from icon4py.model.common.grid import icon as icon_grid
+
+
+log = logging.getLogger(__name__)
+
+
+@dataclasses.dataclass
+class PrescribedTendenciesConfig:
+    """Parameters for the tendencies read from the serialized data."""
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+    data_path: pathlib.Path | None = None
+
+    @classmethod
+    def from_fortran_dict(
+        cls,
+        *,
+        atm_dict: dict[str, Any],
+        data_path: pathlib.Path,
+    ) -> PrescribedTendenciesConfig:
+        run_nml = atm_dict.get("run_nml", {})
```
jcanton:
```suggestion
        run_nml = atm_dict["run_nml"]
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+        if run_nml.get("ltestcase", True):
```
jcanton:
```suggestion
        if run_nml["ltestcase"]:
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+            return cls(data_path=None)
+        return cls(data_path=data_path / fortran_config.SER_DATA_SUBDIR)
+
+
+class SerializedTendencies:
+    """Reader of the prescribed tendencies of a time step."""
```
jcanton:
```suggestion
```

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+
+    def _available_dates(self) -> str:
+        dates = sorted(
+            {
+                savepoint.metainfo["date"]
+                for savepoint in self._serializer.savepoint_list()
+                if savepoint.name == "solve-nonhydro-init"
+            }
+        )
+        return ", ".join(dates) if dates else "none"
```
jcanton: same comment as in the other PR: give this a nonhydro-related name since it only returns those dates

model/standalone_driver/src/icon4py/model/standalone_driver/prescribed_tendencies.py
```diff
@@ -0,0 +1,178 @@
...
+) -> SerializedTendencies | None:
+    """The reader of the prescribed tendencies, or None when they are zero."""
+    if config.data_path is None:
+        return None
+    return SerializedTendencies(
+        data_path=config.data_path,
+        grid=grid,
+        backend=backend,
+        rank=exchange.my_rank(),
+    )
```
jcanton: do we need a separate `create` method or `__init__` is enough?

OngChia (approved):
Looks good to me. I am concerned about how the config in the standalone driver and the `PrescribedTendencies` component (I would like to see it designed as a component that returns certain tendencies, similar to other physics) can be refactored. Since adding more features to the driver to run simulations is the top goal, I think we can leave it in the future.

For the time being, perhaps we can initialize `PrescribedTendencies` with `PrescribedTendenciesConfig` instead of treating them as independent classes. I will leave the decision to you.

jcanton:

PR #1374, opened 2026-07-13
Standalone driver: replace the skip values of the grid (out-of-bounds read)

jcanton:
from a friend:

Part 6 of the MCH_CH_R04B09 stack. Stacked on #1373. **This one is a real out-of-bounds read, independent of the rest of the stack.**

## Problem

`driver_utils.create_grid_manager` builds the grid with `keep_skip_values=True`, which keeps the `-1` entries in the neighbor tables. On the limited-area MCH grid, `E2C` has **428** of them — exactly the outermost row of lateral-boundary edges (`E2C[e][1] == -1` for `e < 428`).

The dycore computes over the full horizontal domain (`horizontal_start = 0` for `compute_rho_theta_pgrad_and_update_vn`) and dereferences the neighbor tables **directly** — GT4Py only inserts `can_deref` guards for *reductions*, not for plain `field(E2C[1])`. So `compute_horizontal_gradient_of_exner_pressure_*` reads the cell field at index `-1`:

- **macOS**: SIGSEGV inside the gtfn program (this is why the MCH driver test could not be run locally).
- **Linux**: the read lands in mapped memory → **silently wrong numbers**.

Isolated experimentally: same grid file, same backend, same full domain, all inputs zeroed, only `keep_skip_values` flipped → `True` segfaults, `False` survives.

Every other consumer of the dycore already uses `keep_skip_values=False`: the datatest fixture (`fixtures/datatest.py`) and the stencil-test fixture (`fixtures/stencil_tests.py`). `_replace_skip_values` (`common/grid/base.py`) exists precisely to make the full-domain overcompute safe, and its docstring says so.

This is long-standing: the flag has been `True` since the driver was written for JW, where it was harmless (a global grid has no skip values in `E2C`). MCH is the first limited-area experiment the driver runs.

## Change

One line: `keep_skip_values=False`, with a comment explaining why.

The underlying defect — unguarded neighbour access in an over-computed domain — is already tracked in the dycore as a `TODO(havogt)` in `compute_edge_diagnostics_for_dycore_and_update_vn.py`. This PR does not touch it.

## Verification

MCH driver datatests run and pass. The seven other integration datatests are unchanged (the flag only matters for limited-area grids; global/torus grids have no `E2C` skip values, and pentagon skip values in `V2E`/`V2C` are still preserved because `_should_replace_skip_values` short-circuits on them).


model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -132,7 +132,12 @@ def create_grid_manager(
     )
     grid_manager(
         allocator=allocator,
-        keep_skip_values=True,
+        # The stencils compute over the full domain, including the outermost boundary
+        # lines where some neighbors do not exist, and they dereference the neighbor
+        # tables without guarding against skip values (GT4Py only guards reductions).
+        # On a limited area grid the skip values must therefore be replaced, as the
+        # granule tests do. See '_replace_skip_values' in common/grid/base.py.
```
jcanton:
```suggestion
```
when we get to reviewing: why was it set to True here?

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
```diff
@@ -29,10 +29,11 @@
...
 # meaningful rtol can cover the difference, only atol.
-# TODO(jcanton): the MCH_CH_R04B09 values are placeholders. The driver does not
-# reproduce this experiment yet: it misses the initial diffusion call, the
-# spin-up second-order divergence damping is wrong, and the lateral boundary and
-# slow-physics tendencies are zeroed. Measure once those are fixed.
+# MCH_CH_R04B09 is the only real data experiment, and the only one that does not
+# reproduce the reference to roundoff. Its remaining difference is likely in the
+# metrics and interpolation fields, which ICON4Py computes from the grid file while
+# ICON serializes its own: the dycore does not reproduce this experiment exactly
+# either. Under investigation.
```
jcanton:
```suggestion
```
unnecessary, we all know this

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
```diff
@@ -132,7 +132,9 @@ def create_grid_manager(
     )
     grid_manager(
         allocator=allocator,
-        keep_skip_values=True,
+        # the stencils dereference the neighbor tables over the full domain, unguarded:
+        # keeping the skip values reads out of bounds (see _replace_skip_values)
+        keep_skip_values=False,
```
jcanton, line 137: `    limited_area_or_distributed = config.limited_area or config.distributed`

jcanton: @msimberg ?

model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py
```diff
@@ -49,11 +47,11 @@
         "rho": (1.6e-15, 2.4e-15),
     },
     test_defs.Experiments.MCH_CH_R04B09: {
-        "vn": (6e-7, 0.0),
-        "w": (8e-9, 0.0),
-        "exner": (2e-10, 0.0),
-        "theta_v": (1e-7, 0.0),
-        "rho": (9e-10, 0.0),
+        "vn": (3.5e-3, 0.0),
```
jcanton, line 50: test with RBF savepoint and measure

model/standalone_driver/src/icon4py/model/standalone_driver/driver_utils.py
jcanton: investigate the location of the segfault and whether flipping is actually correct or needs more thinking

jcanton: is this only an optimization such that only the pentagon points are handled otherwise and only in the v2c / c2v connectivity?

jcanton:

PR #1375, opened 2026-07-14
Move DiagnosticStateNonHydro to common/states

jcanton:
from a friend:

Prerequisite of the MCH_CH_R04B09 stack, split out of #1371 so that its diff stays about the restart. Stacked on #1370.

## Why

`DiagnosticStateNonHydro` is not internal to the dycore: the driver allocates it, the initial condition fills part of it, and both `solve_nonhydro` and `velocity_advection` mutate it. It is an interface between the driver and the granule, which is what `common/states` is for — the prognostic, diagnostic and tracer states already live there.

Keeping it in the dycore forced #1371 to duplicate three of its field references in a container in `common`, only to dodge a `common -> dycore` import. This deletes the reason for that container.

## Change

`DiagnosticStateNonHydro` and `initialize_solve_nonhydro_diagnostic_state` move to `common/states/nonhydro_diagnostic_state.py`. Mechanical: no re-export from `dycore_states`, all 12 call sites updated (dycore, its tests, the old driver, the standalone driver).

`PrepAdvection` is in the same situation but is left alone here, to keep the move reviewable.

## Verification

Hooks (ruff, mypy, tach) clean. The JW and GAUSS3D driver datatests pass unchanged on gtfn_cpu.


OngChia (approved):
LGTM

jcanton:

jcanton:

PR #1379, opened 2026-07-16
fix least-squares namelist parameter lsq_dim_stencil

OngChia:
1. Fixed docstrings for `lsq_dim_unk`, `lsq_dim_c`, and `lsq_wgt_exp` to reference the correct Fortran source (`mo_interpol_config.f90`, not `mo_intp_coeffs_lsq_bln.f90`).
2. Renamed `lsq_dim_stencil` config field to `lsq_high_ord` to match the actual Fortran namelist parameter name (`lsq_high_ord`). This was a bug — the factory was reading interpol_nml["lsq_high_ord"] but storing it in the wrong field. The field is now documented as reserved for higher-order reconstruction.
3. Removed `lsq_dim_stencil` from the kwargs passed into `compute_lsq_coeffs`, since the stencil size is now correctly derived from `lsq_dim_c`.

The test in Advection package is not up to date. We need to update it in another PR.


OngChia:
cscs-ci run default;SESSIONS=model;MODEL_SUBPACKAGES=common

model/common/src/icon4py/model/common/math/projection.py
```diff
@@ -45,30 +47,29 @@ def gnomonic_proj(
...
+    x_diff = array_ns.where(
+        domain_length - array_ns.abs(x_diff) > array_ns.abs(x_diff),
+        x_diff,
+        x_diff - array_ns.sign(x_diff) * domain_length,
+    )
+    y_diff = array_ns.where(
+        domain_height - array_ns.abs(y_diff) > array_ns.abs(y_diff),
+        y_diff,
+        y_diff - array_ns.sign(y_diff) * domain_height,
+    )
```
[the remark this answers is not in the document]
jcanton: I read only very quickly, but may be related to / worth syncing with
https://github.com/C2SM/icon4py/pull/1354
which also has stuff for torus periodicity (and should be reviewed @OngChia  ;-) )

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1249,19 +1252,19 @@ def compute_z_lsq_mat_c(
...
+    lsq_matrix[valid_cell_mask_with_halo, :lsq_dim_c, :lsq_dim_unk] = (
         lsq_weights_c[valid_cell_mask_with_halo, :lsq_dim_c, array_ns.newaxis]
-        * z_dist_g[valid_cell_mask_with_halo, :lsq_dim_c, :]
+        * neighbor_center_dist[valid_cell_mask_with_halo, :lsq_dim_c, :]
     )
-    return z_lsq_mat_c
+    return lsq_matrix
 
 
 def compute_lsq_coeffs(
```
msimberg, line 1270: Not really the concern of this PR, but this seems like a good place to ask: why is compute_lsq_coeffs only used in tests? Has it simply not been integrated into the standalone driver yet?

model/atmosphere/advection/tests/advection/integration_tests/test_advection.py
```diff
@@ -154,10 +135,9 @@ def test_advection_run_single_step(  # noqa: PLR0917 [too-many-positional-argume
         domain_length=geometry.grid.grid_params.domain_length,
         domain_height=geometry.grid.grid_params.domain_height,
         grid_sphere_radius=constants.EARTH_RADIUS,
-        lsq_dim_unk=2,
-        lsq_dim_c=3,
-        lsq_wgt_exp=2,
-        lsq_dim_stencil=3,
+        lsq_dim_unk=experiment.config.interpolation.lsq_dim_unk,
+        lsq_dim_c=experiment.config.interpolation.lsq_dim_c,
+        lsq_wgt_exp=experiment.config.interpolation.lsq_wgt_exp,
```
msimberg, line 140: I'd like to hear @nfarabullini's input on these changes.

Important question: will the implementation actually handle these being set to anything other than the previously hardcoded values? If not, we should check and raise an exception saying other values are currently unsupported/untested? If yes, it'd be good to have some tests for other possible values.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1224,22 +1228,21 @@ def compute_lsq_pseudoinv(
...
     return lsq_weights_c
 
 
-def compute_z_lsq_mat_c(
+def compute_lsq_matrix(
     *,
     cell_owner_mask: data_alloc.NDArray,
     lsq_weights_c: data_alloc.NDArray,
-    z_dist_g: data_alloc.NDArray,
+    neighbor_center_dist: data_alloc.NDArray,
```
msimberg: No huge objection to more readable names, but just want to confirm: is this rename necessar (I see many other names not changed, meaning this will likely now (or already did?) have a mix of fortran names and "better" names, which might be more confusing than having all of them named the same as in fortran.

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
```diff
@@ -441,3 +442,54 @@ def test_compute_pos_on_tplane_e(
             )
     assert test_helpers.dallclose(pos_on_tplane_e_x, pos_on_tplane_e_x_ref, atol=1e-8, rtol=1e-9)
     assert test_helpers.dallclose(pos_on_tplane_e_y, pos_on_tplane_e_y_ref, atol=1e-8, rtol=1e-9)
+
+
+@pytest.mark.level("unit")
+@pytest.mark.datatest
```
msimberg, line 448: Just checking: did this move only because it's a unit test and it was in the wrong file? (keep the change if yes)

msimberg (commented):
@OngChia do you think you could separate the vectorization changes into a separate PR from the correctness changes in this PR? It'd make reviewing a lot easier.

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
```diff
@@ -441,3 +442,54 @@ def test_compute_pos_on_tplane_e(
...
+    domain_length = experiment.grid.params.domain_length
+    domain_height = experiment.grid.params.domain_height
+    start_idx = icon_grid.start_index(cell_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_2))
+    min_rlcell_int = icon_grid.end_index(cell_domain(h_grid.Zone.LOCAL))
+
+    lsq_pseudoinv = compute_lsq_coeffs(
+        cell_center_x=cell_center_x,
+        cell_center_y=cell_center_y,
+        cell_lat=cell_lat,
+        cell_lon=cell_lon,
```
nfarabullini:
```suggestion
        cell_center_x=grid_savepoint.cell_center_cart_x().ndarray,
        cell_center_y=grid_savepoint.cell_center_cart_y().ndarray,
        cell_lat=grid_savepoint.cell_center_lat().ndarray,
        cell_lon=grid_savepoint.cell_center_lon().ndarray,
```

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
```diff
@@ -441,3 +442,54 @@ def test_compute_pos_on_tplane_e(
...
+    grid_savepoint: sb.IconGridSavepoint,
+    icon_grid: base_grid.Grid,
+    backend: gtx_typing.Backend,
+) -> None:
+    c2e2c = icon_grid.get_connectivity(dims.C2E2C).ndarray
+    cell_owner_mask = grid_savepoint.c_owner_mask().ndarray
+    cell_center_x = grid_savepoint.cell_center_cart_x().ndarray
+    cell_center_y = grid_savepoint.cell_center_cart_y().ndarray
+    cell_lat = grid_savepoint.cell_center_lat().ndarray
+    cell_lon = grid_savepoint.cell_center_lon().ndarray
```
nfarabullini:
```suggestion
```

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
```diff
@@ -441,3 +442,54 @@ def test_compute_pos_on_tplane_e(
...
+
+    lsq_pseudoinv = compute_lsq_coeffs(
+        cell_center_x=cell_center_x,
+        cell_center_y=cell_center_y,
+        cell_lat=cell_lat,
+        cell_lon=cell_lon,
+        c2e2c=c2e2c,
+        cell_owner_mask=cell_owner_mask,
+        domain_length=domain_length,
+        domain_height=domain_height,
```
nfarabullini:
```suggestion
        domain_length=experiment.grid.params.domain_length,
        domain_height=experiment.grid.params.domain_height,
```

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
```diff
@@ -441,3 +442,54 @@ def test_compute_pos_on_tplane_e(
...
+    backend: gtx_typing.Backend,
+) -> None:
+    c2e2c = icon_grid.get_connectivity(dims.C2E2C).ndarray
+    cell_owner_mask = grid_savepoint.c_owner_mask().ndarray
+    cell_center_x = grid_savepoint.cell_center_cart_x().ndarray
+    cell_center_y = grid_savepoint.cell_center_cart_y().ndarray
+    cell_lat = grid_savepoint.cell_center_lat().ndarray
+    cell_lon = grid_savepoint.cell_center_lon().ndarray
+    domain_length = experiment.grid.params.domain_length
+    domain_height = experiment.grid.params.domain_height
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: `compute_lsq_coeffs` is actually used in the interpolation factory to compute `lsq_pseudoinv` (https://github.com/C2SM/icon4py/blob/147c4fc340a9c0c30a065646a420d4e91ae3b246/model/common/src/icon4py/model/common/interpolation/interpolation_factory.py#L318, pseudo inverse of the weighted distance matrix), which is used in the horizontal advection of tracers (https://github.com/C2SM/icon4py/blob/147c4fc340a9c0c30a065646a420d4e91ae3b246/model/atmosphere/advection/src/icon4py/model/atmosphere/advection/advection_horizontal.py#L297).

model/atmosphere/advection/tests/advection/integration_tests/test_advection.py
nfarabullini: I'm happy with these changes. I've just made a small suggestion, but it's really minor. 

As far as I remember in the f90 code, these values are indeed hardcoded and do not depend on any grid. I hope that answers your question

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Ah, but you are right. There is currently no physics in the driver, thus no tracers, that is why it is only used in the tests.

msimberg: I don't know why my previous search showed it only in tests :( thanks for checking and correcting me! In any case, expanding its usage is not for this PR, so no action needed for this comment.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: I renamed some variables when I came across names that I thought could be improved for better readability while working on a PR. I did not change some of the Fortran-derived names in `compute_lsq_coeffs`, such as `lsq_dim_unk` and `lsq_weight_c`, because I think they are already reasonably clear.
However, I wonder whether I am being too casual about renaming variables. What would be a better approach?

model/common/tests/common/interpolation/unit_tests/test_interpolation_fields.py
OngChia: Yes.

OngChia:
> @OngChia do you think you could separate the vectorization changes into a separate PR from the correctness changes in this PR? It'd make reviewing a lot easier.

Yes. I agree as it is related to https://github.com/C2SM/icon4py/pull/1354. I will separate it. It would be nice to do the vectorization after that PR is merged.

OngChia:
cscs-ci run default;SESSIONS=model;MODEL_SUBPACKAGES=common

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
msimberg: We don't really have a policy on that. We have long running "discussions" about eventually renaming things and using cf convention names as much as possible, but that goes against having nice short names in bodies for expressions. If you're not 100% certain that this new name is what we want to go for, I'd probably prefer leaving the old names, just because it makes it easier to change them all consistently in one go later. I won't block on renaming them back though.

msimberg (commented):
Almost-approval from me. Thanks @OngChia. The main question that's keeping me from an approval is the question of testing with the new configuration options: https://github.com/C2SM/icon4py/pull/1379#discussion_r3613161258.

model/atmosphere/advection/tests/advection/integration_tests/test_advection.py
OngChia: Thanks for the suggestion. This is a tricky question. In short, yes/no. 
I try to break down a few things.
The key variables are `lsq_dim_c` (rank of the matrix, or the maximum number of neighboring cells to be included for the interpolation) and `lsq_dim_unk` (the order of interpolation polynomial), and `lsq_dim_stencil` (deleted in this PR for a reason I will explain below, actual stencil size). 

**The current version of `compute_lsq_coeffs` only works for linear least-squares reconstruction** using the hardcoded parameters. It does not work for higher-order reconstruction because the computations for higher-order terms are missing (moreover, higher-moment computations for mass-conservation least-square reconstruction are also missing). **We should raise an exception when lsq_dim_unk > 2.** 

IF the interpolation stencil remains the same throughout all cells, there is no need to pass in `lsq_dim_stencil`. One particular exception is that if we use `c2v2c` as the reconstruction stencil. On regular triangles, it has 12 neighbors (thus, `lsq_dim_c = 12` in this case). **However, on pentagon points, the actual stencil size will be 11. We need to exclude looping over the extra column in the computations. This is not considered in the current version. We may raise an exception if the neighboring table contains an invalid value.**

Extra note: we need to change the input argument `c2e2c` to a more general name, ex. `lsq_neighboring_table`,

OngChia: I deleted lsq_dim_stencil because I did not find any of our experiments uses c2v2c to construct least-squares.

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
OngChia: Renaming has been a difficult subject. I love the not-too-long name for intermediate variables such as this one `neighbor_cell_dist`. But this is only my preference. I agree to revert them back, before we have a clear strategy and we let AI do the remaining labour job.

model/atmosphere/advection/tests/advection/integration_tests/test_advection.py
OngChia: I said **yes** at the beginning, because if the interpolation stencil remains the same throughout all cells and lsq_dim_unk = 1 / 2, let's say we use stencil c2e2c2e2c or c2v2c, the current version should correctly compute the pseudo inverse. Although it is untested. I am not sure whether we want to test it analytically now (or again serialize ICON data).

OngChia: So, in conclusion, I will raise an exception when `lsq_dim_unk != 2` or `lsq_dim_c != 3` in this PR.

msimberg: Thanks @OngChia, that sounds very good.

msimberg (approved):

nfarabullini (approved):

PR #1380, opened 2026-07-17
Clean up process_props/exchange/reductions API

msimberg:
This PR cleans up icon4py APIs that currently accept an `exchange` object and/or a `global_reductions` object alongside `process_props`. Wherever it makes sense, these constructors now take only a required `process_props` and derive exchange and reductions internally. `process_props` already contains everything needed to build both halo-exchange and global-reduction facilities, so passing derived objects through multiple layers was redundant.

The production APIs that changed are `GridGeometry`, `InterpolationFieldsFactory`, `MetricsFieldsFactory`, `Icon4pyDriver`, and `create_static_field_factories`. Each now takes a required `process_props` argument and uses `create_exchange` / `create_reduction` internally.

The module-level `single_node_exchange` and `single_node_reductions` singletons, and their `create_single_node_*` factory registrations, are removed. The `create_single_node_exchange` and `create_single_reduction_exchange` singledispatch registrations remain in `definitions.py`.

GHEX exchange and global-reduction instances are cached in `mpi_decomposition.py`, keyed by `(id(process_props), id(decomposition_info))`. Entries are held with strong references: production has one `(process_props, decomposition_info)` pair per run, and an autouse test fixture calls `clear_caches()` between tests.

Test call sites and helpers are updated to construct `SingleNodeProcessProperties()` or pass `process_props` directly where needed. Leaf consumers that already receive a derived `exchange` (including `Diffusion`, `SolveNonhydro`, advection constructors, states/provider plumbing, and `initialize_granules`) are intentionally unchanged because they use the exchange directly rather than constructing one.


msimberg:
cscs-ci run default

model/testing/src/icon4py/model/testing/grid_utils.py
```diff
@@ -30,12 +30,14 @@ def get_grid_manager_from_experiment(
     experiment: definitions.Experiment,
     keep_skip_values: bool,
     allocator: gtx_typing.Allocator,
+    process_props: decomposition.ProcessProperties | None = None,
```
msimberg, line 38: It's not clear to me why this should be allowed to be nullable. In general we are explicit. Is this particular use case special in any way that would warrant allowing passing `None` and defaulting to single node process props instead of being explicit?

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -583,42 +584,108 @@ def get_single_node_properties(s: SingleNodeRun, comm_id: int | None = None) ->
...
-    process_props: SingleNodeProcessProperties, decomp_info: DecompositionInfo
-) -> ExchangeRuntime:
-    return SingleNodeExchange()
+create_exchange.register = _create_exchange_impl.register  # type: ignore[attr-defined]
 
 
 @functools.singledispatch
+def _create_reduction_impl(
+    process_props: ProcessProperties,
+    decomposition_info: DecompositionInfo | None = None,
```
msimberg: In which cases can/should DecompositionInfo be None? Why is that allowed or necessary? Same below in create_reduction.

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -583,42 +584,108 @@ def get_single_node_properties(s: SingleNodeRun, comm_id: int | None = None) ->
...
+def clear_exchange_cache() -> None:
+    _exchange_cache.clear()
+
+
+def clear_reductions_cache() -> None:
+    _reduction_cache.clear()
+
+
 @functools.singledispatch
+def _create_exchange_impl(
```
msimberg: General question about the changes in this file: why does register/dispatch/etc. setup need changing? What purpose does it serve and why was the old setup not sufficient?

msimberg: The previous setup had separate singledispatch registrations for `create_single_node_exchange` and `create_single_reduction_exchange`; those were removed and the single-node fallback moved into `_create_exchange_impl`/`_create_reduction_impl`. `create_exchange`/`create_reduction` are now caching wrappers that forward `.register` to the underlying implementation so MPI-specific registrations still work.

🤖 *Posted by an agent on behalf of @msimberg*

model/common/src/icon4py/model/common/decomposition/definitions.py
msimberg: `DecompositionInfo` is `None` only on the single-node fallback path in `_create_exchange_impl`/`_create_reduction_impl`. MPI implementations with `comm_size > 1` require a real `DecompositionInfo` and raise if one is missing.

🤖 *Posted by an agent on behalf of @msimberg*

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -583,42 +584,108 @@ def get_single_node_properties(s: SingleNodeRun, comm_id: int | None = None) ->
...
+        ):
+            del _exchange_cache[key]
+        else:
+            return entry.exchange
+    exchange = _create_exchange_impl(process_props, decomp_info)
+    _exchange_cache[key] = _ExchangeCacheEntry(
+        exchange=exchange,
+        process_props_ref=weakref.ref(process_props),
+        decomp_info_ref=weakref.ref(decomp_info) if decomp_info is not None else None,
+    )
```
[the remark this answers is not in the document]
msimberg: The cache intentionally stores strong references; the weak-reference check only detects stale keys when the same `(process_props, decomp_info)` pair is reused. Production runs create one `ProcessProperties`/`DecompositionInfo` pair for the whole run, and the test autouse fixture clears these caches between tests. Failing fast for missing `decomp_info` in distributed runs is a sensible follow-up but out of scope here.

🤖 *Posted by an agent on behalf of @msimberg*

model/testing/src/icon4py/model/testing/grid_utils.py
msimberg: `process_props` is nullable only as a testing-layer convenience for single-node tests that do not need decomposition information. `GridManager.__call__` still requires `process_props` in production. We recorded this as discussion item R21.

🤖 *Posted by an agent on behalf of @msimberg*

model/common/src/icon4py/model/common/decomposition/definitions.py
```diff
@@ -583,42 +584,108 @@ def get_single_node_properties(s: SingleNodeRun, comm_id: int | None = None) ->
...
+        ):
+            del _reduction_cache[key]
+        else:
+            return entry.reductions
+    reductions = _create_reduction_impl(process_props, decomposition_info)
+    _reduction_cache[key] = _ReductionCacheEntry(
+        reductions=reductions,
+        process_props_ref=weakref.ref(process_props),
+        decomp_info_ref=weakref.ref(decomposition_info) if decomposition_info is not None else None,
+    )
```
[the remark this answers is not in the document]
msimberg: The reductions cache uses strong references for the same reason as the exchange cache, and stale entries are cleaned the same way. The single-pair-per-run lifetime and the test fixture keep this manageable. Adding a distributed-run fail-fast check is a separate follow-up.

🤖 *Posted by an agent on behalf of @msimberg*

msimberg:
- The branch is now up to date with `main` and all checks are green.
- The merge regression R23 (reintroduced `decomposition.single_node_exchange` in the relocated `test_compute_lsq_coeffs`) was fixed in commit `ba162f3b8a`.
- The two Copilot design suggestions about exchange/reductions cache resource management and fail-fast for missing `decomp_info` in distributed runs remain open for maintainer decision.

🤖 *Posted by an agent on behalf of @msimberg*

msimberg:
cscs-ci run default

msimberg:
cscs-ci run default

🤖 *Posted by an agent on behalf of @msimberg*

msimberg:
cscs-ci run default

model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -54,6 +58,20 @@
 log = logging.getLogger(__name__)
 
 
+# Exchange and reduction objects are expensive to build (GHEX context, halo
+# patterns), so they are cached per (process_props, decomposition_info) pair.
+# Keys are object identities; pairs are expected to be few and long-lived.
+# The pytest plugin clears these between tests via clear_caches().
```
nfarabullini:
```suggestion
```
I would remove these two lines

nfarabullini (approved):
LGTM. I literally just have the one small suggestion

PR #1381, opened 2026-07-17
Update merge pipeline to only produce success status on PRs

msimberg:
This is required to have merge queues work smoothly. Updates the `merge` pipeline to automatically trigger a dummy job when run on PRs so that a green checkmark is produced on the PR and the PR can be added to the merge queue. Github's required checks are the same for adding a PR to the merge queue and for merging from the merge queue to main, and we don't want to have to first run the full `merge` pipeline on the PR and then repeat it in the merge queue.


msimberg:
I'll go ahead and add this to the merge queue right away. Comments always welcome afterwards, but not waiting for reviews to get this working sooner rather than later.

PR #1391, opened 2026-07-23
Minimal config system

DropD:
This is a minimal way to read ICON4Py experiment configurations from a number of file formats (currently set up for yaml). 

## Current feature set:

- read  and write yaml files
- read and write ExperimentConfig from file
- read and write any of the component configs from file separately
- refer to some enum values by name (check DiffusionConfig for examples) in the config
- write configs (roundtrip not tested yet)
- unserialize union types with non-invasive tagging
- output yaml consistent with the repos's  pre-commit yaml formatting rules
- round-trippable both ways

## design & implementation considerations for this minimal increment (may change with feature set choices)

- no changes to existing datastructures
- serialization / deserialization code is added for some data types but kept separate
- no extra layer of data structures; file contents -> config dataclass is the full process
- validation left up to the dataclasses' __post_init__ as before

## User experience considerations (all fixable / changeable by extending this implementation)

- no cascading of option values (ndyn_substeps has to be configured for each component separately)
- `DriverConfig.start_of_timestepping` does not automatically default to `DriverConfig.start_of_simulation` - it has to be provided
- union types have to be written as 
 - "{type: \<type-key\>, value: \<value\>}" in case of single-value types (e.g. datetime types)
    ```yaml
    end_of_simulation:
      type: numstep
      value: 100
    ```
 - "{type: \<type-key\>, \<attr\>: \<value\>, \<attr\>: \<value\>...}" in case of types with attributes (e.g. topography configs)
   ```yaml
   topography:
     config:
       type: jablonowsky_williamson
       eta_0: 0.253
       u0: 35.79
   ```
- to use all default values for a mandatory component config the user is required to set it to "{}"
   ```yaml
   metrics: {}
   ```

## dependencies

- no completely new dependencies
- `cattrs` promoted from transitive (via xarray) to direct dependency of 'common'
- `pyyaml` promoted from transitive (via multiple paths) to direct dependency of 'common'


model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,95 @@
...
+
+CONV = cattrs.Converter(forbid_extra_keys=True)
+
+
+def read[T](yaml_str: str, config_cls: type[T]) -> T:
+    return CONV.structure(yaml.safe_load(yaml_str), config_cls)
+
+
+def structure_enum(val: str, enum_type: type[enum.Enum]) -> enum.Enum:
+    return enum_type.__members__[val.upper()]
```
msimberg, line 39: Would this benefit from a custom try/except for a better error message? Or am I missing it being handled somewhere else?

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,95 @@
...
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+from __future__ import annotations
+
+import enum
+import typing
+
+import cattrs
```
msimberg: TIL about cattrs! It looks very convenient for this purpose and seems to take care of a large part of the work needed here.

model/common/src/icon4py/model/common/topography/config.py
```diff
@@ -31,6 +32,51 @@
...
+        case "gaussian_hill":
+            topoclass = gausshill_topo.GaussianHillConfig
+        case "from_file":
+            topoclass = from_file_topo.FromFileConfig
+        case _:
+            raise TypeError(f"Unsupported topography type: '{topotype}'.")
+
+    return confreader.CONV.structure(config_dict, topoclass)  # type: ignore[return-value]
+
+
```
msimberg: Not a must: is it useful/possible to abstract this one more level by defining the `[("flat", flat_topo.FlatTopographyConfig), ...]` mapping separately and then generically take the mapping and generate the hooks? May hurt readability, but this may not be the last place where a custom mapping is needed. It may also be premature, I don't know if all future structure/unstructure hooks will look the same as this.

model/standalone_driver/tests/standalone_driver/unit_tests/test_config.py
```diff
@@ -141,3 +143,35 @@ def test_restart_starts_the_time_loop_at_start_of_timestepping() -> None:
...
+                start_of_simulation: 2020-01-01T00:00:00
+                start_of_timestepping: 2020-01-01T00:00:00
+                end_of_simulation:
+                    type: numstep
+                    value: 5
+            """
+        ),
+        driver_config.ExperimentConfig,
+    )
+    assert conf.driver.experiment_name == "foo"
```
msimberg: Just thinking of more tests: I think it should be possible to roundtrip a config between yaml and the dataclasses and get the same result. A partially specified config (i.e. with defaults) should be a subset of the full config after roundtripping yaml -> config -> yaml.

I think dumping the full config in some directory is probably also useful functionality for recording exactly what settings a simulation used (our defaults may change over time, so dumping the explicit exact config is useful as an artifact). This is of course not necessary for this PR, but something to keep in mind.

msimberg (commented):
I think this looks like a great direction.

On the ndyn_substeps constraint of not being able to refer to other configs, I think that's an ok simplification (as already discussed in the past). Just to understand the situation a bit better:
- do you have a feeling for how complex of a feature it would be to allow referring to other configs?
- (possibly a question for @yiluchen1066 @OngChia @jcanton) should things like ndyn_substeps ever be different across different components that might specify it? If it should never be different, we shouldn't even specify it in multiple places, right? Then we should rather have ndyn_substeps sit in a separate or top level config (though I don't know how much this complicates e.g. initializing components, they'd need to take another config as well then)?

OngChia:
> I think this looks like a great direction.
> 
> On the ndyn_substeps constraint of not being able to refer to other configs, I think that's an ok simplification (as already discussed in the past). Just to understand the situation a bit better:
> 
> * do you have a feeling for how complex of a feature it would be to allow referring to other configs?
> * (possibly a question for @yiluchen1066 @OngChia @jcanton) should things like ndyn_substeps ever be different across different components that might specify it? If it should never be different, we shouldn't even specify it in multiple places, right? Then we should rather have ndyn_substeps sit in a separate or top level config (though I don't know how much this complicates e.g. initializing components, they'd need to take another config as well then)?

`ndyn_substeps` is used in two granules: `solve_nonhydro` and `diffusion`. It is a runtime variable that is increased when the CFL number exceeds a certain limit.

For `solve_nonhydro`, `ndyn_substeps` is passed as an input argument from the driver because changing the number of substeps changes the time step of each substep and, consequently, the resulting tendencies.

For diffusion, `ndyn_substeps` is used to compute the diffusion coefficients. However, I think that it is effectively treated as a constant mainly because diffusion is called only once per full time step. Therefore, the diffusion tendency does not change with the number of dynamical substeps. I have not thought of whether we can remove its dependency on `ndyn_substeps` by modifying the computation of diffusion coefficient. I will take a look.

As far as I investigated, there is no way to remove `ndyn_substeps` from diffusion. We set `ndyn_substeps` in the driver, and diffusion takes the initial value of `ndyn_substeps` for initialization of some constants.

model/common/src/icon4py/model/common/config/config_io.py
DropD: In practice, this gets called by the `cattrs` dispatcher, which will wrap it in it's own try/except. Here's some sample output from trying to read a misspelled enum value:

```bash
cattrs.structure({"diffusion_type": "smagorinsky_4th_orde", "apply_to_vertical_wind": False}, df.DiffusionConfig)
  + Exception Group Traceback (most recent call last):
  |   File "/Users/ricoh/.cache/uv/archive-v0/pjp-piC7UUd48r91/lib/python3.12/site-packages/IPython/core/interactiveshell.py", line 3748, in run_code
  |     exec(code_obj, self.user_global_ns, self.user_ns)
  |   File "<ipython-input-8-5d9bb9fc0363>", line 1, in <module>
  |     cattrs.structure({"diffusion_type": "smagorinsky_4th_orde", "apply_to_vertical_wind": False}, df.DiffusionConfig)
  |   File "/Users/ricoh/Code/icon4py/.venv/lib/python3.12/site-packages/cattrs/converters.py", line 591, in structure
  |     return self._structure_func.dispatch(cl)(obj, cl)
  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |   File "<cattrs generated structure icon4py.model.atmosphere.diffusion.diffusion.DiffusionConfig>", line 166, in structure_DiffusionConfig
  |     if errors: raise __c_cve('While structuring ' + 'DiffusionConfig', errors, __cl)
  |                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | cattrs.errors.ClassValidationError: While structuring DiffusionConfig (1 sub-exception)
  +-+---------------- 1 ----------------
    | Traceback (most recent call last):
    |   File "<cattrs generated structure icon4py.model.atmosphere.diffusion.diffusion.DiffusionConfig>", line 6, in structure_DiffusionConfig
    |     res['diffusion_type'] = __c_structure_diffusion_type(o['diffusion_type'])
    |                              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/Users/ricoh/.pyenv/versions/3.12.4/lib/python3.12/enum.py", line 757, in __call__
    |     return cls.__new__(cls, value)
    |            ^^^^^^^^^^^^^^^^^^^^^^^
    |   File "/Users/ricoh/.pyenv/versions/3.12.4/lib/python3.12/enum.py", line 1171, in __new__
    |     raise ve_exc
    | ValueError: 'smagorinsky_4th_orde' is not a valid DiffusionType
    | Structuring class DiffusionConfig @ attribute diffusion_type
    +------------------------------------
```

It contains all the right info. It's true, I would not consider it polished enough to be *nice*. Should do for now though.

model/common/src/icon4py/model/common/topography/config.py
DropD: I have experimented with that, since there are already two places where this pattern occurs. I find it no less readable.

Another thought experiment was to instead do an automatic mapping (`snake_case` <-> `CamelCase` and drop/add suffixes like `Config`), but then renaming classes becomes conflated with changing config syntax and vice versa...

model/standalone_driver/tests/standalone_driver/unit_tests/test_config.py
DropD: I want to have roundtrip tests both ways before merging

DropD:
> > I think this looks like a great direction.
> > On the ndyn_substeps constraint of not being able to refer to other configs, I think that's an ok simplification (as already discussed in the past). Just to understand the situation a bit better:
> > 
> > * do you have a feeling for how complex of a feature it would be to allow referring to other configs?
> > * (possibly a question for @yiluchen1066 @OngChia @jcanton) should things like ndyn_substeps ever be different across different components that might specify it? If it should never be different, we shouldn't even specify it in multiple places, right? Then we should rather have ndyn_substeps sit in a separate or top level config (though I don't know how much this complicates e.g. initializing components, they'd need to take another config as well then)?
> 
> `ndyn_substeps` is used in two granules: `solve_nonhydro` and `diffusion`. It is a runtime variable that is increased when the CFL number exceeds a certain limit.
> 
> For `solve_nonhydro`, `ndyn_substeps` is passed as an input argument from the driver because changing the number of substeps changes the time step of each substep and, consequently, the resulting tendencies.
> 
> For diffusion, `ndyn_substeps` is used to compute the diffusion coefficients. However, I think that it is effectively treated as a constant mainly because diffusion is called only once per full time step. Therefore, the diffusion tendency does not change with the number of dynamical substeps. I have not thought of whether we can remove its dependency on `ndyn_substeps` by modifying the computation of diffusion coefficient. I will take a look.
> 
> As far as I investigated, there is no way to remove `ndyn_substeps` from diffusion. We set `ndyn_substeps` in the driver, and diffusion takes the initial value of `ndyn_substeps` for initialization of some constants.

I was under the impression that `ndyn_substeps` was also in the components in order to be able to run them separately, if that is not a concern, we could turn it into a parameter to `run`. Or if we don't want to pass it every time, a property which can be set by the driver...

model/common/src/icon4py/model/common/config/config_io.py
msimberg: Nice, agree. Good enough for now, but could definitely be polished later.

model/common/src/icon4py/model/common/topography/config.py
msimberg: Nice, if you think it's not too much extra effort for this PR, sounds like it'd be worth doing right away then!

DropD:
> I think this looks like a great direction.
> 
> On the ndyn_substeps constraint of not being able to refer to other configs, I think that's an ok simplification (as already discussed in the past). Just to understand the situation a bit better:
> 
> * do you have a feeling for how complex of a feature it would be to allow referring to other configs?

As far as I can tell there is no super-straightforward way, so it would contain some experimentation. The problem is that the config hierarchy is built leaves to root. Here are some general approaches (without having the user encode this with yaml references or such):

- intermediary layer of config classes: basically build a config-IR and then transform that to the actual config hierarchy.
- have placeholders as defaults on the leaves and let the root config find and replace them in `__post__init__` (will require some circumventing "frozenness" of config classes
- pre-process the dict we get from reading the yaml files
- `cattrs` might have a customization mechanism I don't know much about yet (like "strategies") that could be used.

Most of these are probably relatively simple for a given hierarchy with only few config options that need to be taken care of, with the downside of making it harder to change that hierarchy later. If we need a general solution, we probably want to reach for something more heavy-weight than `cattrs`.

OngChia:
> > > I think this looks like a great direction.
> > > On the ndyn_substeps constraint of not being able to refer to other configs, I think that's an ok simplification (as already discussed in the past). Just to understand the situation a bit better:
> > > 
> > > * do you have a feeling for how complex of a feature it would be to allow referring to other configs?
> > > * (possibly a question for @yiluchen1066 @OngChia @jcanton) should things like ndyn_substeps ever be different across different components that might specify it? If it should never be different, we shouldn't even specify it in multiple places, right? Then we should rather have ndyn_substeps sit in a separate or top level config (though I don't know how much this complicates e.g. initializing components, they'd need to take another config as well then)?
> > 
> > 
> > `ndyn_substeps` is used in two granules: `solve_nonhydro` and `diffusion`. It is a runtime variable that is increased when the CFL number exceeds a certain limit.
> > For `solve_nonhydro`, `ndyn_substeps` is passed as an input argument from the driver because changing the number of substeps changes the time step of each substep and, consequently, the resulting tendencies.
> > For diffusion, `ndyn_substeps` is used to compute the diffusion coefficients. However, I think that it is effectively treated as a constant mainly because diffusion is called only once per full time step. Therefore, the diffusion tendency does not change with the number of dynamical substeps. I have not thought of whether we can remove its dependency on `ndyn_substeps` by modifying the computation of diffusion coefficient. I will take a look.
> > As far as I investigated, there is no way to remove `ndyn_substeps` from diffusion. We set `ndyn_substeps` in the driver, and diffusion takes the initial value of `ndyn_substeps` for initialization of some constants.
> 
> I was under the impression that `ndyn_substeps` was also in the components in order to be able to run them separately, if that is not a concern, we could turn it into a parameter to `run`. Or if we don't want to pass it every time, a property which can be set by the driver...

Yes. Granules `SolveNonhydro` and `Diffusion` need `ndyn_substeps` to run. We already pass `ndyn_substeps_var` (this is a run-time variable in `Icon4pyDriver`, it is equal to `ndyn_substeps` when CFL number is low) to  `SolveNonhydro.time_step` in the time loop inside `Icon4pyDriver._do_dyn_substepping`. `Diffusion` only needs `ndyn_substeps` for initialization of some run-time constants. My suggestion is that we place `ndyn_substeps` in the top level config `DriverConfig` (or, perhaps in the future, we can put it in `TimeConfig` that controls the timing of calls of all components and dycore as well as start/end date). Then, we pass `ndyn_substeps` to `Diffusion` during initialization. 
```
diffusion_granule = diffusion.Diffusion(
            grid=grid,
            config=diffusion_config,
            ndyn_substeps=driver_config.ndyn_substeps,
            ...
)
```
But the problem is we need to modify the data tests in `Diffusion` and the wrapper. It looks too troublesome to me,I agree that configuring `ndyn_substeps` separately for `Icon4pyDriver` and `Diffusion` is a reasonable simplification.

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
+---
+tags: [config, io, yaml]
+---
+
+# [Configuration IO]
+
+- **Status**: valid
+- **Authors**: Rico Häuselmann (@DropD)
+- **Created**: 2026-07-29
+- **Updated**: 2026-07-29
```
nfarabullini: do you need both of these since they're the same date?

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+While implementing and discussing the first iteration of ICON4Py's own configuration files, the following design choices were made.
```
nfarabullini:
```suggestion
While implementing and discussing the first iteration of ICON4Py's own configuration files, some design choices were made.
```
If you say `following`, I expect a list right after

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Context
+
+The configuration file is the primary way for a normal user to interact with ICON4Py, hence this iteration has been preceded by experiments and discussions. This ADR is a record of the results.
```
nfarabullini: iteration?

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+The following features / trade-offs were discussed and investigated
```
nfarabullini:
```suggestion
The following features / trade-offs were discussed and investigated:
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+- compatibility with ICON
+- configuration file format choice
```
nfarabullini:
```suggestion
- choice of configuration file format 
```
I like this phrasing better, but it's just a personal preference

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+- usability of the configuration file format
+- potential lock-in into specific frameworks
+- round-tripping may or may not change a configuration file
```
nfarabullini:
```suggestion
- possibility of round-tripping changing a configuration file
```
To be in line with the phrasing of the other bullet points

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Decision
+
+### ICON compatibility
+
+ICON4Py is currently able to read ICON configurations (namelists) on a best-effort basis (not all options supported). This may be replaced by a tool to convert namelists to ICON4Py configuration files
```
nfarabullini, line 30:
```suggestion
ICON4Py is currently able to read ICON configurations (namelists) on a best-effort basis (not all options supported). This may be replaced by a tool to convert namelists to ICON4Py configuration files.
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Configuration file format
+
+Desired: human readable well-defined format.
+
+Current choice: YAML as default, other options supported by `cattrs`, the library we currently use.
+
+Change if:
+
+ - a demonstrably better format is identified
```
nfarabullini:
```suggestion
Change if a demonstrably better format is identified
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+We decided to remove these duplicated options from the components, as they were historically necessary but serve only convenience now. This is a work in progress.
+
+### Usability - Round Trips
+
+Ideally, round-tripping configuration files should not change them.
+
+Current state:
+
+- true for configuration files written by ICON4Py, not true in general
```
nfarabullini:
```suggestion
- true for configuration files written by ICON4Py, false otherwise
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+- the `pyyaml` library, which powers our YAML support does not support the general case
+
+Change if:
+
+- if this is considered worth the (small) effort, `pyyaml` could be replaced with `ruamel` or another library with round-tripping support
```
nfarabullini:
```suggestion
Change if this is considered worth the (small) effort, `pyyaml` could be replaced with `ruamel` or another library with round-tripping support.
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+### Avoiding lock-in
+
+If possible avoid frameworks that only work with data structures they themselves provide.
+
+If possible avoid frameworks that require custom serialization / de-serialization logic to be defined inside the data structures to be customized.
```
nfarabullini:
```suggestion
If possible, avoid frameworks that only work with data structures they themselves provide.

If possible, avoid frameworks that require custom serialization / de-serialization logic to be defined inside the data structures to be customized.
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+We decide to start out using `cattrs` because it
+
+- can work with the existing configuration data classes
+- serialization and de-serialization can be customized without touching the classes we customize for
+- was already a transitive dependency at the time
```
nfarabullini:
```suggestion
We decide to start out using `cattrs` for the following reasons:

- it can work with the existing configuration data classes
- serialization and de-serialization can be customized without touching the classes we customize for
- it was already a transitive dependency at the time
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Alternatives Considered
+
+### Using `omegaconf`
+
+An earlier iteration ([#936](https://github.com/C2SM/icon4py/pull/936)) used `omegaconf`.
+
+- required little change in existing configuration classes
+- avoided cleaning up the duplicated configuration options in component configuration classes by automatically synchronizing the component's values with the top-level (if given)
+- introduced additional indirection: a `ConfigManager` class was built from the configuration file, which then in turn built the configuration classes.
```
nfarabullini:
```suggestion
- introduced additional indirection: a `ConfigManager` class was built from the configuration file, which in turn built the configuration classes.
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+- introduced the need to wrap configuration classes in `ConfigHandler`s in some cases
+
+All in all it traded complexity for the feature that config classes can define options that refer to options in a config class higher up the hierarchy.
```
nfarabullini: This sentence is a bit convoluted. Some rephrasing would be nice

model/atmosphere/diffusion/tests/diffusion/unit_tests/data/test_config.yml
```diff
@@ -0,0 +1,3 @@
+shear_type: vertical_horizontal_of_horizontal_wind
```
nfarabullini: did you run `pre-commit`? Any icon4py file should start with 
```
# ICON4Py - ICON inspired code in Python and GT4Py
#
# Copyright (c) 2022-2024, ETH Zurich and MeteoSwiss
# All rights reserved.
#
```

model/atmosphere/diffusion/tests/diffusion/unit_tests/test_config.py
```diff
@@ -0,0 +1,30 @@
...
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+import pathlib
+
+from icon4py.model.atmosphere.diffusion import diffusion
+from icon4py.model.common.config import config_io
+
+
+def test_read() -> None:
```
nfarabullini:
```suggestion
def test_diffusion_config_read() -> None:
```
Or something analogous to be more precise in what the test does

model/atmosphere/diffusion/tests/diffusion/unit_tests/test_config.py
```diff
@@ -0,0 +1,30 @@
...
+    config = config_io.read(
+        (pathlib.Path(__file__).parent / "data" / "test_config.yml").read_text(),
+        diffusion.DiffusionConfig,
+    )
+    assert (
+        config.shear_type
+        is diffusion.TurbulenceShearForcingType.VERTICAL_HORIZONTAL_OF_HORIZONTAL_WIND
+    )
+    assert config.ndyn_substeps == 4
+    assert not config.apply_zdiffusion_t
```
nfarabullini: if these are set in the yml file, do you actually need a test?

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,119 @@
...
+
+
+CONV = cattrs.preconf.pyyaml.PyyamlConverter(forbid_extra_keys=True)
+
+
+CONV.register_unstructure_hook(ta.wpfloat, lambda v: CONV.unstructure(float(v)))
+yaml.add_representer(type(None), lambda d, _: d.represent_scalar("tag:yaml.org,2002:null", ""))
+
+
+def read[T](yaml_str: str, config_cls: type[T]) -> T:
```
nfarabullini: I would prefer more specific names for all of these

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,119 @@
...
+        case "numsteps":
+            timeclass = time.NumTimeSteps
+    if not timeclass:
+        raise TypeError(f"unsupported end of simulation time type: '{timetype}'")
+    return CONV.structure(endtime_dict["value"], timeclass)
+
+
+@CONV.register_unstructure_hook
+def unstructure_endtime(endtime: time.EndOfSimulation) -> dict:
+    timetype = ""
```
nfarabullini:
```suggestion
    timetype: str = ""
```
To be consistent with structure_endtime where timeclass has its data types specified

nfarabullini (changes requested):

model/atmosphere/diffusion/tests/diffusion/unit_tests/data/test_config.yml
DropD: Yes, I have; my understanding is that this rule applies to _Python_ modules only, at least for now.

model/atmosphere/diffusion/tests/diffusion/unit_tests/test_config.py
DropD: I thought it would be good to have a test (but only one) that would trigger if suddenly values were not read from file anymore. 

It would be unfortunate if we only notice that every configuration file leads to the same default values is after some poor PhD candidate has spent a year trying to figure out why they keep getting weird results...

model/common/src/icon4py/model/common/config/config_io.py
DropD: These are things I considered putting in the name but didn't:
- `config`: because it will always (per coding guidelines) be referred to as `config_io.read`, so it's redundant
- `yaml`: because we are only supporting one format for now, so it's redundant
- `from_string`: because I was planning to name the one reading directly from file `read_from_file` instead, if we ever want a convenience wrapper around `read(path.read_text())`

Did I forget something or do you disagree with any of the above?

model/atmosphere/diffusion/tests/diffusion/unit_tests/test_config.py
DropD: I am probably going to move that test to unit tests of the `config_io` module though.

docs/development/ADRs/0003_configuration_io.md
DropD: That is subtly wrong, because one can write a round-trippable config by hand. Maybe I will just drop that second clause, as the next bullet point makes it redundant.

docs/development/ADRs/0003_configuration_io.md
DropD: replaced with "the first implementation"

model/atmosphere/diffusion/tests/diffusion/unit_tests/test_config.py
nfarabullini: makes sense

model/common/src/icon4py/model/common/config/config_io.py
nfarabullini: what about `read_yaml_str` ? I think having the `yaml` helps

docs/development/ADRs/0003_configuration_io.md
jcanton:
```suggestion
The configuration file is the primary way for a user to interact with ICON4Py, hence this iteration has been preceded by experiments and discussions. This ADR is a record of the results.
```

docs/development/ADRs/0003_configuration_io.md
```diff
@@ -0,0 +1,95 @@
...
+
+Current choice: YAML as default, other options supported by `cattrs`, the library we currently use.
+
+Change if:
+
+ - a demonstrably better format is identified
+
+### Usability - Duplicated Options
+
+At the time of writing this ADR, some component configuration classes contain attributes which should be configured only once for the whole experiment. They also provide default values for those options. This makes the configuration format obviously error prone.
```
jcanton, line 44: add an example?

model/common/src/icon4py/model/common/config/config_io.py
DropD: I can agree to that.

docs/development/ADRs/0003_configuration_io.md
DropD: I simplified it.

model/common/src/icon4py/model/common/initial_condition/config.py
```diff
@@ -39,14 +40,25 @@
 log = logging.getLogger(__name__)
 
 
+type ICCONFIG = (
```
jcanton: opinion: I found this hard to read initially, can we have `IC_CONFIG`?

model/common/src/icon4py/model/common/topography/config.py
```diff
@@ -31,14 +32,27 @@
 log = logging.getLogger(__name__)
 
 
+type TOPOCONFIG = (
```
jcanton: if you change ICCONFIG change this one too for consistency

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
jcanton, line 1: is it possible to use 4 spaces indent? or yml default is 2?

jcanton (commented):
very small input from my side, others already did a nice review of a nice pr

I'll leave it to whoever blocked you with requested changes to unblock though ;-)

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
```diff
@@ -0,0 +1,91 @@
...
+  SLEVE_decay_exponent: 1.2
+  _SLEVE_minimum_layer_thickness_1: 100.0
+  _SLEVE_minimum_layer_thickness_2: 500.0
+  _SLEVE_minimum_relative_layer_thickness_1: 0.3333333333333333
+  _SLEVE_minimum_relative_layer_thickness_2: 0.5
+topography:
+  config:
+    type: jablonowski_williamson
+    u0: 35.0
+    eta_0: 0.252
```
OngChia, line 49: Just a reminder that these two parameters are in a similar situation as `ndyn_substeps`. They have to be equal to `u0` and `eta_0` in the initial_condition for the JW test to be valid (otherwise the wind field and thermodynamics itself do not make physical sense).

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,119 @@
...
+        case time.NumTimeSteps():
+            timetype = "numsteps"
+        case _:
+            raise TypeError(f"Unsupported time type: '{type(endtime)}'.")
+    return {"type": timetype, "value": CONV.unstructure(endtime)}
+
+
+def register_config_union[T](union_type: type[T], mapping: typing.Mapping[str, type[T]]) -> None:
+    def structure(config_dict: dict, _: typing.Any) -> T:
+        topotype = config_dict.pop("type")
```
OngChia: how about change the name to `config_type` to be more general?
```suggestion
        config_type = config_dict.pop("type")
```

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
DropD: I think this is determined by the configuration of the yaml formatter we use in `pre-commit`.

docs/development/ADRs/0003_configuration_io.md
DropD: This is straight from the template. I think it helps to be reminded that one should change `updated` when the line is there. Anyway, by the time this gets merged they won't be the same anymore.

docs/development/ADRs/0003_configuration_io.md
DropD: Added `ndyn_substep` as well as `u0`, `eta_0` mentioned above

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
DropD: Found it:

https://github.com/C2SM/icon4py/blob/f94d2d44e9790e2668f45277fd69bf2ee75fbca4/.pre-commit-config.yaml#L18-L19

To clarify, we can read files with any valid indentation automatically, we can customize what indentation we write (2 spaces is the default, yes), but I strongly suggest that it be changed in tandem with the pre-commit hook setting.

With the current implementation, there is no easy way to remember which indentation a config file uses, in order to avoid changing whitespace when roundtripping.

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
DropD: Thank you, I will take that into account. I also added them as examples to the ADR.

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,119 @@
...
+    return enum_type.__members__[val.upper()]
+
+
+def unstructure_enum(val: enum.Enum) -> str:
+    return val.name.lower()
+
+
+def register_enum[T](enum_type: type[T]) -> type[T]:
+    CONV.register_structure_hook(enum_type, structure_enum)
+    CONV.register_unstructure_hook(enum_type, unstructure_enum)
```
nfarabullini, line 48: Since `unstructure_enum` is only used here (as far as I can tell), we can directly apply the functionality without the function definition

model/common/src/icon4py/model/common/config/config_io.py
```diff
@@ -0,0 +1,119 @@
...
+
+@CONV.register_structure_hook
+def structure_abstime(abstime_val: str, _: typing.Any) -> time.AbsoluteTime:
+    if isinstance(abstime_val, time.AbsoluteTime):
+        return abstime_val
+    return time.AbsoluteTime.fromisoformat(abstime_val)
+
+
+@CONV.register_unstructure_hook
+def unstructure_abstime(abstime: time.AbsoluteTime) -> str:
```
nfarabullini, line 60: where is this used?

model/common/tests/common/config/unit_tests/test_config_io.py
```diff
@@ -0,0 +1,180 @@
...
+    assert abstime.year == 2026
+    assert abstime.second == 25
+    assert config_io.write_yaml_str(abstime) == abstime_str
+
+
+def test_roundtrip_reltime() -> None:
+    reltime_str = "300\n...\n"
+    reltime = config_io.read_yaml_str(reltime_str, time.RelativeTime)
+    assert reltime.seconds == 300
+    assert config_io.write_yaml_str(reltime) == reltime_str
```
nfarabullini: maybe parametrize these two?

model/common/tests/common/config/unit_tests/test_config_io.py
```diff
@@ -0,0 +1,180 @@
...
+    required_flag: bool
+    optional_int: int = 5
+
+
+@dataclasses.dataclass
+class AlternativeConfig:
+    unrelated_int: int = 4
+
+
+@config_io.register_enum
```
nfarabullini, line 32: I would put all of these classes in a separate file from the `test_*.py` ones, e.g. `common.py` within the test folder. Usually we keep the `test_*.py` strictly for tests

model/common/src/icon4py/model/common/config/config_io.py
DropD: One concrete example I know of is writing out the driver config. To get all the places, simply comment out the decorator and run `uv run pytest model/standalone_driver/tests/standalone_driver/unit_tests/test_config.py -k io -v`

model/common/tests/common/config/unit_tests/test_config_io.py
DropD: Do you mean:
a) fold them into one test method parametrized by `(time_type, reference, checking_function)` 
b) parametrize each with multiple examples

I thought about a) but didn't try it. I expected it wouldn't be worth paying the readability price for the reduction in lines.

nfarabullini: a) :) I think it's still worth it

model/common/tests/common/config/unit_tests/test_config_io.py
DropD: I am hesitant to agree with this, as none of these classes have any value whatsoever outside this particular test module right now. If that changes, by all means, let's do it!

When I checked around how this is handled in other test modules I came across these among the first five

- `bindings/tests/bindings/test_dycore_wrapper.py` (contains a huge fixture)
- `bindings/tests/bindings/test_icon4py_export.py` (various things)
- `model/common/tests/common/test_model_options.py` (a fieldop and a program)

I looked at more, but stopped when I lost count. I would argue these are not fundamentally different cases.

If you think this is wrong, we should update the coding guidelines, as they currently recommend separate modules only for "shared" fixtures and utilities. That would be a separate PR though.

>The `<component>` folder should be a Python package and contain subfolders for every kind of test (e.g. `unit_tests`, `integration_tests`, ...). If needed, it may contain at any level of the file tree a `conftests.py` module for changing the pytest configuration, a `fixtures.py` module with shared fixture definitions and a `utils.py`python modules testing utilities used in the tests.

nfarabullini: I see your argument. Perhaps some other reviewer can chip in with their thoughts @jcanton @OngChia @msimberg

model/common/src/icon4py/model/common/config/config_io.py
DropD: We can (same for `structure_enum`). Should we?

Assuming you mean the following:

```python
def register_enum[T](enum_type: type[T]) -> type[T]:
    CONV.register_structure_hook(enum_type, lambda val, enum_type: enum_type.__members__[val.upper()])
    CONV.register_unstructure_hook(enum_type, lambda val: val.name.lower())
```

While developing, I found the error messages a little harder to read when lambdas are involved.

If you find the above code so much more readable that it outweighs that minor downside, then we should.

nfarabullini: nah it's not a big deal

model/standalone_driver/tests/standalone_driver/unit_tests/data/test_config.yml
jcanton: ok, was just to have it same as the .py files :-)

model/common/tests/common/config/unit_tests/test_config_io.py
jcanton: I agree with @DropD and prefer them defined where used, since it's only one place for now; we can move them in the future if it turns out to be necessary

model/common/tests/common/config/unit_tests/test_config_io.py
DropD: I took the liberty of folding all the tests for custom (de)structuring into one parametrized test, that way it should also be obvious how to add one for future customizations.

nfarabullini (approved):

PR #1395, opened 2026-07-27
replace imports for testing.definitions …

### LINKED ISSUE #1238: improve testing.definitions import
replace all imports of `from icon4py.model.testing import definitions` to use  an alias as `from icon4py.model.testing import definitions as test_defs` and update all related usages

nfarabullini:
replace all imports of from icon4py.model.testing import definitions to from icon4py.model.testing import definitions as test_defs

FIXES #1238

### DIFF
--- a/model/common/tests/common/fixtures.py
+++ b/model/common/tests/common/fixtures.py
@@ -28,7 +28,6 @@ from icon4py.model.testing.fixtures.datatest import (
     data_provider,
     decomposition,
     decomposition_info,
-    definitions,
     download_ser_data,
     experiment,
     experiment_description,
@@ -38,6 +37,7 @@ from icon4py.model.testing.fixtures.datatest import (
     linit,
     metrics_savepoint,
     process_props,
+    test_defs,
     topography_savepoint,
 )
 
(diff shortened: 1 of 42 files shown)


model/common/tests/common/fixtures.py
```diff
@@ -38,6 +37,7 @@
     linit,
     metrics_savepoint,
     process_props,
+    test_defs,
```
OngChia, line 40: `test_defs` is imported from `icon4py.model.testing.fixtures.datatest`. Why not just do
`from icon4py.model.testing import definitions as test_defs`

model/common/tests/common/fixtures.py
```diff
@@ -28,7 +28,6 @@
     data_provider,
     decomposition,
```
OngChia, line 29: `decomposition` is imported from `icon4py.model.testing.fixtures.datatest`. Why not just do `from icon4py.model.common.decomposition import definitions as decomp_defs`?

OngChia (commented):
I left two comments about the import statement in `fixtures.py`. Could you have a look? Thanks

model/common/tests/common/fixtures.py
nfarabullini: because this import is inherited from `dataset.py` where it is imported as `from icon4py.model.testing import definitions as test_defs`

model/common/tests/common/fixtures.py
nfarabullini: sure, that makes sense

nfarabullini: Though then I will have to change it in many other places and this PR was only to tackle the `test_defs` import. Perhaps we can do this in a separate PR

OngChia (approved):

nfarabullini:
cscs-ci run default

PR #1398, opened 2026-07-27
Change some CSCS CI variable choices to use all instead of explicit lists

msimberg:
By default, all subpackages are meant to be tested, so replacing the explicit lists with `all` in both `default` and `merge`. Same with `SESSIONS`. For `merge` I'm updating the list of backends to use `all` as well. `default` keeps the single backend.

This avoids the need to update the lists whenever a new package is added, removed, or renamed (for CSCS CI).


msimberg:
cscs-ci run default

edopao (approved):
LGTM

nfarabullini (approved):

jcanton (approved):
thanks

OngChia (approved):

PR #1404, opened 2026-07-28
enable tracer advection in standalone driver and fix tracers from prognostic

### LINKED ISSUE #1402: TracerState should not inside the PrognosticState, since it cannot be swapped after every sub-timestep, it only swaps after 1 timestep
TracerState should not inside the PrognosticState, since it cannot be swapped after every sub-timestep, it only swaps after 1 timestep

yiluchen1066:


model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
```diff
@@ -216,6 +216,36 @@ def show_timer_report(
...
+    if not tracer_advection_enabled:
+        return None
+    if prep_adv is not None:
+        return tracer_advection_states.AdvectionPrepAdvState(
+            vn_traj=prep_adv.vn_traj,
+            mass_flx_me=prep_adv.mass_flx_me,
+            mass_flx_ic=prep_adv.dynamical_vertical_mass_flux_at_cells_on_half_levels,
+        )
+    return tracer_advection_states.AdvectionPrepAdvState(
+        vn_traj=data_alloc.zero_field(grid, dims.EdgeDim, dims.KDim, allocator=allocator),
```
OngChia, line 246: I think not for this PR. `AdvectionPrepAdvState` is a duplication of `PrepAdvection` in `dycore_state.py`. This is another example that suggest we need something else to store those variables such as a global set of variables. Since historically we have been building granule by granule, this is a legacy of that strategy.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -122,6 +124,32 @@ def _diagnostics_computer(self) -> driver_io.DiagnosticsComputer:
...
+        """Airmass program (``rho * ddqz_z_full``) with its static inputs bound.
+
+        The deep-atmosphere metric factor is identically 1 (shallow atmosphere).
+        """
+        return model_options.setup_program(
+            program=compute_airmass.compute_airmass,
+            backend=self.backend,
+            constant_args={
+                "ddqz_z_full_in": self.static_field_factories.metrics.get(metrics_attr.DDQZ_Z_FULL),
+                "deepatmo_t1mc_in": data_alloc.constant_field(
```
OngChia, line 141: We can remove this since we are not going to simulate deep atmosphere (the assumption that all metrics can be considered flat Euclidean locally, in other words we do not need to consider the curvature of the Earth) in the near future and we have excluded the deep atmosphere computations so far

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
```diff
@@ -122,6 +124,32 @@ def _diagnostics_computer(self) -> driver_io.DiagnosticsComputer:
         """Reuses its scratch/output buffers across output steps (allocated once)."""
         return driver_io.DiagnosticsComputer(grid=self.grid, backend=self.backend)
 
+    @functools.cached_property
+    def _compute_airmass(self) -> Callable[..., None]:
+        """Airmass program (``rho * ddqz_z_full``) with its static inputs bound.
+
+        The deep-atmosphere metric factor is identically 1 (shallow atmosphere).
```
OngChia: I suggest remove this

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton: I kept `deepatmo_t1mc` for now, because dropping it only here would be inconsistent with the rest of the code: the deep-atmosphere metrics are not excluded entirely: `driver_utils.py:430-434` uses  `DEEPATMO_DIVH`/`DEEPATMO_DIVZL`/`DEEPATMO_DIVZU` into `AdvectionMetricState`, and `compute_advection_metrics.py` computes them from the actual `grid_sphere_radius / radial_distance` (≈ 1 − z/R, not exactly 1). So tracer advection already runs with deep-atmosphere factors, while its airmass would not.

I removed the misleading docstring line instead. 

If you want to remove the argument from the `compute_airmass` stencil I would do it in a follow-up PR that removes the deep-atmosphere metrics everywhere and makes it consistent (feel free to go ahead with it if you want ;-) )

OngChia: (Just a reminder that deepatmos mode is not ported for dycore. We also need to modify quite some of the existing metric fields if deepatmos is turned on.)
I guess the question is if we are okay with porting the deepatmo mode in advection only while not sure whether we will port the entire deepatmos mode in the near future. I am okay with that. However, I think it does not make sense to run advection with deepatmos while the rest of the model in non-deepatmos mode, because the velocity field will be inconsistent with dycore not considering deepatmos while advection does, except maybe for some idealized advection test cases. Even worse is that advection metrics fields include deepatmos calculations while compute_airmass assumes value 1 for deepatmos metric..
Maybe we need to add an option to turn it off or a warning message when running from the driver because the deepatmos infrastructure is not entirely ready yet. When turning it off, those metric fields will be set to 1.

OngChia: So, this will potentially become a bug. I am not sure how large impact it will be when we run the driver with tracer advection and microphysics and compare the results with serialized data. Maybe the difference is small as atmosphere is shallow in APE?

model/standalone_driver/src/icon4py/model/standalone_driver/driver_states.py
jcanton: >   The duplication has already produced a disconnected data path. In standalone_driver.py:
>   - dycore accumulates into prep_adv: PrepAdvection (dycore_states.py:215) during substepping — vn_traj, mass_flx_me, mass_flx_ic
>   - advection reads tracer_prep_adv: AdvectionPrepAdvState (tracer_advection_states.py:46) — vn_traj, mass_flx_me, mass_flx_ic
>   - driver_states.py:283 and :290 allocate both, separately, as zero_field
>   - grep across model/ finds no code that copies one into the other
> 
>   So the advection is fed zeros. Same three quantities, two boxes, no link. (Honest caveat: standalone_driver.py:277 carries # Precondition: passing data test with ntracer > 0, so this is likely known-incomplete rather than a shipped
>   regression — I'll confirm against PR 1404.) Note also the same quantity has two names across the boxes: dynamical_vertical_mass_flux_at_cells_on_half_levels vs mass_flx_ic.

model/standalone_driver/src/icon4py/model/standalone_driver/standalone_driver.py
jcanton: from a friend:

You are right on both counts, and I have verified them against the Fortran. Fixed in ff093e11.

**Confirmations**

- dycore has no deep-atmosphere mode: `solve_nonhydro.py` raises `NotImplementedError("Deep atmosphere mode not implemented")` for `deepatmos_mode`, and no dycore stencil references deepatmo.
- the advection metrics were computed deep unconditionally: `metrics_factory.py` registers `compute_advection_deepatmo_fields` with no switch, and `driver_utils.initialize_granules` fed those into `AdvectionMetricState`.
- ICON initialises `deepatmo_divh_mc`, `deepatmo_divzL/U_mc` and `deepatmo_vol_mc` to `1._wp` in `mo_nonhydro_state.f90`, and only overwrites them inside the `IF (ldeepatmo)` branch of `mo_vertical_grid.f90` (line 1706). So with `ldeepatmo = .FALSE.` all of them are exactly 1 — which is the mode we support.

So this was already a fidelity bug against `ldeepatmo = .FALSE.`, independent of the airmass question. Note also that the tracer advection granule tests already build `AdvectionMetricState` with `constant_field(1.0)` for all three (`tests/tracer_advection/utils.py`) to match the serialized reference data — the driver was the only place getting values different from 1.

**On "this will potentially become a bug"**

It is worse than a mismatch in one term, there were two inconsistencies:

1. the icon4py dycore rho update uses `divh = 1` implicitly, while `integrate_tracer_horizontally` used `divh != 1` — so the tracer flux divergence no longer matched the density change the dynamics actually produced (consistency with continuity broken; a constant tracer field would not stay constant),
2. `airmass = rho * ddqz_z_full * 1` while `apply_density_increment` used `divzL/divzU != 1`.

And it only became live with this PR, since `AdvectionPrepAdvState` now shares the dycore `PrepAdvection` buffers by reference instead of getting zero fields.

Magnitude (a = 6371 km, dz = 500 m), so not negligible for a mass-conserving scheme:

| z | 1 - divh | divzL/U deviation | vol (t1mc) deviation |
|---|---|---|---|
| 10 km | 0.16% | 8e-5 | +0.32% |
| 30 km | 0.47% | 8e-5 | +0.95% |
| 75 km | 1.17% | 8e-5 | +2.38% |

**Fix**

Rather than adding a switch, `initialize_granules` now passes `constant_field(1.0)` for the three factors. There is nothing to switch on: the dycore already refuses `deepatmos_mode`, so shallow is the only reachable state, and this matches what ICON gives us with `ldeepatmo = .FALSE.`. Driver, granule tests and airmass now all agree.

Verified with `model/standalone_driver/tests/standalone_driver/integration_tests/test_standalone_driver.py`, 4 passed on `gtfn_cpu`.

**Follow-up**

I am opening a separate PR to delete `compute_advection_metrics.py`, its `metrics_factory` provider and the `DEEPATMO_*` attributes, so nobody can pick those values up again and reintroduce this. I will leave the `deepatmo_*` arguments in the stencil signatures, since `mo_advection_stepping.f90` passes them unconditionally too and dropping them would diverge the port from the Fortran for no runtime gain. Will tag you on it.

jcanton: very verbose, but long story short I've asked it to fix everything:-)

jcanton (approved):

OngChia (approved):

PR #1405, opened 2026-07-29
Use multiple output domains for compute_rho_theta_pgrad_and_update_vn

havogt:


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
```diff
@@ -180,8 +180,9 @@ def _compute_rho_theta_pgrad_and_update_vn(
 ]:
     # TODO(havogt): it would be nice if we could shrink the start of the compute domain to `start_edge_lateral_boundary_level_7 <= dims.EdgeDim`,
     # but that would require to put the correct lateral boundary condition where this is consumed.
-    # TODO(havogt): most likely it is possible to remove the `end_edge_halo` bound here (and shrink the compute domain), the corresponding
-    # Fortran code states "Initialize halo edges with zero in order to avoid access of uninitialized array elements".
+    # Note: the zeros outside are required, the Fortran code states "Initialize halo edges with
+    # zero in order to avoid access of uninitialized array elements", therefore this cannot be
+    # expressed as an output domain.
```
havogt, line 185: I'd like to double-check this

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
```diff
@@ -180,8 +180,9 @@ def _compute_rho_theta_pgrad_and_update_vn(
 ]:
     # TODO(havogt): it would be nice if we could shrink the start of the compute domain to `start_edge_lateral_boundary_level_7 <= dims.EdgeDim`,
     # but that would require to put the correct lateral boundary condition where this is consumed.
```
havogt, line 182: Would be interesting to discuss this in the team.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
havogt: Checked this against `mo_solve_nonhydro.f90`, the note is correct, and for a slightly stronger reason than the quoted comment alone gives.

Producer, predictor only (`IF (istep == 1)`, :894):
- zeroes the `min_rledge_int-2` (halo level 2) blocks, :899-913, which is where the quoted comment sits
- zeroes everything before the compute start for nested and limited area runs (`IF (jg > 1 .OR. l_limited_area)`), :926-940
- computes over `rl_start = 7` to `rl_end = min_rledge_int-1`, :919-920, the same range as the `concat_where` condition here

Consumer, :1704-1705: `rl_start = 5`, `rl_end = min_rledge_int - 2`, computing `mass_fl_e = z_rho_e * ...` at :1820.

So the consumed range is wider than the computed range on both ends: edges at refin_ctrl 5 and 6 below, and halo level 2 above, are read but never computed. ICON writes zeros there and repeats it in every predictor step. An output domain only restricts writes, so those regions would keep the values of the previous step instead of zeros. The `concat_where` therefore cannot be replaced by an output domain.

Two details that came up while checking:

- The blanket `z_rho_e = 0._wp` at :409 is guarded by `IF (p_test_run)`, so it is a debug path and does not make the per-step zeroing redundant.
- `z_rho_e` is halo exchanged, but only in the predictor (:1686), after the compute and before the consumer. In the corrector the consumer reads what the predictor left behind, which is a further reason the zeroed regions have to be well defined.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
havogt: For the discussion: the Fortran shows the obstacle concretely. The consumer of `z_rho_e` and `z_theta_v_e` starts at refin_ctrl 5 (`mo_solve_nonhydro.f90:1704`) while the producer starts at 7 (:919). The edges at refin_ctrl 5 and 6 are read but never computed, and ICON covers them by zeroing everything before the compute start for nested and limited area runs (:926-940).

So the "correct lateral boundary condition where this is consumed" is currently just those zeros. Shrinking the start here would mean supplying them at the consumer instead.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_edge_diagnostics_for_dycore_and_update_vn.py
```diff
@@ -531,10 +519,24 @@ def compute_rho_theta_pgrad_and_update_vn(
...
+            },
+            {
+                dims.EdgeDim: (start_edge_nudging_level_2, end_edge_local),
+                dims.KDim: (vertical_start, vertical_end),
+            },
+            {
+                dims.EdgeDim: (start_edge_lateral_boundary, end_edge_local),
+                dims.KDim: (vertical_start, vertical_end),
+            },
+        ),
```
jcanton, line 539: is this positional only (prone to mistakes?)? or can be indexed by kw somehow?

havogt: currently only this is possible

jcanton: any plans for kw support?

havogt: Actually not trivial to find a syntax that works. Note, you can have out arguments that are tuples, so the syntax needs to support that case as well.

havogt:
cscs-ci run benchmark-bencher

jcanton (approved):
almost forgot

PR #1409, opened 2026-07-30
Serialization improvements

jcanton:
from a friend:

Tooling around the serialized test data.

- Datatest runs print, in the terminal summary, the ICON revision behind every archive they used. It comes from the slurm log already inside each archive, so it works for everything ever published without regenerating anything. The string is a `git describe`, so `git log OLD..NEW -- <path>` explains a changed reference value directly.
- `ExperimentDescription.version` loses its shared default: experiments are versioned and regenerated individually, so bumping one no longer invalidates the other five.
- `run-serialization` gains `--dry-run`, a preflight that prints the ICON source tree revision and refuses a modified icon4py checkout, and per-task failure isolation so one bad task no longer aborts an 18-task campaign.
- `scripts/python/inspect_savepoints.py`: inspect the savepoints and fields of an archive.
- `docs/testdata_generation.md` replaces the HackMD page.


model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -156,7 +156,10 @@ class ExperimentDescription:
     name: str
     long_name: str
     grid: GridDescription
-    version: int = 5
+    # Version of the published serialized-data archive. Deliberately without a default:
+    # experiments are regenerated individually, and a shared default means bumping one
+    # of them silently invalidates the archives of all the others.
```
jcanton: remove these comments, unnecessary

scripts/python/run_serialization.py
```diff
@@ -41,6 +42,31 @@
...
 
+RUN_SUMMARY_FNAME: Final = "run_summary.json"
+REPORTS_DIRNAME: Final = "reports"
+
+
+@dataclasses.dataclass(frozen=True)
+class TaskResult:
+    """Outcome of a single (experiment, comm_size) serialization task."""
+
+    experiment: str
```
jcanton: if string, then

```suggestion
    experiment_name: str
```
experiment is for Experiment

scripts/python/run_serialization.py
```diff
@@ -88,12 +119,12 @@ def defaults(cls) -> SerializationSettings:
         SBATCH_UENV_VIEW = "default"
         JOB_POLL_SECONDS = 10
 
-        # Directories (derived from this script's location in icon4py/)
+        # Directories (derived from this script's location in the icon4py checkout).
+        # The checkout is expected to sit next to 'icon' and 'build_serialize' inside the
+        # icon-exclaim tree; that layout is validated where it is needed, so that the
+        # command still works from a checkout under another name.
```
jcanton: are these comments actually useful if
the command still works from a checkout under another name?

scripts/python/run_serialization.py
```diff
@@ -505,13 +557,95 @@ def generate_update_script(
...
+    comm_size: int,
+    job_id: str | None,
+    *,
+    settings: SerializationSettings,
+) -> pathlib.Path:
+    """Write the machine-readable identity of an archive next to its 'ser_data'.
+
+    The ICON revision is taken from the version banner in the slurm log, which is the
+    only artifact with a verified link to the binary that produced the data.
+    """
```
jcanton: docstring is redundant: code is already self explaining

scripts/python/run_serialization.py
```diff
@@ -505,13 +557,95 @@ def generate_update_script(
...
+        "provenance": {
+            "icon": serdata.read_icon_banner_from_archive(dest_dir),
+            "icon4py": serdata.harvest_git(settings.icon4py_repo_dir),
+            "runtime": {
+                "slurm_job_id": job_id,
+                "uenv": settings.sbatch_uenv,
+                "uenv_view": settings.sbatch_uenv_view,
+                "partition": settings.sbatch_partition,
+                "account": settings.sbatch_account,
+            },
```
jcanton: any other useful info that we should include?

scripts/python/run_serialization.py
```diff
@@ -505,13 +557,95 @@ def generate_update_script(
...
     *,
     settings: SerializationSettings,
-) -> None:
-    """Execute a single experiment with the given communicator size."""
+) -> TaskResult:
+    """Execute a single experiment with the given communicator size.
+
+    Failures are reported back rather than raised: a campaign is 18 slurm tasks over
+    several hours, and one broken task must not discard the others.
+    """
```
jcanton: revert docstring .
even better: remove entirely, doesn't add anything

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+        tar_path = tar_folder(dest_dir, experiment_description, comm_size, settings=settings)
+
+        log_status(f"Comparing {experiment_description.name} with {comm_size} ranks")
+        verdict, report_path = compare_with_previous_version(
+            dest_dir, experiment_description, comm_size, settings=settings
+        )
 
         log_status(f"Completed {experiment_description.name} with {comm_size} ranks")
+        return TaskResult(
+            experiment=experiment_description.name,
```
jcanton:
```suggestion
            experiment_name=experiment_description.name,
```

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+            version=experiment_description.version,
```
jcanton:
```suggestion
            experiment_version=experiment_description.version,
```

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+            comm_size=comm_size,
+            status="failed",
+            error=f"{type(e).__name__}: {e}",
+        )
+
+
+def report_results(results: list[TaskResult], *, settings: SerializationSettings) -> None:
+    """Print the per-task outcome table and persist it next to the archives."""
```
jcanton:
```suggestion
```

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+        json.dump([result.as_dict() for result in results], f, indent=4)
+    log_status(f"Wrote task summary to {summary_path}")
+
+
+def print_next_steps(results: list[TaskResult], *, settings: SerializationSettings) -> None:
+    """Spell out what still has to happen by hand.
+
+    Regeneration happens a few times a year; the steps after it are the ones that get
+    forgotten, so they are printed with the values already filled in.
+    """
```
jcanton:
```suggestion
```

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+    if reviewable:
+        print(f"  1. Review {len(reviewable)} archive(s) whose comparison is not OK:")
+        for result in reviewable:
+            print(f"       {result.verdict:<11} {result.report_path}")
+    print("  2. Run the datatests against the fresh data before publishing anything:")
+    print(
+        f"       ICON4PY_TEST_DATA_PATH={settings.experiments_dir} "
+        "ICON4PY_ENABLE_TESTDATA_DOWNLOAD=0 \\\n"
+        "         uv run --group test --frozen pytest -n0 --datatest-only "
+        "--backend=gtfn_cpu model/common"
```
jcanton: model/common ?
-n0 ?

scripts/python/run_serialization.py
```diff
@@ -543,17 +677,101 @@ def run_experiment(
...
+    )
+    print("  3. Upload the archives (see docs/testdata_generation.md):")
+    for result in results:
+        print(f"       {result.tar_path}")
+    print("  4. Set these versions in model/testing/src/icon4py/model/testing/definitions.py:")
+    for experiment, version in sorted(versions):
+        print(f"       {experiment}: version={version}")
```
jcanton: the versions should have been read from there, so there should not be any need to set them, or?

scripts/tests/python/test_serdata.py
```diff
@@ -0,0 +1,493 @@
...
+
+@pytest.fixture(scope="module")
+def banner_log() -> str:
+    return (DATA_DIR / "LOG.banner_sample.o").read_text()
+
+
+def test_parse_banner_extracts_icon_revision(banner_log: str) -> None:
+    banner = parse_icon_log_banner(banner_log)
+
+    assert banner["sha"] == "97986bc3592cc05c799717c70345f27a8c275d8d"
```
nfarabullini: perhaps we should put all of these long strings in a separate file?

docs/testdata_generation.md
```diff
@@ -0,0 +1,284 @@
+# Generating serialized test data
+
+icon4py's datatests validate against Serialbox dumps produced by an instrumented ICON
```
nfarabullini:
```suggestion
Icon4py's datatests validate against Serialbox dumps produced by an instrumented ICON
```

docs/testdata_generation.md
```diff
@@ -0,0 +1,284 @@
...
+### Pin the ICON branch you generate from
+
+Serialization branches on `git@gitlab.dkrz.de:icon/icon-nwp.git` get rebased and deleted.
+**Push a tag for whatever commit you generate from**, otherwise the revision recorded in
+`archive_metadata.json` will eventually point at nothing.
+
+This matters: in June 2026 a regeneration silently picked up `a9435ae531`
+("[nwp] Improve wave energy propagation near the coast"), which moved three geometry
+fields from `rl_start=2` to `rl_start=1` in `src/shr_horizontal/mo_intp_coeffs.f90` and
+broke three icon4py datatests with no recorded provenance to explain it. That incident is
```
nfarabullini:
```suggestion
broke three ICON4Py datatests with no recorded provenance to explain it. That incident is
```

model/testing/src/icon4py/model/testing/definitions.py
nfarabullini:
```suggestion
```
we don't need this, the name is self-explanatory

scripts/python/run_serialization.py
nfarabullini: too long of a comment

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+
+    if comparison.unclassified:
+        lines.append(f"UNCLASSIFIED savepoints: {', '.join(comparison.unclassified)}")
+
+    return "\n".join(lines)
+
+
+# ---------------------------------------------------------------------------
+# Backfilling archives generated before this metadata existed
+# ---------------------------------------------------------------------------
```
jcanton: to be removed once not necessary anymore

scripts/python/run_serialization.py
```diff
@@ -505,13 +557,95 @@ def generate_update_script(
...
+    reconstruct either.
+
+    The baseline is the previous version if it is still on disk. When it is not, the
+    report says so rather than reporting a clean comparison that never happened; the
+    previous archive can be extracted and passed to 'serdata diff --baseline' by hand.
+    """
+    previous = dataclasses.replace(
+        experiment_description, version=experiment_description.version - 1
+    )
+    baseline_dir = get_serdata_dst_dir(previous, comm_size, settings=settings)
```
nfarabullini: Does `get_serdata_dst_dir` output `None`? Maybe we could put it in there as a possible return option and we don't need to check where for line 623 if statement

scripts/python/run_serialization.py
```diff
@@ -505,13 +557,95 @@ def generate_update_script(
...
+
+    comparison = serdata.compare_archives(
+        baseline_dir if baseline_dir.is_dir() else None,
+        dest_dir,
+        experiment=experiment_description.name,
+    )
+    rendered = serdata.render_diff_report(comparison)
+    print(rendered)
```
nfarabullini:
```suggestion
```

jcanton (commented):
is the tests/data necessary or once we backfill we can run tests against data on the server and pick one experiment and test current version against previous version (the server always has the two most recent or more)

scripts/python/run_serialization.py
```diff
@@ -563,41 +781,59 @@ def run_serialization() -> None:
...
         log_status(
-            f"Completed communicator size {rank_idx}/{len(settings.comm_sizes)}: {comm_size} ranks"
+            f"Completed communicator size {rank_idx}/{len(settings.comm_sizes)}: {comm_size_value} ranks"
         )
 
+    report_results(results, settings=settings)
+    failed = [result for result in results if result.status == "failed"]
+    if failed:
+        log_status(f"{len(failed)}/{len(results)} tasks failed.")
+        raise typer.Exit(code=1)
```
nfarabullini, line 680: here you could write a better error message than just `typer.Exit(code=1)`

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+@dataclasses.dataclass
+class Fingerprint:
+    """Per-field digests of one archive, keyed per rank."""
+
+    records: dict[RecordKey, str]
+    classes: dict[str, SavepointClass]
+    ranks: int
+    # Savepoints that look time-invariant but are not in 'STATIC_SAVEPOINTS'. They are
+    # compared as 'evolving', which means they are not guarded, so they have to be
+    # classified by a human before the next regeneration.
```
nfarabullini:
```suggestion
```

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+    archive = json.loads(archive_path.read_text())
+
+    savepoints = meta["savepoint_vector"]["savepoints"]
+    fields_per_savepoint = meta["savepoint_vector"]["fields_per_savepoint"]
+    if len(savepoints) != len(fields_per_savepoint):
+        raise ValueError(f"Inconsistent savepoint vector in '{meta_path}'.")
+
+    # 'fields_table[field]' is an ordered list of [byte offset, digest]; the occurrence
+    # index used by 'fields_per_savepoint' is the list position, and element 0 is an
+    # offset into the .dat file, never an identifier to look up.
```
nfarabullini:
```suggestion
```

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+            key: value["value"] for key, value in (savepoint.get("meta_info") or {}).items()
+        }
+        described.append((name, meta_info))
+        classes[name] = classify_savepoint(name, meta_info)
+        # 'field_map' is a superset of what was actually written; only the fields listed
+        # per savepoint are guaranteed to have a digest.
+        for field, occurrence in (fields[name] or {}).items():
+            # Digests are Serialbox's own 32 byte tokens, printed byte by byte with
+            # leading zeros stripped. They are 54-64 characters long, are not
+            # interchangeable with 'hashlib.sha256', and must never be padded.
```
nfarabullini:
```suggestion
```

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+    savepoints_added: list[str]
+    savepoints_removed: list[str]
+    fields_added: list[str]
+    fields_removed: list[str]
+
+
+def diff_fingerprints(old: Fingerprint, new: Fingerprint) -> FingerprintDiff:
+    """Compare two archives record by record, split by savepoint class."""
+    classes = {**old.classes, **new.classes}
+
```
nfarabullini: I'm not a huge fan of all of these functions-within-functions. If you could reduce the amount, that would make for much cleaner code

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+                changed.append((section, key, old_section[key], new_section[key]))
+
+    return NamelistDiff(changed=changed, added=added, removed=removed)
+
+
+# ---------------------------------------------------------------------------
+# Comparing two archives
+# ---------------------------------------------------------------------------
+
+ARCHIVE_METADATA_FNAME: Final = "archive_metadata.json"
```
nfarabullini: I would place all of these at the top of the file

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+            json.dump(metadata, f, indent=4)
+        written.append(metadata_path)
+
+    return written
+
+
+# ---------------------------------------------------------------------------
+# Command line
+# ---------------------------------------------------------------------------
+
```
nfarabullini: I would place all cli commands in a separate file called `cli_commands.py`

scripts/tests/python/data/v06/NAMELIST_ICON_output_atm.json
```diff
@@ -0,0 +1,18 @@
+{
+ "grid_nml": {
+  "nroot": 2,
+  "start_lev": 4,
+  "lfeedback": [
+   true,
+   true
```
nfarabullini: why double true here?

model/testing/src/icon4py/model/testing/definitions.py
jcanton: from a friend: Removed.

jcanton: from a friend: Removed.

scripts/python/run_serialization.py
jcanton: from a friend: Applied: `experiment_name`, and `version` -> `experiment_version` alongside it.

scripts/python/run_serialization.py
jcanton: from a friend: Applied.

scripts/python/run_serialization.py
jcanton: from a friend: Applied.

scripts/python/run_serialization.py
jcanton: from a friend: You're right, it explained a non-problem. Cut to a single line.

jcanton: from a friend: Cut to a single line.

scripts/python/run_serialization.py
jcanton: from a friend: Docstring removed. I kept two `#` lines saying the ICON revision comes from the slurm log rather than from the repo checkout, because that is the reason the field can be trusted at all and it is not visible from the code. Say the word and it goes too.

scripts/python/run_serialization.py
jcanton: from a friend: Removed entirely.

scripts/python/run_serialization.py
jcanton: from a friend: Removed.

scripts/python/run_serialization.py
jcanton: from a friend: Removed.

scripts/python/run_serialization.py
jcanton: from a friend: Dropped `-n0` and added a comment on why `model/common`: it holds the grid, metric and interpolation datatests, which is where a change in a published archive lands first.

scripts/python/run_serialization.py
jcanton: from a friend: Correct, and it was worse than redundant: it printed back the versions it had just read from `definitions.py`, so it could never say anything new. Step removed.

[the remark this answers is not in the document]
jcanton: from a friend: Fixed by filtering to `status == "ok"`. It could not fire as written, since the call sits after the early exit on failures, but the guard removes the class. Regression test added.

scripts/python/run_serialization.py
```diff
@@ -563,41 +781,59 @@ def run_serialization() -> None:
...
+
     log_status(f"All {total_tasks} tasks completed successfully!")
```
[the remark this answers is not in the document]
jcanton: from a friend: Correct and my bug: the edit that was supposed to add the call silently did not apply. Now called, with a test asserting the call site exists.

scripts/python/serdata.py
```diff
@@ -0,0 +1,812 @@
...
+    """Reconstruct metadata for archives generated before it existed."""
+    written = backfill_archive_metadata(testdata_root)
+    for path in written:
+        print(f"  wrote {path}")
+    print(f"{len(written)} archive(s) backfilled")
+
+
+if __name__ == "__main__":
+    sys.exit(cli())
+
```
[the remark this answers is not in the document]
jcanton: from a friend: Correct and my bug. The `__main__` guard now sits at the end of the file, and a test asserts no `@cli.command()` follows it. `./scripts/python/serdata.py --help` lists all commands again; I had only ever exercised the CLI through `./scripts/run`, which imports the module and hides this.

scripts/tests/python/test_serdata.py
jcanton: from a friend: Moved to named module-level constants at the top of the test file rather than a separate file, so the assertions stay readable in place.

docs/testdata_generation.md
jcanton: from a friend: Applied as `ICON4Py`, which is the spelling in `CODING_GUIDELINES.md` and the license headers (this suggestion said `Icon4py` and the one below said `ICON4Py`).

docs/testdata_generation.md
jcanton: from a friend: Applied.

scripts/python/run_serialization.py
jcanton: from a friend: I would rather not: `get_serdata_dst_dir` is also used by `cleanup_exp_output` and `copy_ser_data`, where the directory legitimately does not exist yet because they are about to create it, so returning `None` would break both and mix path construction with an existence check. Extracted `previous_version_dir(...) -> Path | None` instead, so the caller reads clean.

scripts/python/run_serialization.py
jcanton: from a friend: Assuming this targets the docstring paragraph: removed, the function is now a one-liner docstring plus the extracted helper. Reopen if you meant a different line.

scripts/python/run_serialization.py
jcanton: from a friend: Now names the tasks that failed and points at `run_summary.json`.

scripts/python/serdata.py
jcanton: from a friend: Shortened to one line.

scripts/python/serdata.py
jcanton: from a friend: Shortened to one line. Keeping the fact itself: `fields_table[field][i][0]` is a byte offset, not an index, and keying a dict on it raises on the first repeatedly written field.

scripts/python/serdata.py
jcanton: from a friend: Shortened to one line. Keeping the fact: Serialbox strips leading zeros per byte, so the digests are 54-64 characters and padding them silently corrupts every comparison.

scripts/python/serdata.py
jcanton: from a friend: Agreed. Lifted the nested helpers out of `harvest_git`, `diff_fingerprints`, `_render_provenance` and `summarise_differences`.

scripts/python/serdata.py
jcanton: from a friend: Moved to the top of the module.

scripts/tests/python/data/v06/NAMELIST_ICON_output_atm.json
jcanton: from a friend: It is a per-domain array in a synthetic fixture, present so the namelist diff exercises list values rather than scalars. Changed to `[True, False]` so it no longer reads as an accidental duplicate.

scripts/python/serdata.py
jcanton: from a friend: Marked transitional in the source, with the condition for removing it: once every published archive carries `archive_metadata.json`.

scripts/python/run_serialization.py
jcanton: from a friend: Leaving the contents as they are for now, per your comment that this was a question for the other reviewers. Candidates if anyone wants them: resolved grid/extpar paths and their checksums, per-class record counts, the tarball sha256, and the `icon4py_verification_nml` block.

scripts/python/serdata.py
jcanton: from a friend: Agreed, and part 2 already does exactly this: the library moves to `icon4py.model.testing.serialized_data` (a datatest has to import it, and `scripts/` is not importable from one) and `scripts/python/serdata.py` becomes CLI-only. Leaving the split there rather than reshaping this PR into a form part 2 immediately replaces.

jcanton: but part 2 is being deleted and lots of this has been removed too, so resolving

scripts/tests/python/test_inspect_savepoints.py
```diff
@@ -0,0 +1,82 @@
...
+
+def test_summarize_excludes_non_finite_values_from_the_range():
+    # A single NaN would otherwise swallow the min and max of the whole field.
+    stats = _summarize(np.array([1.0, np.nan, 3.0, np.inf]))
+
+    assert (stats.min, stats.max) == (1.0, 3.0)
+    assert stats.n_nonfinite == 2
+
+
+class _FakeSerializer:
```
nfarabullini: do you really need a class for such small tests? It seems a bit heavy

model/testing/src/icon4py/model/testing/provenance.py
```diff
@@ -0,0 +1,45 @@
...
+    The full describe string rather than the bare commit: it is a valid git revision,
+    so 'git log OLD..NEW' works on it directly.
+    """
+    for log in sorted(archive_dir.glob("LOG.*.o")):
+        with log.open(errors="replace") as lines:
+            for line in lines:
+                match = _REVISION.match(line.rstrip("\n"))
+                if match:
+                    return match[1]
+    return None
```
nfarabullini: why not return empty string?

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -243,10 +243,29 @@ def pytest_runtest_makereport(item, call):
...
+        return
+    terminalreporter.ensure_newline()
+    terminalreporter.line(" Serialized test data ", bold=True, blue=True)
+    for archive, revision in sorted(provenance.seen.items()):
+        terminalreporter.line(f"  {archive:<44} {revision}")
+
+
+# 'optionalhook' because the archives are downloaded in the xdist workers while the
+# summary is rendered on the controller, but this plugin is loaded for every session
+# in the repository, including those without xdist installed.
```
nfarabullini: summarize

model/testing/src/icon4py/model/testing/provenance.py
jcanton: None makes mypy flag a caller that forgets the missing case, while returning "" would let one silently print a blank column in the provenance table...
in a PR whose whole point is not showing people a wrong or empty provenance, None means "there was no revision", while "" says "the revision is the empty string" ?

model/testing/src/icon4py/model/testing/provenance.py
```diff
@@ -0,0 +1,45 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
```
msimberg: I'm really not sure what this file is here for. Can you explain? I only see `read_icon_revision` used in one test.

If it really is needed, wouldn't it be easier to cache `read_icon_revision` instead of the manual cache-like thing with `record`?

msimberg (commented):
No important comments. Question on the provenance tracking.

model/testing/src/icon4py/model/testing/provenance.py
OngChia: I also have the same question. I am curious about what is read_icon_revision used for?

scripts/python/run_serialization.py
```diff
@@ -547,13 +556,68 @@ def run_experiment(
...
+    )
 
 
 @cli.command()
-def run_serialization() -> None:
+def run_serialization(
+    dry_run: Annotated[
+        bool, typer.Option("--dry-run", help="List the archives that would be written, then stop.")
+    ] = False,
+    allow_dirty: Annotated[
```
OngChia, line 620: Maybe you have told me before. And sorry if this is a stupid question, as I have not generated serialized data for a long time. May I know the reason for raising an error when a developer wants to generate the serialized data with modified icon4py checkout? Is it because you will record the icon version in icon4py using `read_icon_revision` function and compare it with the icon build that generates the serialized data? Therefore, it is important to check that you are did not modify icon4py?

OngChia (commented):
Looks good. I think you have used the script and know if this works well. I just left two-related questions because I didn't have enough understanding of how it works.

scripts/python/run_serialization.py
jcanton: exactly: if you generate data without having committed your changes it will record the commit that does not match the code (both for icon and icon4py)

model/testing/src/icon4py/model/testing/provenance.py
jcanton: Leftovers. `read_icon_revision` did have a second caller — `record()` right below it, which `datatest_utils.download_experiment` called once per datatest. `record` kept a session-wide `{archive: revision}` dict that `pytest_terminal_summary` printed as a "Serialized test data" block, to inform where the used data is coming from.

That was part of a larger idea: having the test session itself track and report which ICON build produced the data it was validating against. We removed that and just record the provenance in the archive — the slurm log ICON writes carries its own `git describe`, the data version is pinned, and CI never runs against data nobody has looked at. When new data does turn up and a test starts disagreeing, the revision is recorded in the archive and both repos can be bisected by hand, which is what `docs/testdata_generation.md` describes. These two functions were simply left behind.

🧑 and edited by @jcanton himself

model/testing/src/icon4py/model/testing/provenance.py
jcanton: Moot now: the whole module is gone as of e60b2421d, see the thread below.

scripts/python/inspect_savepoints.py
```diff
@@ -0,0 +1,585 @@
...
+from collections.abc import Sequence
+from typing import TYPE_CHECKING, Annotated, Final
+
+import typer
+
+
+if TYPE_CHECKING:
+    import numpy as np
+
+    from icon4py.model.testing import definitions as test_defs
```
nfarabullini: since this is also imported in `resolve_experiment`and `experiment_data_path`, can you take it out of the if statement directly and remove the import below?

scripts/python/inspect_savepoints.py
```diff
@@ -0,0 +1,585 @@
...
+            stats.savepoint,
+            stats.date or "",
+            stats.full_name,
+            "x".join(str(size) for size in stats.shape),
+            _number(stats.min),
+            _number(stats.max),
+            _number(stats.mean),
+            f"{stats.nonzero_fraction:.3f}",
+            str(stats.n_nonfinite) if stats.n_nonfinite else "",
+            "ALL ZERO" if stats.all_zero else "",
```
nfarabullini:
```suggestion
            "ALL ZEROS" if stats.all_zero else "",
```

scripts/tests/python/test_inspect_savepoints.py
```diff
@@ -0,0 +1,74 @@
...
+    return summarize(values, savepoint="sp", date=None, field="f", component=None)
+
+
+def test_summarize_reports_range_and_mean():
+    stats = _summarize(np.array([[1.0, 2.0], [3.0, 4.0]]))
+
+    assert (stats.min, stats.max, stats.mean) == (1.0, 4.0, 2.5)
+    assert stats.shape == (2, 2)
+    assert stats.nonzero_fraction == 1.0
+    assert not stats.all_zero
```
nfarabullini: do you need both here? Or rather, shouldn't ` assert not stats.all_zero` go before `assert stats.nonzero_fraction == 1.0`?

scripts/python/inspect_savepoints.py
```diff
@@ -0,0 +1,585 @@
...
+def _transition(before: str | None, after: str | None) -> str | None:
+    """Label a pair of endpoints, collapsing to a single value when they are equal."""
+    return after if before == after else f"{before} -> {after}"
+
+
+def summarize(
+    values: np.ndarray, *, savepoint: str, date: str | None, field: str, component: str | None
+) -> FieldStats:
+    """Reduce an array to the statistics that tell whether it carries information."""
+    import numpy as np  # noqa: PLC0415 [import-outside-top-level]
```
nfarabullini: same here. I think this import can be easily pulled out

scripts/python/inspect_savepoints.py
```diff
@@ -0,0 +1,585 @@
...
+    def read(self, reference: SavepointRef, field: str) -> np.ndarray:
+        """Raw array of *field*, exactly as serialized (halo entries included)."""
+        return self._serializer.read(field, self._raw[reference.index])
+
+    def find(
+        self, *, name: str | None = None, date: str | None = None, field: str | None = None
+    ) -> list[SavepointRef]:
+        """Savepoints matching all given filters; *name* and *date* match on substrings."""
+        found = self.savepoints
+        if name is not None:
```
nfarabullini, line 277: IMO if you set variables as `None`s by default, you kind of allow that variable to be None and later be set to something else within the function. I saw a few places where a variable is set to None as default and, without being changed inside the function, other actions are performed in case the value is not None. I would say that instead of doing all of this, you can remove the None default  (and make the variable a positional argument rather than keyowrd) and thus force a setting of a value upfront. This way you can remove a fair amount of None checks

nfarabullini (commented):
some comments on my side

scripts/python/run_serialization.py
```diff
@@ -207,7 +203,16 @@ def cleanup_exp_output(
 def run_command(
     cmd: list[str], check: bool = True, cwd: pathlib.Path | None = None
 ) -> subprocess.CompletedProcess:
-    return subprocess.run(cmd, check=check, text=True, capture_output=True, cwd=cwd)
+    result = subprocess.run(cmd, check=False, text=True, capture_output=True, cwd=cwd)
+    if check and result.returncode != 0:
+        # 'CalledProcessError' reports the exit status but not the captured output, and
+        # the output is the part that says what went wrong.
```
nfarabullini: I think this comment can be removed

nfarabullini (commented):
one more comment

scripts/python/inspect_savepoints.py
jcanton: Measured this: importing `icon4py.model.testing.definitions` transitively pulls the GT4Py grid stack and costs ~7s. Top-level it would make every CLI call (including `--help` and `savepoints`) pay that; lazy, only experiment-resolving commands do. The TYPE_CHECKING import exists so mypy still sees the annotations, and the runtime import only happens where the module is actually needed. Keeping it lazy.

scripts/python/inspect_savepoints.py
jcanton: These are optional filters fed by optional CLI flags (`-n`/`-d`/`-f`), so None is a real value typer passes when the flag is absent; the `if x is not None` checks are the filtering logic itself, not boilerplate. Making the parameters required positional would push the None onto every call site without removing any check. Same for `version` in `for_experiment`, where None means "use the version pinned in definitions". Keeping the None defaults.

scripts/python/inspect_savepoints.py
jcanton: this is also already lazily imported in run_serialization for the same reason

scripts/python/inspect_savepoints.py
```diff
@@ -0,0 +1,576 @@
...
+    timestamps are. With '--against' the comparison is against another savepoint at
+    the same dates instead, which shows whether the process in between changes the
+    field at all (for instance an init against its exit).
+    """
+    archive = open_archive(experiment, version, comm_size, rank, path)
+    references = _one_savepoint(archive.find(name=name, field=field), field, "--name")
+
+    if against is not None:
+        others = _one_savepoint(archive.find(name=against, field=field), field, "--against")
+        by_date = {reference.date: reference for reference in others}
```
yiluchen1066: here ```by_date``` assumes every savepoint has a unique date. But just to confirm, is that always true?

model/testing/src/icon4py/model/testing/definitions.py
```diff
@@ -203,7 +203,7 @@ class Experiments:
         name="exclaim_ch_r04b09_dsl",
         long_name="Regional setup used by EXCLAIM to validate the icon-exclaim.",
         grid=Grids.MCH_CH_R04B09_DSL,
-        version=6,
+        version=7,
```
yiluchen1066: I saw in the doc, it says: 
````ExperimentDescription.version` in `model/testing/.../definitions.py` has no default: each
experiment is versioned and regenerated on its own.```
shall we here remove the default in this pr, with each experiment getting its won explicit ```version=```

File: scripts/python/inspect_savepoints.py [hunk 42]
```diff
@@ -0,0 +1,576 @@
...
+def _transition(before: str | None, after: str | None) -> str | None:
+    """Label a pair of endpoints, collapsing to a single value when they are equal."""
+    return after if before == after else f"{before} -> {after}"
+
+
+def summarize(
+    values: np.ndarray, *, savepoint: str, date: str | None, field: str, component: str | None
+) -> FieldStats:
+    """Reduce an array to the statistics that tell whether it carries information."""
+    values = np.squeeze(values)
```
yiluchen1066:
```suggestion
```
I think none of the statistics below depend on the shape, so the ```squeeze``` only effect is that ```shape=tuple(values.shape)``` after squeezing. We can drop it safely makes the printed shape the serialized one, right?

docs/testdata_generation.md
```diff
@@ -0,0 +1,94 @@
...
+git -C <icon checkout> log --oneline --no-merges "$OLD".."$NEW" -- src/shr_horizontal/mo_intp_coeffs.f90
+```
+
+Scope by the file the failing test's field comes from, not by subsystem: the commit that
+last moved three geometry fields was titled *"[nwp] Improve wave energy propagation near the
+coast"*, so only the path is decisive.
+
+## Versions
+
+`ExperimentDescription.version` in `model/testing/.../definitions.py` has no default: each
```
yiluchen1066, line 34: see the comment above, we should make the doc aligning withe the codes

docs/testdata_generation.md
```diff
@@ -0,0 +1,94 @@
...
+```bash
+ICON4PY_TEST_DATA_PATH=<...>/build_serialize/experiments ICON4PY_ENABLE_TESTDATA_DOWNLOAD=0 \
+  uv run --group test --frozen pytest --datatest-only --backend=gtfn_cpu model/common
+```
+
+Then upload, and open a PR with the version bump:
+
+```bash
+aws --profile cscs-icon4py s3 sync . s3://testdata/experiments/ --exclude "*" --include "*.tar.gz"
+```
```
yiluchen1066, line 69:
```suggestion
```bash
cd <output root> # the directory holding the fresh *.tar.gz
aws --profile cscs-icon4py s3 sync . s3://testdata/experiments/ --exclude "*" --include "*.tar.gz"
```
```
then it is safer for people to do ```*.tar.gz```

File: scripts/python/run_serialization.py [hunk 45]
```diff
@@ -82,18 +81,15 @@ def defaults(cls) -> SerializationSettings:
...
+        SBATCH_UENV = "icon/26.7:v1"
         SBATCH_UENV_VIEW = "default"
         JOB_POLL_SECONDS = 10
 
         # Directories (derived from this script's location in icon4py/)
         _THIS_FILE = pathlib.Path(__file__).resolve()
         ICON4PY_REPO_DIR = _THIS_FILE.parents[2]
-        assert ICON4PY_REPO_DIR.name == "icon4py", (
-            f"Expected icon4py repo dir, got {ICON4PY_REPO_DIR}"
-        )
```
yiluchen1066, line 96: I think here is the only check that ```parents[2]``` really is the repo root, shall we keep it?

yiluchen1066 (commented):
Thanks for the PR! Looks good to me! I left some small points, some are also my questions

model/testing/src/icon4py/model/testing/definitions.py
jcanton: revert to 6, dunno how this got updated

docs/testdata_generation.md
```diff
@@ -0,0 +1,94 @@
...
+
+```bash
+ICON4PY_TEST_DATA_PATH=<...>/build_serialize/experiments ICON4PY_ENABLE_TESTDATA_DOWNLOAD=0 \
+  uv run --group test --frozen pytest --datatest-only --backend=gtfn_cpu model/common
+```
+
+Then upload, and open a PR with the version bump:
+
+```bash
+aws --profile cscs-icon4py s3 sync . s3://testdata/experiments/ --exclude "*" --include "*.tar.gz"
```
msimberg, line 68: Why not add the more detailed instructions about setting up awscli from the hackmd page here as well? Not at all a blocker for this PR.

I'd say this is all "developer documentation" that eventually belongs in the repo and a rendered docs page, so if we can slowly move towards keeping this kind of information in the repo that's a good direction I think.

msimberg (approved):
Looks good to me now. Minor comment about expanding the docs files, but that can be done separately (never?) as well, does not block this PR.

model/testing/src/icon4py/model/testing/definitions.py
jcanton: I have already removed the default

docs/testdata_generation.md
jcanton: I added the link to the doc, the hackmd for this is well written, here I only reported the one line that's useful to copy-paste when actually running the script.
if you prefer we can move the doc from hackmd to a setup_aws_and_ceph.md file in the repo though, but I would not include those instructions in this one

scripts/python/run_serialization.py
jcanton: it didn't work when you have worktrees and nested things, that's why it was gone, I re-introduced it looking for `pyproject.toml` instead, which should make it a bit more robust

scripts/python/inspect_savepoints.py
jcanton: yes you're right, it made the printed table look nicer but it's unnecessary, I removed it

scripts/python/inspect_savepoints.py
jcanton: good find, no it didn't: nonhydro savepoints also have dyn_substeps and istep so it didn't work for those, but I had only looked at physics savepoints and had not thought of that.
I added the missing fields

PR #1301, opened 2026-06-04
Physics driver with Muphys

yiluchen1066:
icon4py previously had no physics orchestration: the standalone driver ran a hardcoded solve_nh → diffusion → tracer_advection sequence. This PR adds:

1. A generic physics driver — a new workspace package ```icon4py-atmosphere-physics-driver```, modelled on ICON-MPIM's AES physics interface (aes_phy_main), which runs registered physics processes with per-process time control 
2. we plugged gt4py muphys as the first process into it, called from the standalone driver's time loop and validated end-to-end on the ```EXCLAIM_APE_AES``` moist-aquaplanet experiment against ICON serialized data.

Design overview: 
```
common/components/           Component + PhysicsState protocols (the generic contracts)
physics_driver/                       PhysicsDriver, PhysicsProcess, ForcingMode, ProcessTimeControl
muphys/                                  MuphysComponent (scheme wrapper, state-update → tendencies)
                                                 State (adapter: gather/diagnose inputs, scatter tendencies back)
standalone_driver/          builds the muphys PhysicsProcess and calls physics.run() per step
```
Physics runs once per timestep — after dynamics, diffusion and transport, just before the state swap, matching ICON. For each process, the driver gathers inputs from the prognostic state, checks the process's time control (its active window, and that the firing interval is an integer multiple of dtime — otherwise it raises), then either computes new forcing or reuses the cached one, and finally applies the tendencies. 

The muphys code is split in two: the component wraps the microphysics scheme and never sees the model state — it takes and returns plain field dicts, which is also how it's validated standalone against ICON savepoints. The state adapter owns all the model coupling. Components return tendencies (muphys converts its updated-state result to tendencies, with precipitation passed through as diagnostics), and tendencies are applied on the prognostic cells only, as in ICON.

One important thing: 
the driver runs without tracer transport, but the ICON reference run for EXCLAIM_APE_AES transported tracers (ltransport=.TRUE., MIURA/PPM). This is why qv carries a ~2e-6 residual — the other tracers match bit-for-bit. The MIURA/PPM advection the reference used is already ported and has its own passing tests. What's missing is the driver wiring for advection's inputs: airmass is never computed (the zero field makes advection divide by zero everywhere), and the dycore mass fluxes (vn_traj/mass_flx_me/mass_flx_ic) are never accumulated. This is fixed in PR #1401

PR #1410, opened 2026-07-31
Fix boundary-row computation of pos_on_tplane_e and primal/dual_normal_cell

yiluchen1066:
Icon4py computes the edge-geometry fields ```pos_on_tplane_e```, ```primal_normal_cell``` and ```dual_normal_cell``` on all edge rows, starting from the outermost boundary row (```rl_start = 1```), skipping only the missing cell neighbor of each boundary edge. Icon4py started these computations at ```LATERAL_BOUNDARY_LEVEL_2```, leaving row 1 at zero. This was not noticed as long as v05 serialized references also contained zeros there; the v06 references carry ICON's true row-1 values. This PR ports the ICON deviation faithfully: 

- **`pos_on_tplane_e`** : the computation now starts at `LATERAL_BOUNDARY` and guards each of the two neighbor-cell slots individually — the slot of a missing neighbor stays zero, the existing neighbor is still computed. Indexing uses a clipped connectivity so the `-1` skip value cannot wrap around silently.
  
- **`primal_normal_cell` / `dual_normal_cell`**:  the
  computation were moved from a GT4Py stencil to an ndarray function, registered as a `NumpyDataProvider` producing the sparse `(EdgeDim, E2CDim)` fields
  directly. A field operator cannot check whether an `E2C` neighbor exists, so the
  per-neighbor guard ICON applies on the boundary row is only expressible outside the
  stencil. Same formula and normalization as before, same registration pattern as
  `pos_on_tplane_e`. 
  
With #1301 merged (v06 test data is now the default), this branch is updated to
main and the interim test slices marked "TODO (Yilu): revert" are reverted here:
test_geometry.py, test_interpolation_fields.py, test_interpolation_factory.py,
test_parallel_geometry.py and test_parallel_interpolation.py now compare the full
fields again, including the boundary rows, and pass against the v06 references.


model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -565,37 +565,29 @@ def _register_normals_and_tangents_icosahedron(self) -> None:
         )
         self.register_provider(normal_vert_wrapper)
 
-        normal_cell = factory.ProgramFieldProvider(
-            func=stencils.compute_zonal_and_meridional_component_of_edge_field_at_cell_center,
+        # Computed as ndarray from edge row 1 with the missing-neighbor guards,
+        # like ICON does (mo_intp_coeffs: complete_patchinfo); a GT4Py stencil
+        # cannot skip the invalid E2C neighbor of the boundary edges.
```
nfarabullini:
```suggestion
```

model/common/src/icon4py/model/common/grid/geometry.py
```diff
@@ -637,37 +629,27 @@ def _register_normals_and_tangents_icosahedron(self) -> None:
...
-                "v_cell_2": "v_cell_2",
-            },
-            domain={
-                dims.EdgeDim: (
-                    self._edge_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_2),
-                    self._edge_domain(h_grid.Zone.LOCAL),
+            connectivities={"e2c": dims.E2CDim},
+            params={
+                "horizontal_start": self.grid.start_index(
+                    self._edge_domain(h_grid.Zone.LATERAL_BOUNDARY)
```
nfarabullini, line 644: this is a different boundary from the previous version. Was the original one wrong?

model/common/src/icon4py/model/common/interpolation/interpolation_fields.py
```diff
@@ -1086,49 +1086,56 @@ def compute_pos_on_tplane_e_x_y(
     llb = horizontal_start
     pos_on_tplane_e_x = array_ns.zeros(e2c.shape)
     pos_on_tplane_e_y = array_ns.zeros(e2c.shape)
+    # Row-1 boundary edges have only one cell neighbor; ICON skips the missing one
+    # and leaves its entry at zero (mo_intp_coeffs: calculate_tangent_plane_at_edge).
+    # The missing neighbor is either MISSING or a duplicate of the first one,
+    # depending on how the connectivity was constructed.
```
```suggestion
```
code is self-explanatory

nfarabullini (commented):
just a few minor comments

model/common/src/icon4py/model/common/grid/geometry.py
yiluchen1066: Yes! the previous start at ```LATERAL_BOUNDARY_LEVEL_2``` was a deviation from ICON. ICON computes ```primal_normal_cell/dual_normal_cell``` starting from the outermost edge row, skipping only the missing cell neighbor of each neighbor of each boundary edge. icon4py started one row later, so on limited-area grids the first row stayed zero. This went unnoticed as long as the v05 serialized references also had zeros there; the v06 references carry ICON's true row-1 values, which is how the deviation surfaced (see the PR description).

nfarabullini (approved):

PR #1411, opened 2026-07-31
Gate published test data on the records it was blessed with

jcanton:
from a friend:

Stacked on #1409 — review that first.

Part 1 makes a regeneration explain itself. This makes it hard to publish something that quietly disagrees with what CI already validated against.

- The archive helpers move from `scripts/python/serdata.py` into `icon4py.model.testing.serialized_data`, because a datatest has to import them and `scripts/` is not importable from one (nor does it exist in an installed wheel). `serdata.py` becomes CLI-only, which is also what @nfarabullini asked for on #1409.
- `serdata bless` pins the static and initial-state records of an archive into `model/testing/testdata/<experiment>.lock.json`, and a datatest recomputes them from the downloaded archive. The trajectory is not pinned, and records the lock does not know about are ignored, so adding instrumentation is not a violation.
- Every datatest session now prints the ICON revision behind each archive it used, so a failure says which Fortran build produced the reference it disagrees with. Warning only.
- `run-serialization --run-tests` runs the datatests against the freshly generated tree before anything is published.

Lockfiles for the six experiments at their current versions are included.

Not included: `known_hash` on download. It needs a sha256 per published tarball, and with only a handful available locally the registry would be mostly empty and the check mostly a no-op; the lockfile covers the same failure at content level.


jcanton:
from a friend:

Superseded — see the discussion on #1409. The whole fingerprint/diff/lockfile approach is being dropped in favour of recording the ICON revision alone.

Two measurements decided it:

- On the real v05 → v06 MCH pair the differ names 7 changed static records. 6 of them are named by the 3 failing datatests in their own function names (`test_primal_normal_cell`, `test_dual_normal_cell`, `test_compute_pos_on_tplane_e`), and the 7th (`v_edge_orientation`) is nproma padding, i.e. a false positive. Marginal information over the tests that fail anyway: negative.
- The ICON revision is already inside every archive ever published — 17/17 archives on hand carry exactly one `LOG.*.o` with exactly one ` revision:` line. And a `git describe` string is a valid git rev, so the whole diagnosis is one command with no tooling at all:

```
$ git log --oneline --no-merges <old-describe>..<new-describe> -- src/shr_horizontal/mo_intp_coeffs.f90
a9435ae531 [nwp] Improve wave energy propagation near the coast
```

Also worth recording: the terminal-summary provenance line added here prints nothing under `pytest-xdist` (`_SEEN_ARCHIVES` is filled in the workers, `pytest_terminal_summary` runs on the controller), so it was absent in exactly the CI runs it was meant for. Fixed in the reworked #1409.

PR #1412, opened 2026-08-03
remove config duplicates & share option values

DropD:
- remove extraneous config options, reducing possibility for expressing inconsistent configs
- allow expressing config options where values are shared between components only in some cases
  - makes it easier to express consistent config
  - eliminating inconsistent configs altogether seems impractical in these cases
  - current usecase: sharing between parametrizations for initial conditions and topography
  
Example for the option value sharing:
  
```yaml
shared:
  - u0: 35.0
    eta_0: 0.252
    consumers:
      - initial_condition
      - topography
...
initial_condition:
  type: jablonowski_williamson
  ...
topography:
  type: jablonowski_williamson
...
```


DropD:
cscs-ci run default;MODEL_SUBPACKAGES=common:standalone_driver;SESSIONS=model;LEVELS=integration;GRIDS=simple;BACKENDS=gtfn_cpu

bindings/src/icon4py/bindings/diffusion_wrapper.py
```diff
@@ -255,6 +253,8 @@ def diffusion_init(  # noqa: PLR0917 [too-many-positional-arguments]
             cell_params=grid_wrapper.grid_state.cell_geometry,
             backend=actual_backend,
             exchange=grid_wrapper.grid_state.exchange_runtime,
+            substep_as_float=float(ndyn_substeps),
```
jcanton: not you, but while you're at it I would rename this to
`ndyn_substeps_as_float`

alternatively: why not just cast to float where used?

jcanton (approved):
I think I like it a lot, just one minor (non-blocking) rename/cast comment
possibly @OngChia also should take a look?

model/common/src/icon4py/model/common/boundaries.py
```diff
@@ -0,0 +1,69 @@
...
+import dataclasses
+import typing
+
+from icon4py.model.common.config import config_io
+from icon4py.model.common.initial_condition import config as ic_config
+from icon4py.model.common.topography import config as topo_config
+
+
+@dataclasses.dataclass
+class BoundariesConfig:
```
OngChia: As initial_condition is not restricted to boundary. I suggest a few options:
1. `ModelSetupConfig` (but we already have `ExperimentConfig`, it is may be bit confusing as these two names are similar)
2. `DomainSetupConfig`
3. `InitialSetupConfig`
I prefer option 2 because I feel that both initial condition and topography are subsets of numerical domain setup. If you think 1 or 3 is better with a reason (or you have other better names), I am also fine.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
```diff
@@ -511,6 +487,8 @@ def __init__(
         | model_backends.BackendDescriptor
         | None,
         exchange: decomposition.ExchangeRuntime,
+        substep_as_float: float,
```
OngChia: I like the removal of these duplicated parameters from the config in diffusion and solve_nonhydro. I think corresponding changes need to be made in `diffuion_wrapper.py` that is used in blueline.
Similarly for the removal of `max_nudging_coefficient` from `NonHydrostaticConfig`. `dycore_wrapper.py` needs to be modified too.

OngChia (commented):
Thanks for the refactoring! It looks geat so far.

model/common/src/icon4py/model/common/boundaries.py
DropD: I like "domain" but "config" and "setup" are often used interchangeably. Is there a reason not to just call it `DomainConfig`?

OngChia: You are right. `Config` and `Setup` together sound redundant. Among the options so far, `DomainConfig` does fit better.

model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
DropD: I thought I had caught all the modifications necessary to the wrappers. Can you see any I missed?

OngChia: Ah, that's true. I overlooked the changes you made in the wrappers.

bindings/src/icon4py/bindings/diffusion_wrapper.py
DropD: I like the idea that the caller does not have to do the conversion.

In some functions / methods the converted value is used multiple times, in those cases I created a temporary and named it `ndyn_substeps_as_float`.

model/common/src/icon4py/model/common/boundaries.py
jcanton: initial condition is a **temporal boundary**!! so it was fine like it was, even though unusual maybe
domain is confusing because it usually refers to the whole internal thing, you would never call lateral boundary conditions "domain"...
 I do not like it and preferred boundaries

jcanton: - Boundary and initial conditions (BICs) : probably the clearest and most widely understood when you specifically mean initial conditions + boundary conditions.
- Auxiliary conditions : sometimes used collectively for initial and boundary conditions, especially in applied mathematics and PDE texts.
- Data or prescribed data : e.g. “the PDE together with the initial and boundary data,” particularly in a mathematical-analysis context.

but definitely **not** domain

jcanton (changes requested):
domain -> something else

model/common/src/icon4py/model/common/boundaries.py
jcanton: for an initial-boundary value problem (IBVP), the initial condition plus the boundary conditions together are usually just called "auxiliary conditions" or "side conditions" (data that, together with the PDE, make the problem well-posed). Some texts also just say "the data of the problem."

jcanton: `InitialAndBoundaryConfig`

we decided with @OngChia after nice discussion :-)
can you please update to this?
sorry about it

OngChia: Blame me for this extra comment. I just remember that why I had the feeling of a generic name such as DomainConfig is nice. Just to confirm one thing: isn't it that we do not set any boundary conditions in the topography setup? Or is it that defining the shape or location of the boundary is a subset of boundary conditions? The lower boundary is always free-slip. And we can't do anything with the lateral and upper boundary. With the Boundary, it sounds like we are setting some conditions on the boundary. I was thinking last week that we construct the geometry of the computational domain based on the topography. The boundary is set or defined somewhere else (in the future we can also set the boundary condition here such as surface or lateral boundary conditions if necessary). And, I could not think of a good definition the initial condition with geometry (or whatever topological space that is open set on which we solve the PDE). So, something like domain or setup is nice to include all these (computational domain's initial condition, domain's shape, domain's boundary condition, ...)
So, since topography is such a common word in climate community, should we give it with a more specific name such as `InitConAndTopographyConfig` for the time being?

jcanton: I think boundary is still fine and topography is too specific and restrictive (this will later include other static fields such as land mask and similar, as well as actual boundary condition config parameters such as if you have an inflow speed in a LAM setup).
topography effectively specifies the domain shape/extents/extremes where you perform the integration, where you apply boundary conditions, so it defines the boundary
we called it Initial and Boundary Config, not InitialConditions and BoundaryConditions Config

OngChia: okay!👍 No objection. Although personally it always makes me automatically fill in the word conditions when I see words `initial` and `boundary` in the config's name, that is just my issue.

DropD:
> domain -> something else

Instead, as discussed:
- move `initial_condition` and `topography` back to top-level
- add `shared` section to top-level with explicit consumer addressing
  - keep the logic for that generic enough to use it at different levels later (IOConfig)
- try to keep `experiment.initial_condition` instead of `experiment.initial_condition.config`

model/common/src/icon4py/model/common/boundaries.py
DropD: We decided to go with a more generic way to describe a parameter used in multiple places.

DropD:
cscs-ci run default;MODEL_SUBPACKAGES=common:driver;SESSIONS=model;LEVELS=integration;GRIDS=simple;BACKENDS=gtfn_cpu

model/common/src/icon4py/model/common/initial_condition/config.py
OngChia:
```suggestion
        "weisman_klemp": wk_ic.WeismanKlempConfig,
```

DropD: ✅

jcanton (approved):
nice job!

DropD:
cscs-ci run default;MODEL_SUBPACKAGES=common:driver;SESSIONS=model;LEVELS=integration;GRIDS=simple;BACKENDS=gtfn_cpu

PR #970, opened 2025-12-05
Add single precision option to the dycore (intermediate state)

starkphi:
Adds the option to set wp/vpfloat to single via the env var `ICON4PY_FLOAT_PRECISION` (previous options were 'double' and 'mixed') to run in single precision.

This PR was created to later reduce merge complexity of #886.

**Changes**

- Replace native python float types by either `wpfloat` or `vpfloat`
    - Change type annotations where there were native Python `floats` were present (which are by default passed as gtx.float64 to `gtx.field_operator`s and `gtx.program`s) and add casts around float literals
    - Swap `wpfloat` <-> `vpfloat` where an inconsistency with an operator was caught (to hopefully make fixing mixed precision easier in the future)
    - Replace `DBL_EPS` by `WP_EPS` and `VP_EPS`
 - Setup pytests for single precision
    - Add pytest marker `single_precision_ready` (running a test with `ICON4PY_FLOAT_PRECISION=single` deselects all tests without) and mark integration tests for the standalone driver and most components (required much larger tolerances for some cases)
    - Remove pytest option `--enable-mixed-precision` (`wpfloat` or `vpfloat` are already set according to `ICON4PY_FLOAT_PRECISION` at import time of a pytest file)
    - Added `PRECISION_VARIANTS` as dimension to the ci pipeline matrix
- Keep calculations and results in `gtx.float64` internally in geometry, metrics or interpolation factories independently of the selected precision. Mainly to keep the precision of the RBF and other geometry factors acceptable when running in `ICON4PY_FLOAT_PRECISION`
    - `factory.store_allfloats_as_double` forces float allocations in a `FieldSource` to double-precision
    - `FieldSource.export_field` casts to the dtype declared in the field metadata where the fields leave the factories

PR #1415, opened 2026-08-05
Hash CI_JOB_NAME for gt4py cache directory

msimberg:
CI_JOB_NAME_SLUG is truncated and can lead to cache directory conflicts if truncated at the wrong place. Instead use the full CI_JOB_NAME and hash it to guarantee the full name is used.

Discovered in #970 by @starkphi. Separating the change from #970 since it's a good change on its own.


msimberg:
cscs-ci run default

edopao (approved):
Good catch!

PR #1416, opened 2026-08-05
Update merge pipeline to run more tests

msimberg:
`SESSIONS` was already effectively using `all`, just explicitly spelled out, so that is not a functional change.

`MODEL_SUBSETS` and `MODEL_MPI_SUBSETS` being updated to `all` means that non-datatests now get run on merges as well. These should generally be the cheapest tests but are not entirely free, so let's see if we end up using way too much CI resources with this.

### DIFF
--- a/ci/merge.yml
+++ b/ci/merge.yml
@@ -4,11 +4,11 @@ include:
 variables:
   # See default.yml and generate_ci_pipeline.py for details on how these are
   # used. This pipeline gates merges and variables should not be overridden.
-  SESSIONS: "model:model_mpi:tools"
+  SESSIONS: "all"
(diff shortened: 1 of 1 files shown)


msimberg:
cscs-ci run default;MODEL_SUBSETS=all;MODEL_MPI_SUBSETS=all

msimberg:
As an example, on the default pipeline I triggered above, we're now getting seven more CI jobs per backend compared to before because not all packages have non-datatests (https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504669/-/pipelines/2734082717; 24 vs 17).

kotsaloscv (approved):

PR #1417, opened 2026-08-05
Only test python 3.12 in github workflows

msimberg:
Reduce the matrix since testing all three python versions is relatively expensive. We're already testing python 3.13 in CSCS CI so there's no need to test that again in github workflows. Removing 3.14 is arguable, but we're currently not recommending it for production, so let's add it back only once we want to recommend it.

### DIFF
--- a/.github/workflows/icon4py-test-model.yml
+++ b/.github/workflows/icon4py-test-model.yml
@@ -22,7 +22,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


havogt (approved):
lgtm, in recent gt4py community meeting the other groups also mentioned problems in python 3.14 anyway and suggest to skip it for production.

msimberg:
I'll skip the merge queue since this doesn't touch CSCS CI any real code, once the github workflows have finished and the merge queue is empty.

edopao (approved):

PR #1422, opened 2026-08-07
refactor(physics_driver): remove unused ForcingMode

yiluchen1066:
`ForcingMode` was declared as the AES `fc_xxx` analogue in #1301 but never used:
only `APPLY` was implemented and the driver raised `NotImplementedError` for
`DIAGNOSTIC`. Nothing in the codebase sets or reads the field.

We agreed offline that the declared-but-unimplemented switch is confusing; removing it
until a concrete compute-without-applying use case exists. If one arrives (e.g.
in-loop validation of a future component against a fixed trajectory), it should come
back together with an implementation — which first needs `scatter_to_prognostic`
split into apply-tendencies vs store-diagnostics.


msimberg (approved):
No complaints, very nice. Thank you!

PR #1423, opened 2026-08-07
Name the fields compared in the muphys datatest

jcanton:
from a friend:

Running the muphys datatest with `ICON4PY_DALLCLOSE_PRINT_INSTEAD_OF_FAIL=1` printed eighteen unattributed rows of max abs/rel diff, so there was no way to tell which field each belonged to — which is exactly what you need to tighten the provisional tolerances.

`assert_dallclose` already renders `err_msg`; the eight call sites just never passed one. Now:

```
... tend_qc in cloud             max abs diff 0.0                     max rel diff 0.0
... tend_temperature in cloud    max abs diff 1.8947806286936006e-16  max rel diff inf
... rsfl (rain)                  max abs diff 0.0                     max rel diff 0.0
... ssfl (ice + snow + graupel)  max abs diff 0.0                     max rel diff 0.0
```

The two assertions per tracer are labelled separately because they check different things — the second one verifies the granule produces near-zero tendencies above the cloudy region, where ICON does not run the scheme.


OngChia (approved):
LGTM

PR #1424, opened 2026-08-07
docs: shared code placement and generic naming

jcanton:
from a friend:

Generic operations keep getting implemented inside components under names taken from the caller's variables, so the next component does not find them and ports the same Fortran expression again. `dycore/compute_tangential_wind` and `tracer_advection/compute_edge_tangential` are the same `neighbor_sum` over `E2C2E`; `copy_cell_kdim_field_to_vp` and `copy_cell_kdim_field` are the same identity copy; `diffusion/calculate_nabla2_for_z` is named after the Fortran temporary `z_nabla2_e`.

- `CODING_GUIDELINES.md`: new "Shared code and generic naming" section with a test for whether something belongs in `model/common`, a naming rule with the forbidden categories, and before/after examples.
- `docs/stencil_naming_convention.md`: turn the existing "do not use signature variable names" hint into a rule and cross-reference the guidelines.
- `AGENTS.md`: pointer to the new section.

Docs only, no renames. Existing offenders are meant to be fixed opportunistically, one move per PR.

### DIFF
--- a/CODING_GUIDELINES.md
+++ b/CODING_GUIDELINES.md
@@ -35,22 +35,12 @@ then the code is generic and belongs in `model/common`. If the sentence needs a 
 
 For example `tracer_advection/compute_tendency` computes `(new - old) / dtime`. Nothing in that sentence is a tracer, so the stencil belongs in `model/common`.
 
-This does not contradict the YAGNI item above: do not _generalise_ code speculatively so that it might be shared one day. But code that is _already_ generic should be placed and named generically from the start, because moving it later also means moving its tests and coordinating stencil renames with [icon-exclaim](https://github.com/C2SM/icon-exclaim).
+This does not contradict the YAGNI item above: do not _generalise_ code speculatively so that it might be shared one day. But code that is _already_ generic should be placed and named generically from the start, because moving it later also means moving its tests and coordinating stencil renames.
 
-Where to put it inside `model/common`:
-
-| Content                                                      | Location                   |
-| ------------------------------------------------------------ | -------------------------- |
-| differential operators and algebra on fields                 | `math/`                    |
-| cell/edge/vertex and full/half level interpolation           | `interpolation/`           |
-| allocation, copies, index manipulation                       | `utils/`                   |
-| thermodynamic and other diagnostic quantities                | `diagnostic_calculations/` |
-| physical constants                                           | `constants.py`             |
-| configuration enums and options shared by several components | `model_options.py`         |
 
 ### Naming
 
-A generic name states **what the operation does** and **which grid entities it acts on**. It must not encode:
+A generic name states **what the operation does** and **which grid entities it acts on**. It must **not** encode:
 
 - the caller's argument or output variable names: `diffusion/calculate_nabla2_for_z` is named after the Fortran temporary `z_nabla2_e`, while the stencil computes a coefficient-weighted normal gradient of a cell field onto edges;
 - Fortran temporary prefixes (`z_`, `p_`, `opt_`) or Fortran module names (`mo_intp_rbf_rbf_vec_interpol_cell`);
@@ -61,19 +51,7 @@ Use the swap test: replace an input with an unrelated field of the same type. If
 
 Generic does not mean meaningless: name the operation, not placeholders. `add_fields` is a generic name, `compute_a_plus_b` is not.
 
-| Too specific                                                         | Generic                                      |
-| -------------------------------------------------------------------- | -------------------------------------------- |
-| `compute_tangential_wind`, `compute_edge_tangential`                 | `interpolate_normal_to_tangential_on_edges`  |
-| `calculate_nabla2_for_z`                                             | `compute_weighted_normal_gradient_on_edges`  |
-| `copy_cell_kdim_field_to_vp`                                         | `copy_cell_field`                            |
-| `init_cell_kdim_field_with_zero_wp`, `init_constant_cell_kdim_field` | `set_cell_field_to_constant`                 |
-| `mo_intp_rbf_rbf_vec_interpol_cell`                                  | `interpolate_edge_vector_to_cell_center_rbf` |
-
 The same applies outside stencils: shared enums, options and dictionaries are named after what they configure, not after the component that introduced them.
-
-### Existing code
-
-Do not open repository-wide renaming pull requests. Rename and move opportunistically, while you are already modifying a file, and keep one move per pull request so that the diff stays reviewable. Names under `stencil_tests/` are consumed by [icon-exclaim](https://github.com/C2SM/icon-exclaim), so coordinate before renaming them.
 
 ## Code Style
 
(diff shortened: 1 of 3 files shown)


CODING_GUIDELINES.md
```diff
@@ -20,6 +20,39 @@ TODO: we should add a doc folder (separate for `tools` and `model`? separate for
...
+### Naming
+
+A generic name states **what the operation does** and **which grid entities it acts on**. It must **not** encode:
+
+- the caller's argument or output variable names: `diffusion/calculate_nabla2_for_z` is named after the Fortran temporary `z_nabla2_e`, while the stencil computes a coefficient-weighted normal gradient of a cell field onto edges;
+- Fortran temporary prefixes (`z_`, `p_`, `opt_`) or Fortran module names (`mo_intp_rbf_rbf_vec_interpol_cell`);
+- the floating point precision (`_wp`, `_vp`), which is already part of the signature;
+- the physical meaning of an operand that the mathematics does not depend on.
+
+Use the swap test: replace an input with an unrelated field of the same type. If the code still works but the name no longer reads correctly, the name is too specific.
```
msimberg: This is more or less duplicated from stencil_naming_convention.md, right?

docs/stencil_naming_convention.md
```diff
@@ -19,7 +19,19 @@ The `program` name should start with a verb.
 Commonly used verbs are accumulate, add, apply, compute, copy, correct, extrapolate, interpolate, return, set, and
 solve.
 The verb can be followed by more describing words.
-If possible one should not use the names of variables that are part of the signature already as describing words.
+
+The describing words must state what the stencil computes and on which grid entities, and must not repeat the names of the variables in the signature.
+In particular, a stencil name must not contain:
+
```
havogt, line 26: I totally agree with the idea, but I see clashes which might confuse agents more than humans because the instructions are very specific and don't leave much room for exceptions, see following comments.

docs/stencil_naming_convention.md
```diff
@@ -19,7 +19,19 @@ The `program` name should start with a verb.
...
+- the caller's argument or output variable names, for example `calculate_nabla2_for_z`, named after the Fortran temporary `z_nabla2_e`;
```
havogt: if the computations is specific to the quantity that you compute, it might actually make sense (this one is probably the weakest of my comments)

docs/stencil_naming_convention.md
```diff
@@ -19,7 +19,19 @@ The `program` name should start with a verb.
...
+- Fortran temporary prefixes (`z_`, `p_`, `opt_`) or Fortran module names, for example `mo_intp_rbf_rbf_vec_interpol_cell`;
+- the floating point precision (`_wp`, `_vp`), which is already part of the signature;
```
havogt: in the future yes, but we need dtype generics in gt4py, because there might be the same function for both precisions, so probably today the recommendation for `common` operators should we to append the suffix always.

docs/stencil_naming_convention.md
```diff
@@ -19,7 +19,19 @@ The `program` name should start with a verb.
...
+- the physical meaning of an operand that the mathematics does not depend on.
+
+To check a name, replace an input with an unrelated field of the same type: if the stencil still works but the name no longer reads correctly, the name is too specific.
+
+A stencil whose name passes this check is usually generic enough to belong in `model/common`.
```
havogt: Since all of the above criteria don't allow exceptions, this trivially applies to all operators, no?

DropD: I was wondering the same. This stands in contrast with the section "Does it belong in `model/common`" in CODING_GUIDELINES.

CODING_GUIDELINES.md
```diff
@@ -20,6 +20,39 @@ TODO: we should add a doc folder (separate for `tools` and `model`? separate for
...
+
+Generic does not mean meaningless: name the operation, not placeholders. `add_fields` is a generic name, `compute_a_plus_b` is not.
```
DropD: While we are here, do we want to consider advising against using verbs that are completely redundant for field operators / programs, like "compute", "calculate", "process"?

For example, what information does `compute_a_plus_b` add over `a_plus_b`?

Or maybe there are some field operators / programs that do not compute or calculate anything and process some fields to that end. In which case of course this would be a bad idea.

msimberg: I think these names come from magdalena. I don't love them either, but there is a logic to them. There are functions that are e.g. `update_*` or `apply_*`. in `stencil_naming_conventions.md`:

> Commonly used verbs are accumulate, add, apply, compute, copy, correct, extrapolate, interpolate, return, set, and solve.

I think the idea is that if you're using copy, extrapolate, solve, etc. then you want a verb for the ones that "just compute", where I guess "compute" is the catchall for anything that doesn't have a more specific verb.

CODING_GUIDELINES.md
jcanton: Fixed: the "Naming" section in CODING_GUIDELINES.md is now just a link to stencil_naming_convention.md. The only remaining content is the non-stencil rule (enums/options/dictionaries), which is not covered by the stencil doc.

docs/stencil_naming_convention.md
jcanton: Thanks! All three clashes are addressed in ef1677002, see the replies below: the quantity-specific exception on the caller-variable bullet, the precision suffix as an explicit exception, and the separation of naming rules from model/common placement.

docs/stencil_naming_convention.md
jcanton: Agreed, softened: the bullet now allows naming after the quantity when the mathematics genuinely depends on it, which is exactly the case where the swap test does not pass.

docs/stencil_naming_convention.md
jcanton: Fixed: the precision suffix is now an explicit exception and is required today when an operator exists in both precisions (e.g. interpolate_edge_field_to_half_levels_vp/_wp, as already done in model/common), to be dropped once GT4Py has dtype generics.

docs/stencil_naming_convention.md
jcanton: Fixed: replaced with "These naming rules apply to every stencil, wherever it lives. Whether a stencil belongs in model/common is a separate question, answered by the section on shared code and generic naming in CODING_GUIDELINES.md."

and in the same way, naming and placement are now explicitly separated, so the "usually generic enough" claim is gone and there is no longer a contrast with the "Does it belong in model/common" section.

jcanton: Fixed in the same way: naming and placement are now explicitly separated, so the "usually generic enough" claim is gone and there is no longer a contrast with the "Does it belong in model/common" section.

CODING_GUIDELINES.md
jcanton: I kept the verb as part of the convention (program names must start with a verb as @msimberg pointed out from magda's doc), but added "prefer the most specific verb that describes the operation; compute is the catch-all for operations that have no more specific verb". The "name the operation, not placeholders" example (add_fields vs compute_a_plus_b) now lives in the stencil doc.

jcanton: Exactly what we went with: "Prefer the most specific verb that describes the operation; compute is the catch-all for operations that have no more specific verb."

msimberg (approved):
Very nice! (Note that this can skip the merge queue)

PR #1403, opened 2026-07-28
Distributed output: zarr and parallel netCDF backends, gather/rank-block write modes

kotsaloscv:
## What

Output now works in distributed (MPI) runs. Until now the driver disabled IO with a
warning as soon as `comm_size > 1`.

### Output distributions (`common/io/distributed.py`)

New strategy layer between the field-group monitor (scheduling) and the writers (format):

- **`GatherDistribution`** (default): halo entries dropped via the owner masks, owned
  entries `Gatherv`-ed to rank 0 and placed at their global indices — output identical to
  a single-rank run, works with both backends, but root-memory-bound.
- **`RankBlockDistribution`**: no data communication; every rank writes its own owned
  entries into a rank-contiguous block of a shared store. Horizontal axes are padded
  to a uniform block size; `global_index_<dim>` coordinates recover the global order.
- **`SingleNodeDistribution`**: passthrough for single-rank runs.

### Zarr writer (`common/io/zarr_writers.py`)

`ZarrWriter` (zarr format 3) next to `NETCDFWriter`, both behind a `FieldWriter` protocol
and sharing all coordinate/CF attribute definitions, so a store reads like the equivalent
netCDF file. Rank-block stores are chunked so that no chunk (or shard) crosses a
rank-block boundary — one chunk per rank block by default — so concurrent writes never
touch the same chunk file; the root rank owns all store-metadata operations, ordered by
one barrier per append.

### Parallel netCDF (`common/io/netcdf_writers.py`)

`NETCDFWriter` supports the rank-block mode too: on an MPI-parallel netCDF4
installation, `distributed` + `netcdf` writes one shared file (`parallel=True` on the
monitor's communicator) with the exact layout of the rank-block zarr store — padded
per-rank blocks, HDF5 chunks never crossing block boundaries, `global_index_<dim>`
coordinates — so reassembly and any downstream consumer work on either format.
Metadata operations run collectively on every rank; variables touching the unlimited
time dimension use collective access, with every rank writing its full block (padding
included) so ranks owning no entries still participate in every collective write.
Vertical chunk sizes shrink automatically to respect HDF5's 4 GiB chunk limit.

The PyPI wheels of `netCDF4` are serial builds (`__has_parallel4_support__ == 0`), so
the combination is rejected at configuration time on such installations — the error and
the `common.io` module docs spell out the exact steps to an MPI-parallel build
(MPI-enabled netCDF-C/HDF5, then `pip install --no-binary netcdf4 --no-build-isolation
[...]

PR #1426, opened 2026-08-10
Use `data_alloc.as_numpy` on global halo indices before passing to GHEX to satisfy nanobind 2.14's stricter conversion

msimberg:
With nanobind 2.13 or older the elements of the index list would get converted element-wise using `PyNumber_Long`, which in practice means `__int__`. numpy and cupy arrays would get converted as sequences, and each element of the arrays would get converted through `__int__`. Starting with nanobind 2.14 the conversion of elements now happens through `PyNumber_Index`, i.e. `__index__`. See
https://github.com/wjakob/nanobind/blob/master/docs/changelog.rst#version-2140-aug-7-2026 and https://github.com/wjakob/nanobind/commit/d33465e4a2d1f48e6210dbe6a7c44f7bf2b1fae2 for details. cupy arrays for some reason do not have `__index__` and the conversion fails.

This PR explicitly converts the arrays to numpy arrays because 1. GHEX anyway expects a host array and 2. numpy array elements have `__index__`. This will work with nanobind 2.14 and older. I've also changed two previous `.tolist`s passed to GHEX (for `DomainDescriptor`) that also expect `std::vector<int>` like `HaloGenerator`.

This came up on https://github.com/C2SM/icon4py/pull/1283 not because it in itself changed anything related to halo generators or similar. It only came up on #1283 because uv.lock was changed, the base image was rebuilt, and nanobind 2.14 had just been released (August 7th), _and_ GHEX gets built against the latest nanobind release even though nanobind is pinned to 2.12 for the actual venv used by icon4py.

### DIFF
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -50,8 +50,6 @@ variables:
     reports:
       dotenv: build.env
   variables:
-    # TODO: Remove
-    CSCS_REBUILD_POLICY: "always"
     DOCKERFILE: ci/docker/venv.Dockerfile
     DOCKER_BUILD_ARGS: '["PYVERSION=$PYVERSION", "BASE_IMAGE=$BASE_IMAGE", "ICON4PY_NOX_UV_CUSTOM_SESSION_EXTRAS=$ICON4PY_NOX_UV_CUSTOM_SESSION_EXTRAS"]'
 
--- a/ci/base.yml
+++ b/ci/base.yml
@@ -14,8 +14,6 @@ variables:
   PYVERSION: 3.13.13
   PYVERSION_SHORT: 3.13
   ICON4PY_NOX_UV_CUSTOM_SESSION_EXTRAS: "cuda12"
-  # TODO: Remove
-  CSCS_REBUILD_POLICY: "always"
 
 # Base image build step with SHA256 checksum for caching
 .build_baseimage:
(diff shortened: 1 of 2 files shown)


msimberg:
cscs-ci run default

kotsaloscv (approved):
Looks good to me ;)

msimberg:
cscs-ci run default;BACKENDS=gtfn_cpu:gtfn_gpu

msimberg:
@kotsaloscv I changed this to actually use the same `as_numpy` that you had in #1403. It avoids the conversion to plain python ints that `tolist` did, and I change the other `tolist`s to use `as_numpy` sa well.

That said, GHEX should possibly be taking ndarray's instead of `std::vector<int>` to avoid the element-wise conversion in the first place, but that is for the future. This is only done when setting up GHEX patterns, not every halo exchange.

msimberg:
cscs-ci run default;BACKENDS=gtfn_cpu:gtfn_gpu

ci/base.yml
```diff
@@ -50,6 +50,8 @@ variables:
     reports:
       dotenv: build.env
   variables:
+    # TODO: Remove
+    CSCS_REBUILD_POLICY: "always"
```
msimberg:
```suggestion
```

egparedes (approved):
LGTM

PR #1427, opened 2026-08-10
Enable DaCe external workspace for AMD platform

edopao:
## Summary

Enable DaCe's external workspace feature for the AMD (ROCm) platform. The external workspace avoids the overhead of runtime allocations for transient SDFG arrays, which are expensive in the AMD runtime.

The workspace is sized via `workspace_size` only; alignment of the workspace slab is handled by the underlying allocator, so no separate alignment configuration is exposed.

### New module: `model/common/src/icon4py/model/common/backend_configuration.py`

- **`BackendConfig`** dataclass for workspace sizing (`workspace_size`), with validation in `__post_init__` (size must be positive).
- **`backend_config_from_env()`** reads `ICON4PY_BACKEND_WORKSPACE_SIZE` from the environment; returns `None` when it is not set.
- **`IconWorkspaceAllocator`** singleton that caches a single workspace slab per device (via `_get_slab`) and reuses it across every compiled program. On cache hit, the slab size is validated.
- **`ICON_WORKSPACE_ALLOCATOR`** module-level instance consumed by `get_dace_options`.

### `model/common/src/icon4py/model/common/model_backends.py`

- Added `external_workspace` parameter to `make_custom_dace_backend`. When provided, `transient_memory_mode` is forced to `EXTERNAL` (or validated if already set).

### `model/common/src/icon4py/model/common/model_options.py`

- `get_dace_options` now takes a `backend_config` parameter; when set, it allocates the external workspace via `ICON_WORKSPACE_ALLOCATOR.allocate(...)` and sets `transient_memory_mode = EXTERNAL`.
- `get_options` passes `backend_config` through as a keyword-only argument.
- `customize_backend` accepts `backend_config` and falls back to `backend_config_from_env()` when `None`.
- `setup_program` accepts a `backend_config` parameter.

### `model/driver/src/icon4py/model/driver/config.py`

- Added `backend_config` field to `DriverConfig` (defaults to `backend_config_from_env()`).

### `model/driver/src/icon4py/model/driver/main.py`

- Moved `customize_backend` / `get_allocator` to after `config = config.with_overrides(...)` so the driver config's `backend_config` is honored.

### `model/driver/tests/driver/unit_tests/data/test_config.yml`

- Added `backend_config:` (empty/None) to the test config.

### `pyproject.toml`

- Reordered commented-out uv source lines (no functional change).

## Configuration

The external workspace is configured via an environment variable:

| Variable                         | Description                          | Default |
| -------------------------------- | ------------------------------------ | ------- |
| `ICON4PY_BACKEND_WORKSPACE_SIZE` | Workspace size in bytes, per device. | -       |

When `ICON4PY_BACKEND_WORKSPACE_SIZE` is not set, no external workspace is allocated.

## Backward Compatibility

- The `external_workspace` parameter defaults to `None` in `make_custom_dace_backend`.
- The `backend_config` parameter defaults to `None` in `customize_backend` / `setup_program`.
- Existing callers that do not pass `backend_config` are unaffected; the default behavior (no external workspace) is preserved unless `ICON4PY_BACKEND_WORKSPACE_SIZE` is set in the environment.

## Testing

Tests in `model/common/tests/common/test_backend_configuration.py` and `model/common/tests/common/test_model_options.py`.


edopao:
cscs-ci run dace

edopao:
cscs-ci run dace

model/common/src/icon4py/model/common/backend_configuration.py
```diff
@@ -0,0 +1,170 @@
...
+from types import ModuleType
+from typing import ClassVar, Final
+
+import gt4py.next as gtx
+from gt4py.next.program_processors.runners.dace.workflow import common as gtx_wfdcommon
+
+from icon4py.model.common.utils import data_allocation
+
+
+_DEFAULT_ALIGNMENT: Final[int] = 256  # Matches DaCe's default transient-storage alignment
```
msimberg: Does this have to match? If yes, can it be taken from dace instead?

model/common/src/icon4py/model/common/backend_configuration.py
```diff
@@ -0,0 +1,170 @@
...
+    is used.
+    """
+    size = os.environ.get("ICON4PY_BACKEND_WORKSPACE_SIZE")
+    if size is None:
+        return None
+    alignment = os.environ.get("ICON4PY_BACKEND_WORKSPACE_ALIGNMENT")
+    return BackendConfig(
+        workspace_size=int(size),
+        workspace_alignment=int(alignment) if alignment is not None else _DEFAULT_ALIGNMENT,
+    )
```
msimberg, line 77: Should a BackendConfig be something one can set in an experiment config? Or live purely as a parallel config world which is set differently? If they get merged, then this way of setting env vars to configs should probably be expressible with the config annotations, but I'm not sure that's necessary or a good idea. It would definitely be a new requirement for the rest of the config system.

model/common/src/icon4py/model/common/backend_configuration.py
```diff
@@ -0,0 +1,170 @@
...
+        try:
+            import cupy as cp  # type: ignore[import-not-found]  # noqa: PLC0415 [import-outside-top-level]
+        except ImportError as err:
+            raise RuntimeError(
+                f"GPU workspace requested but cupy is not available: {err!r}."
+            ) from err
+        return cp
+    import numpy as np  # noqa: PLC0415 [import-outside-top-level]
+
+    return np
```
msimberg: Should this probably not be in data_allocation?

msimberg: Also see #1434.

model/driver/tests/driver/unit_tests/data/test_config.yml
```diff
@@ -84,6 +84,7 @@ driver:
   ndyn_substeps: 5
   enable_statistics_logging: false
   enable_output: false
+  backend_config:
```
msimberg, line 87: Partly answering my own question about whether backend config should be/is in the experiment config. It is added here, and if it should stay then it should probably be aligned a bit better with the other sub-config dataclasses. I'd appreciate @DropD's input on what might make sense here.

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -96,7 +116,9 @@ def customize_backend(
     | model_backends.DeviceType
     | model_backends.BackendDescriptor
     | None,
+    backend_config: backend_cfg.BackendConfig | None = None,
```
msimberg, line 119: Depending on how important it is to remember to set the BackendConfig, would it make sense to make this non-optional? Or alternatively make the default a BackendConfig() so you don't have to deal with the None case later?

msimberg (commented):
Some quick comments. Would appreciate @DropD's comments on config changes.

model/common/src/icon4py/model/common/backend_configuration.py
edopao: No, it doesn't have to match. Besides, there is no requirement for the memory to be aligned, so I could also remove it entirely (@iomaganaris what do you think?). I should definitely remove the comment because it is misleading.

model/common/src/icon4py/model/common/backend_configuration.py
```diff
@@ -0,0 +1,170 @@
...
+
+from icon4py.model.common.utils import data_allocation
+
+
+_DEFAULT_ALIGNMENT: Final[int] = 256  # Matches DaCe's default transient-storage alignment
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class BackendConfig:
+    """External DaCe workspace sizing, configurable per experiment."""
```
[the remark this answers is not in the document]
edopao: I used `ExternalWorkspaceConfig` initially, but I changed it to the generic `BackendConfig` in order to be able to support other configuration options in the future.

model/common/src/icon4py/model/common/backend_configuration.py
edopao: My idea: the experiment config can be used in granule and driver tests and standalone runs, while the environment variables can be used in stencil test/benchmark.

model/common/src/icon4py/model/common/backend_configuration.py
edopao: Most of the functions in `data_allocation` require `allocator` or `grid`. On a first look I missed the function `data_allocation._array_ns()`. Yes, I can use that one , but I have 1. to make it public and 2. to make it throw an error if cupy is requested but not installed.

model/common/src/icon4py/model/common/model_options.py
edopao: For now, `BackendConfig` is only used to setup an external workspace, which is a very special case (a workaround for the AMD platform, indeed). I would propose to keep it optional for now. If we extend `BackendConfig`, it is a good idea to make it non-optional.

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -35,11 +35,26 @@ def _dace_remove_access_node_copies(sdfg: dace.SDFG) -> None:
...
-    is_rocm_device = backend_descriptor.get("device") == model_backends.DeviceType.ROCM
+    device = backend_descriptor.get("device")
     optimization_args = backend_descriptor.get("optimization_args", {})
     optimization_hooks = optimization_args.get("optimization_hooks", {})
+
+    if backend_config is not None:
+        # The workspace memory allows to avoid the overhead of runtime allocations,
+        # which are expensive in the AMD runtime.
+        backend_descriptor["external_workspace"] = backend_cfg.ICON_WORKSPACE_ALLOCATOR.allocate(
+            device,
```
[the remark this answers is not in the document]
edopao: Good idea, done.

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -35,11 +35,26 @@ def _dace_remove_access_node_copies(sdfg: dace.SDFG) -> None:
 
 
 def get_dace_options(
-    program_name: str, **backend_descriptor: Any
+    program_name: str,
+    backend_config: backend_cfg.BackendConfig | None,
+    **backend_descriptor: Any,
 ) -> model_backends.BackendDescriptor:
-    is_rocm_device = backend_descriptor.get("device") == model_backends.DeviceType.ROCM
+    device = backend_descriptor.get("device")
```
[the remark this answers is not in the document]
edopao: Done

model/common/tests/common/test_backend_configuration.py
```diff
@@ -0,0 +1,155 @@
...
+class TestArrayNamespaceFor:
+    def test_returns_numpy_for_cpu(self) -> None:
+        assert bc._array_namespace_for(CPU) is np
+
+    def test_raises_runtime_error_when_cupy_not_available(
+        self, monkeypatch: pytest.MonkeyPatch
+    ) -> None:
+        monkeypatch.setattr(bc.gtx, "CUPY_DEVICE_TYPE", bc.gtx.DeviceType.CUDA)
+        with pytest.raises(RuntimeError, match="cupy is not available"):
+            bc._array_namespace_for(bc.gtx.DeviceType.CUDA)
```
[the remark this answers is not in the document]
edopao: Good point, fixed.

model/driver/tests/driver/unit_tests/data/test_config.yml
edopao: Awaiting input from  @DropD

model/common/src/icon4py/model/common/backend_configuration.py
iomaganaris: I don't know what DaCe does but memory provided to the kernels should be at least 256-byte aligned. `cudaMalloc` provides that by default, `cupy` provides 512-byte aligned vectors and I expect `hipMalloc` to do the same. Having a default alignment of `256` is fine

edopao: If `cupy` by default provides memory aligned memory, we could avoid adding this extra alignment feature.

iomaganaris: `cupy` and the other `malloc`s provide an aligned memory buffer however all the subsequent pointers passed from this aligned buffer also need to be aligned with a certain byte amount. IIUC this variable is used for the alignment of the pointers passed to DaCe. Is that right? If yes, then this is necessary

edopao: Unfortunately not, this function only aligns the start address of the workspace memory buffer. The pointers of transient arrays, inside the workspace buffer, are generated by dace.

iomaganaris: I see now. Based on https://github.com/spcl/dace/blob/5f9a3eb0dbf2197e65e711b3715df046dd93b40b/dace/codegen/targets/framecode.py#L447 it seems like the pointers computed by DaCe are not guaranteed to be aligned in any way. It would be a good idea probably to pass some alignment size on the function and compute the offset based on that. What do you think? Shall I ask a "friend" to take care of that or would you like to try it out?

edopao: You would also need to change `__dace_get_external_memory_size_`, the function that tells the host side how large buffer to allocate.

edopao: With my latest commit, I have now removed the `alignment` argument because our discussion showed that it does not provide added value here. I can experiment on the side to enforce alignment in dace, in the generated code.

iomaganaris: Thanks a lot for looking into this 👍 This can be done in another PR anyway so there's no reason to block this because of that

iomaganaris (approved):
Looks good on my side. Very welcome changes 🚀 
Would be nice to have a rule of thumb regarding what is the required size to set for different grids but this can be defined empirically with time

model/driver/tests/driver/unit_tests/data/test_config.yml
DropD: I will need to take a closer look to see if it makes sense as is.

model/common/src/icon4py/model/common/backend_configuration.py
```diff
@@ -0,0 +1,109 @@
...
+
+from icon4py.model.common.utils import data_allocation
+
+
+@dataclasses.dataclass(frozen=True, kw_only=True)
+class BackendConfig:
+    """External DaCe workspace sizing, configurable per experiment."""
+
+    #: Workspace size in bytes, per device.
+    workspace_size: int
```
DropD: Please add the description in a `ConfigOption` annotation instead of a comment, so it shows up in the interactive documentation. Also maybe adjust it to specify that it's about the DaCe workspace and that it only affects performance, not scientific outcome (if that is true).

model/driver/tests/driver/unit_tests/data/test_config.yml
DropD: After looking it feels close to what we probably want. Possibly good enough for now.

model/driver/src/icon4py/model/driver/config.py
```diff
@@ -247,6 +248,17 @@ class DriverConfig:
             icon_equivalent=None,
         ),
     ] = False
+    backend_config: typing.Annotated[
```
DropD, line 251: I think this makes sense for now. Things to think about in the future (or if you feel like it, now as well):

- It might improve the user experience a lot to have options that tweak performance but not scientific outcomes clearly marked or even segregated. Perhaps we would like there to be a section that reads "performance" (under which "backend_config" could go, or which "backend_config" might be morphed into.
- The `None` default could be improved to a sentinel type that gets deserialized into "read from env" or something like that. Feels more transparent than just `backend_config:` followed by empty space.

edopao:
Thanks @DropD. I have addressed a couple of comments. I leave the rest to future PRs, if more performance configuration is needed.

PR #1307, opened 2026-06-11
feat(standalone_driver): single-node NetCDF/UGRID output via the common io module

kotsaloscv:
## What

Wires `icon4py.model.common.io` into the standalone driver for **single-node
NetCDF/UGRID** output of the full model state (phase 0; parallel output is a follow-up).
Built on the config-driven driver (after merging `main`).

## Output

Opt-in via `--enable-output` (or `DriverConfig.enable_output`; default off). Writes the
initial state and then on a schedule, into one file per segment
`<output_path>/icon4py_output_NNNN.nc` (+ a `*_ugrid.nc` mesh) — all 10 fields together,
by CF name:

- prognostic: `air_density`, `exner_function`, `virtual_potential_temperature`,
  `upward_air_velocity`, `normal_velocity`
- diagnostic (computed at output time): `eastward_wind`, `northward_wind`,
  `temperature`, `virtual_temperature`, `pressure`

Schedule: every N model steps (`output_interval_steps`, default 1) **or** a
simulation-time interval (`output_interval` + `start_time`, e.g. `"10 HOURS"`).

## How

- `driver_io.py` — state → CF/UGRID `DataArray`s (host-side, GPU-safe); metadata driven
  from `states/data.py` (CF names + `icon_var_name` + `is_on_half_levels`);
  `DiagnosticsComputer` (scratch/output buffers allocated once, reused per step).
- `standalone_driver.py` — monitor built in `initialize_driver` (single-node guard),
  per-step store in the time loop, `close()` in `finally`.
- `common/io` — step- or time-based scheduling; `grid_id` strictly `uuid.UUID`;
  `to_data_array(to_host=...)`.
- `pyproject.toml` — `icon4py-common[io]` is a required driver dependency.

## Testing

Unit tests (`test_driver_io.py`, `test_io.py`) + JW integration datatest
(`test_driver_io_output.py`). pre-commit, unit suites, and the datatest are green.

## Out of scope (follow-ups)

- Single-node only — parallel output (gather-to-rank-0 / Zarr) is the next PR.

PR #1430, opened 2026-08-13
Asynchronous zarr output: overlap the data writes with the model computation

kotsaloscv:
Builds on the distributed output of #1403 (now merged; this PR is rebased onto `main`).

When a field group writes asynchronously, `store` splits the work in two. Everything that involves communication or store metadata still runs on the calling thread, exactly as in the synchronous path: the distribution's `prepare` (gather / halo stripping), creating and resizing the arrays in the store, and the broadcast that rank-block mode uses to order those metadata operations before the data writes (`_root_step`). Only the last step — writing this rank's data chunks to disk — is handed to a background thread (one per group, `writers.AsyncWriteQueue`, a bounded FIFO queue), so the file write overlaps with the following model steps instead of blocking them.

**Design**

- The background thread performs no MPI and reads no store metadata: the target zarr arrays are resolved before queueing, and a later time-axis resize only grows an array's shape, never touching the chunks of earlier slices.
- Because all communication stays on the calling thread, `init_mpi` now requests and verifies `MPI_THREAD_FUNNELED` instead of plain `MPI_Init`. The MPI thread levels form a ladder: `SINGLE` (no threads at all), `FUNNELED` (threads may exist, but only the main thread makes MPI calls), `SERIALIZED` (any thread, one at a time), and `MULTIPLE` (concurrent MPI calls from any thread, which implementations support with heavier internal locking and varying quality). `FUNNELED` is the weakest level that permits a writer thread to exist, and it is all we need — the thread only writes local files.
- `FieldGroupIOConfig.asynchronous: bool | None` — the default (`None`) resolves to asynchronous exactly for the backends that support it (zarr). Explicitly asynchronous netCDF is rejected at config time, for two independent reasons. First, distributed netCDF writes are collective MPI operations: every rank must enter variable creation, dimension extension and the collective data writes together, so moving them to a background thread would require `MPI_THREAD_MULTIPLE` plus a dedicated duplicated communicator to keep the IO collectives from mismatching with the halo-exchange collectives on the main thread. Second, netCDF4 sits on HDF5, which keeps global library state and is not thread-safe unless specially built (a build that is in turn unsupported in combination with parallel HDF5) — at a file rollover the main thread would create the next file while the writer thread still flushes the previous one. Zarr has neither problem: chunks are independent objects written with plain filesystem operations and no collectives in the data path. netCDF therefore stays the synchronous fallback.
- This resolves the deep-copy TODO kept in #1307 (see https://github.com/C2SM/icon4py/pull/1307#discussion_r3451869491): `OutputDistribution` now declares `prepare_returns_copy`, and the monitor host-copies the prepared state only in the aliasing single-node path (`utils.host_copy`, exactly one copy on CPU and GPU); the gather and rank-block paths already produce decoupled arrays and pay no extra copy.
- At most `writers.MAX_PENDING_WRITES` (2) captures are in flight, bounding the staging memory; a full queue blocks `store`. A failed background write is re-raised on the owning rank at the next append/close — in rank-block mode the surviving ranks then block in the next barrier, the same failure envelope as a synchronous rank-local write error. `close` drains the queue and joins the thread, so a closed store is complete; `close` is therefore final for asynchronous groups (file rollovers keep the queue alive).

**Timing**

Two phases join the existing report: `async_wait` (time `store` spent blocked on the full queue — the signal that writing does not keep up with the model) and `async_write` (background-thread seconds, overlapped with the model, logged for visibility of the hidden cost).

**Validation**

- io unit tests (126, incl. new queue/equivalence/decoupling/rollover/failure-propagation tests), `mpirun -np 2` MPI io tests (the default zarr rows now run asynchronously; one explicit synchronous row keeps the fallback covered), driver tests, full pre-commit (ruff, mypy, tach).
- Follow-ups deliberately left out: making the queue depth per-group config (only if measurements ask for it), and the dedicated-IO-rank design — the queue front-end introduced here is the seam its consumer would replace, and we plan a performance comparison at scale before committing to it.


model/common/src/icon4py/model/common/decomposition/mpi_decomposition.py
```diff
@@ -58,8 +58,18 @@ def init_mpi() -> None:
...
+        # output writers (io.writers.AsyncWriteQueue) run background threads, but
+        # those only perform local file writes -- all communication, including the
+        # per-append barrier of rank-block zarr output, stays on the main thread.
+        MPI.Init_thread(MPI.THREAD_FUNNELED)
+    provided = MPI.Query_thread()
+    if provided < MPI.THREAD_FUNNELED:
+        raise RuntimeError(
+            f"The MPI library provides thread level {provided}, but 'MPI_THREAD_FUNNELED' "
+            f"({MPI.THREAD_FUNNELED}) is required for asynchronous output."
+        )
```
msimberg, line 72: This triggered me to learn a bit more about what funneled actually means. From what I can tell it's practically indistinguishable from `MPI_THREAD_SINGLE`, and we're likely still fine with `MPI_THREAD_SINGLE` and no practical implementation currently makes this significant. `FUNNELED` seems to really be about _are any threads created at all_ which has of course been true practically forever for icon4py (CUDA spawns threads, OpenMP spawns threads, MPI itself can spawn threads, gt4py spawns worker pools for compilation, etc.).

So I'm a bit torn if we should really change it to this now. If we change this to `FUNNELED` the comment should have no exclusive connection to async IO, because that's not the only reason to use `FUNNELED`.

model/common/src/icon4py/model/common/io/distributed.py
```diff
@@ -110,6 +119,11 @@ def output_horizontal_size(self) -> base.HorizontalGridSize:
     def rank_blocks(self) -> dict[str, "RankBlock"] | None:
         return None
 
+    @property
+    def prepare_returns_copy(self) -> bool:
+        """False: the state is passed through as is."""
+        return False
```
msimberg, line 125: Does this not also depend on whether or not there's a GPU mem to host mem copy (i.e. a host mem buffer is directly usable but is not a copy, a GPU mem buffer will always be copied to host and is usable immediately, maybe?)? Or is that not done at the "prepare" stage? Take this with a grain of salt, I might not have the best picture of what does what.

model/common/src/icon4py/model/common/io/io.py
```diff
@@ -169,6 +181,17 @@ def __post_init__(self) -> None:
...
 
+    @property
+    def is_asynchronous(self) -> bool:
+        """Resolved ``asynchronous`` flag.
+
+        The configured value; when unconfigured (None), asynchronous exactly for the
+        backends that support it (zarr).
+        """
+        if self.asynchronous is None:
+            return self.backend == OutputBackend.ZARR
```
msimberg, line 192: This smells like something that the writer should declare that it supports or not. Can this be resolved through the writer instead? Not entirely convinced this makes it better, but I'd like to know if it was considered because this is hardcoding knowledge about the backend that in principle shouldn't be declared here (but may still be the best option).

model/common/src/icon4py/model/common/io/io.py
```diff
@@ -498,6 +546,11 @@ def __init__(
         self._current_timesteps_in_file = 0
         self._dataset: writers.FieldWriter | None = None
         self._phase_seconds: dict[str, list[float]] = {PHASE_DISTRIBUTE: [], PHASE_WRITE: []}
```
msimberg, line 548: Does this not need to be updated with the new phases?

model/common/src/icon4py/model/common/io/io.py
```diff
@@ -452,6 +492,14 @@ class FieldGroupMonitor(monitor.Monitor):
     With a multi-rank ``process_props``, construction is collective on its communicator
     (the file attributes are broadcast so all ranks carry identical values), just like
     ``store`` is.
+
+    An asynchronous group (``FieldGroupIOConfig.is_asynchronous``) owns one background
+    write thread for the lifetime of the monitor (``writers.AsyncWriteQueue``): at a
+    capture step ``store`` queues the data write and returns, overlapping the file
+    output with the following model steps. All communication -- the ``prepare`` of the
+    distribution as well as the store-metadata operations and barrier of the writer --
```
msimberg, line 500: Only because I've had this result with other agents: are the `--` a result of telling claude not to use em-dashes? It really can't help itself from using them and thinks that it's doing better with `--` instead of an actual em-dash.

No action necessary, just letting off frustration with agents.

model/common/src/icon4py/model/common/io/zarr_writers.py
```diff
@@ -57,6 +71,21 @@ class ZarrWriter:
     ``horizontal_shard_size`` groups whole chunks into one storage file each (see
     ``FieldGroupIOConfig``). In rank-block mode the block size must be a multiple of
     the chunk/shard size (see ``distributed.check_chunks_align_with_blocks``).
+
+    Asynchronous mode (``async_queue`` given): ``append`` performs the store-metadata
+    operations and their broadcast (``_root_step``) synchronously as before, then
```
msimberg, line 76: Can you check the comments for artifacts that refer to what was changed or the past? Like here it says "synchronously as before" but that will have no meaning once this is merged.

model/common/src/icon4py/model/common/io/io.py
```diff
@@ -498,6 +546,11 @@ def __init__(
         self._current_timesteps_in_file = 0
         self._dataset: writers.FieldWriter | None = None
         self._phase_seconds: dict[str, list[float]] = {PHASE_DISTRIBUTE: [], PHASE_WRITE: []}
+        self._write_queue: writers.AsyncWriteQueue | None = (
+            writers.AsyncWriteQueue(name=config.basename, max_pending=writers.MAX_PENDING_WRITES)
+            if config.is_asynchronous
+            else None
+        )
```
msimberg, line 553: Do you think this would benefit from being expressed as two types of queue, one async and one sync, and you just set up the queue once, and submit regardless of sync/async mode, and then the queue decides whether it's blocking or not? You wouldn't have to branch on the `None` case later in that case.

model/common/src/icon4py/model/common/io/writers.py
```diff
@@ -206,3 +213,109 @@ def canonicalize_time_slice(
...
+# ------------------------------------------------------------------------------------
+
+#: Upper bound of queued write tasks of one ``AsyncWriteQueue``. A task holds host
+#: copies of one capture step's fields, so the bound also caps the staging memory at
+#: this many captures. At the bound, ``submit`` blocks until the background thread
+#: catches up (recorded in ``wait_seconds``).
+MAX_PENDING_WRITES: Final[int] = 2
+
+
+class AsyncWriteQueue:
```
msimberg, line 229: Isn't this a much more generic utility than just "write queue"? Did you consider if you could get the same behaviour using something like https://docs.python.org/3/library/concurrent.futures.html?

As it is, the writes are still serial (one at a time), right? And the max pending writes is how many one can queue at the same time?

model/common/src/icon4py/model/common/io/io.py
```diff
@@ -646,12 +711,34 @@ def captures_next_store(self) -> bool:
...
+    def close(self) -> None:
+        """Close the group's writer; final for an asynchronous group.
+
+        Blocks until every queued write is on disk and stops the background write
+        thread -- a subsequent capture would raise. A queued write that failed in the
+        background is re-raised here.
+        """
+        self._close_dataset()
+        if self._write_queue is not None:
+            self._write_queue.shutdown()
```
msimberg, line 744: Reading this naively I'd expect first draining the queue with `shutdown` and then closing the dataset, so I had to go check the implementation and `_close_dataset` first drains, then closes the dataset, and then the queue is shut down. Do you think it'd be more readable if the draining was separated out from the closing of the dataset?

model/common/src/icon4py/model/common/io/zarr_writers.py
```diff
@@ -295,12 +328,17 @@ def _extend_time_axis() -> None:
...
             variable = self._data_array(var_name)
             horizontal_range = self._horizontal_write_range(str(canonical_slice.dims[-1]))
             middle = (slice(None),) * (len(canonical_slice.dims) - 1)
             index: tuple[int | slice, ...] = (time_pos, *middle, horizontal_range)
-            variable[index] = host_data[var_name]
+            writes.append((variable, index, host_data[var_name]))
+        if self._async_queue is None:
+            _write_variables(writes)
+        else:
+            self._async_queue.submit(functools.partial(_write_variables, writes))
```
msimberg, line 341: What do you think about moving this logic outside the writer? The async queue is a very shallow integration, and to me looks like it could live outside the writer itself, and you instead do `queue.submit(<call self._dataset.append(state_to_store, model_time))` on the outside in `FieldGroupMonitor`?

The `FieldGroupMonitor` already has to know about the draining/waiting of the queue, so to me it looks like `ZarrWriter` unnecessarily now gets complicated with the async queue knowledge. Alternatively, if only the `ZarrWriter` does async writes, perhaps it should be fully contained within the writer and `close_dataset` just makes sure the `ZarrWriter` drains the queue. Currently the queue seems to be sitting in both for no obvious reason to me.

That said, you might have had reasons for keeping track of it in both for future changes that I'm not aware of. If that's the case, please let me know what you have in mind then.

msimberg (commented):
Some minor architectural comments and questions, but overall I think this looks good to me, and simple enough. The interesting parts will be how this interacts with IO on subcommunicators.

PR #1432, opened 2026-08-14
Make the dace TopLevelDataFlowStep hook picklable

havogt:
The `TopLevelDataFlowStep` hook set in `get_dace_options` for the vertically implicit solver programs was a lambda, so the customized dace backend executor could not be pickled and the gt4py compilation process pool fell back to compiling in the calling thread. Defining the hook at module scope fixes this.

Before: `pickle.dumps(backend.executor)` raised `AttributeError: Can't get local object 'get_dace_options.<locals>.<lambda>'`. After: it round-trips.

### DIFF
--- a/model/common/src/icon4py/model/common/model_options.py
+++ b/model/common/src/icon4py/model/common/model_options.py
@@ -25,6 +25,14 @@ def dict_values_to_list(d: dict[str, Any]) -> dict[str, list]:
     return {k: [v] for k, v in d.items()}
 
 
+def _remove_access_node_copies(sdfg: Any) -> None:
+    sdfg.apply_transformations_repeated(
+        gtx_transformations.RemoveAccessNodeCopies(),
+        validate=False,
+        validate_all=False,
+    )
+
+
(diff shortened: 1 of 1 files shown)


model/common/src/icon4py/model/common/model_options.py
```diff
@@ -25,6 +25,14 @@ def dict_values_to_list(d: dict[str, Any]) -> dict[str, list]:
     return {k: [v] for k, v in d.items()}
 
 
+def _remove_access_node_copies(sdfg: Any) -> None:
```
edopao: We could be more restrictive on the argument type, though this would only apply to type checking, so not very useful.
```suggestion
def _dace_remove_access_node_copies(sdfg: dace.SDFG) -> None:
```

edopao (approved):

model/common/src/icon4py/model/common/model_options.py
havogt: Applied, thanks.

PR #1434, opened 2026-08-19
Clean up the data_allocation public API and its StencilTest wrapper

egparedes:
Follow-up cleanup of `model/common/src/icon4py/model/common/utils/data_allocation.py`,
focused on making the module's public names say what they do. The module had grown three
overlapping ways to obtain an array namespace, two functions whose names described neither
their input nor their result, and one that only re-spelled a gt4py constructor. Behaviour
is unchanged apart from one dispatch fix, noted below.

- `as_field` is now `reallocate`. It does not build a field; it takes an existing
  `gtx.Field` and returns a copy backed by the given allocator's buffer. The old name
  mirrored `gtx.as_field`, which is a constructor. 8 call sites in driver and
  common/states.
- `field_from_array` is gone. It forwarded its arguments to `gtx.as_field` unchanged,
  differing only in taking the dimensions as varargs, and computed nothing -- while `src/`
  already calls `gtx.as_field` directly 75 times for the same purpose, `serialbox` and
  `factory` doing so with an allocator, character for character what this wrapped. Its one
  caller was `DataAllocationWrapper.field_from_array`, which stays -- it binds the
  allocator, which is the point, so a fixture cannot forget one -- but is renamed
  `as_field` after what it now calls, and goes straight to gt4py like its sibling
  `connectivity_field`. 5 fixture call sites.
- `list2field` is now `scattered_field`, with a docstring for the scatter semantics --
  including that an `indices` entry may be a slice covering a whole dimension, which
  callers rely on and the annotation does not show. 8 call sites in testing/serialbox and
  the Fortran wrappers.
- `array_ns(try_cupy: bool)` is now private `_array_ns`. Its boolean argument made every
  caller run `device_utils.is_cupy_device` first, duplicating what `import_array_ns` is
  for. The three callers now go through `import_array_ns`, leaving two public entry points
  keyed on what the caller actually holds: an allocator, or an array. `SimpleGridData`
  consequently takes the allocator instead of an `on_gpu` flag.
- `scalar_like_array` dispatched with `allocator in (np, xp)`, which recognises exactly two
  module objects by identity and collapses to `(np, np)` when cupy is absent. An
  array-namespace wrapper such as `array_api_compat.cupy` -- what this module's own
  `array_namespace` returns -- fell through to `import_array_ns`, which cannot read a device
  off a module, and so silently allocated on the host. The module form stays in the
  signature because `dycore_wrapper` genuinely uses it, passing a field's own `array_ns`
  with no allocator in scope; the test is now `isinstance(..., ModuleType)`.

`is_ndarray` is deliberately kept rather than removed even though it has no caller outside
this module: it defines `is_rank0_ndarray` and is the only place recording which array
types the public `NDArray` alias covers. The parent branch documents both guards.

Follows #1116, now merged; this branch targets `main` directly, with `main` merged back in.

Verified with `main` at `f7b94f69c` merged in: `pre-commit run --all-files` clean, and 992
passed / 0 failed on `--backend embedded` across the non-datatest model suites, matching
`main` exactly.


egparedes:
cscs-ci run default

jcanton (commented):
**very** bikeshedding worthy, I'm calling the big guns @msimberg @havogt @OngChia @nfarabullini 
I kind of slightly prefer the previous names `list2field` and `as_field`
I don't dislike the new ones but found the old ones more descriptive than `scattered_field` and `reallocate`

model/common/src/icon4py/model/common/utils/data_allocation.py
```diff
@@ -87,20 +87,25 @@ def array_ns(try_cupy: bool) -> ModuleType:
...
+    `allocator` selects where the array is placed: pass an array namespace module to
+    allocate in it directly, or a GT4Py allocator to let its device decide between
+    numpy and cupy. `None` means numpy.
+    """
+    array_ns = allocator if isinstance(allocator, ModuleType) else import_array_ns(allocator)
     return array_ns.asarray(value)
 
 
-def as_field(
+def reallocate(
```
msimberg, line 108: Would `transfer` be a better name for this? I had to double-check the docstring for what this actually does. `reallocate` made me think about sizes changing or just creating a new copy, not directly implying moving to a new allocator. On the other hand, `reallocate` also does fit because it's about the allocator... This is bikeshedding, not convinced at all that my suggestion is better. Feel free to ignore.

msimberg (approved):
Very nice cleanup, thanks @egparedes. One bikeshedding comment about the `reallocate` naming which I initially found a bit confusing, but can probably get used to.

msimberg:
> **very** bikeshedding worthy, I'm calling the big guns @msimberg @havogt @OngChia @nfarabullini I kind of slightly prefer the previous names `list2field` and `as_field` I don't dislike the new ones but found the old ones more descriptive than `scattered_field` and `reallocate`

I didn't see your comments when I started my review and had a similar comment on the `as_field`/`reallocate` renaming. `list2field`/`scattered_field` I feel less strongly about. Repeating my review, but I love the cleanup and agree that the names are very bikesheddable, and don't oppose the current or old names.

model/common/src/icon4py/model/common/utils/data_allocation.py
egparedes: I picked `reallocate` specifically because it gets an allocator, but I do not oppose to any other name.

nfarabullini (approved):

egparedes:
> I kind of slightly prefer the previous names `list2field` and `as_field`
> 

The problem with `list2field` is two-fold: it doesn't get a list in the arguments and it does not reflect that the values are not interpreted as a regular dense field, but in a completely different way, a d this is actually the main point of the function.

`as_field` is okish and we could rename it back if you like, but I find it slightly confusing and less specific than `reallocate` or `transfer` (as suggested by @msimberg) since its main argument already is a field, and the only thing it does is to allocate a copy with another allocator

model/common/src/icon4py/model/common/utils/data_allocation.py
egparedes: > Would `transfer` be a better name for this? I had to double-check the docstring for what this actually does.

`transfer` could also work, yes. I picked `reallocate` specifically because it gets an allocator, but I do not oppose to any other reasonable name.

jcanton:
> > I kind of slightly prefer the previous names `list2field` and `as_field`
> 
> The problem with `list2field` is two-fold: it doesn't get a list in the arguments and it does not reflect that the values are not interpreted as a regular dense field, but in a completely different way, a d this is actually the main point of the function.
> 
> `as_field` is okish and we could rename it back if you like, but I find it slightly confusing and less specific than `reallocate` or `transfer` (as suggested by @msimberg) since its main argument already is a field, and the only thing it does is to allocate a copy with another allocator

I may have forgotten to explain in more detail 🙃  and forgotten to say I also like the nice cleanup!

isn't `list2field` / `scattered_field` essentially transforming a sparse array into a dense one? that's what I was thinking about, didn't mean to say that list and fields were correct names, but scattered seems unspecific (and reminds me of a scatter plot)

between `as_field` / `reallocate` / `transfer` I may like reallocate more (for the reason that @egparedes wrote above), my question there was: is it really both ways or only used to go from `xp` -> `gt4pyField_whatever_the_allocator`?
if it's one way I liked the word field because of that, but if it works both ways then I had not understood.

egparedes:
cscs-ci run default

PR #1436, opened 2026-08-19
refactor(physics_driver): two layer physics state

yiluchen1066:
Restructures the physics coupling into two layers of state under **parallel process
coupling**: every process computes its tendencies from the same frozen step-entry
state, the tendencies are accumulated across processes, and one apply step writes
back to the model state. No copies anywhere. 

## The two layers

**PhysicsState layer** — driver-owned, `physics_driver/physics_state.py`, the only
code that touches the model state:
- `EntryState` — binds *pointers* to `exner, theta_v, rho, vn, w,
  tracers` (same memory, physics names) and owns the six diagnosed fields,
  grouped in the common `DiagnosticState` container so the
  pointer/diagnostic split is visible in the types
  (`temperature, virtual_temperature, pressure, pressure_ifc, u, v`).
  `diagnose_from` runs once per step; frozen afterwards (tested invariant).
- `TendencyAccumulators` — lazily allocated per-variable sums of every component
  output tagged `kind == "tendency"`; zeroed each run. Tendency buffers stay
  per-component, the accumulators are summed after all processes. 
- `DiagnosticsStore` — creates and owns one buffer for each diagnostic
  (non-tendency) output, with the shape taken from the output's `dims`
  metadata. When the driver is constructed, the store gives these buffers to
  the component (`bind_output_buffers`), and the granule then writes its
  results directly into them (`driver.diagnostics[process_name][output]`). So that components still create their own buffers when run standalone, so the granule datatests are unchanged.
- `ApplyToPrognostic` — the single write

**ComponentState layer** — one adapter per process, Renamed from the previous ```state(PhysicsState)``` . It had two methods, and no storage:
`collect_inputs(entry_state)` + `as_component_input()`. The muphys adapter reduces
to a pure input mapping. And ``collect_inputs`` is renaming from `gather` throughout.

And this pr is in the middle of  a stack (#1359: tmx port and #1360: tmx integration). tmx integration build on it. The tmx side of the `DiagnosticsStore` changes lands in #1360.


model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,433 @@
...
+        may touch is reachable as ``entry_state.<name>`` — the raw PrognosticState
+        and TracerState never appear below the PhysicsState layer. Strictly
+        read-only for the processes: only the driver's single apply step writes,
+        through the pointers bound here.
+        """
+        # Physics needs all six moisture species; TracerState fields are optional (a
+        # tracer may be inactive per TracerConfig), so fail loudly once here rather
+        # than feed None into the physics.
+        missing = [name for name in MOISTURE_SPECIES if getattr(tracers, name) is None]
+        if missing:
```
msimberg: Nit:

```suggestion
        if missing := [name for name in MOISTURE_SPECIES if getattr(tracers, name) is None]:
```

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,433 @@
...
+    def zero(self) -> None:
+        """Reset all accumulators; called by the driver at the start of every run."""
+        for buffer in self.acc.values():
+            buffer.ndarray[...] = 0.0  # type: ignore[index] # NDArrayObject Protocol doesn't support this
+
+    def accumulate(self, outputs: dict, outputs_properties: dict) -> None:
+        """Add a process's tendency outputs to the per-variable sums.
+
+        Element-wise sum with no neighbor access — a plain array operation on the
+        field buffers, valid for numpy and cupy backing storage alike.
```
msimberg: I think in general you can avoid comments like "valid for numpy and cupy backing storage alike". I'm not sure it adds anything useful?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,433 @@
...
+        """
+        for name, props in outputs_properties.items():
+            if props.get("kind") != "tendency":
+                continue
+            field = outputs[name]
+            buffer = self.acc.get(name)
+            if buffer is None:
+                buffer = gtx.zeros(field.domain, dtype=field.dtype, allocator=self._backend)
```
msimberg: Is there any case where `buffer` would stay `None`, i.e. where there are no `tendency` fields? If no, I'd set `buffer` up outside the for loop.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,433 @@
...
+                self._apply_tendency(
+                    field_a=tracer,
+                    coeff=dt_seconds,
+                    field_b=acc[key],
+                    output_field=tracer,
+                )
+
+        # 2. Temperature -> exner/theta_v: ONE exact-EOS update from the entry T
+        #    plus the summed tendency, with the final (post step 1) moisture.
+        if "tend_temperature" in acc:
```
msimberg, line 343: Are `tend_temperature`, `tend_u`, and `tend_w` really optional, or should they be checked more strongly (i.e. is it ok if they're silently skipped)?

model/common/src/icon4py/model/common/interpolation/stencils/compute_vn_from_uv.py
```diff
@@ -0,0 +1,69 @@
...
+import gt4py.next as gtx
+from gt4py.next import neighbor_sum
+
+from icon4py.model.common import dimension as dims, field_type_aliases as fa
+from icon4py.model.common.dimension import E2C, E2CDim
+from icon4py.model.common.type_alias import wpfloat
+
+
+@gtx.field_operator
+def _compute_vn_from_uv(
```
msimberg, line 17: According to https://github.com/C2SM/icon4py/blob/main/docs/stencil_naming_convention.md I think this is a more generic program than just computing the wind. Not blocking for the PR, but should be considered at least. There may even already exist a similar more generic program?

model/common/src/icon4py/model/common/components/component_state.py
```diff
@@ -0,0 +1,30 @@
...
+    Binds the frozen ``EntryState`` facade of the PhysicsState layer (the first
+    layer, ``icon4py.model.atmosphere.subgrid_scale_physics.physics_driver.
+    physics_state``) and maps it to this component's input contract, adding any
+    process-specific derived inputs. Stateless beyond the bindings: the
+    component's outputs are routed by the ``PhysicsDriver`` into the PhysicsState
+    layer's sinks (tendency accumulators / diagnostics store) — a ComponentState
+    stores nothing and never writes to the model state.
+
+    ``entry_state`` is typed ``Any`` to keep ``common`` decoupled from the
+    ``physics_driver`` package that defines the facade.
```
msimberg: If this is really generic, then the docstring should not talk about physics specifically (except perhaps as an example). It should describe what this is used for generically.

model/common/src/icon4py/model/common/components/component_state.py
```diff
@@ -0,0 +1,30 @@
...
+    """
+
+    def collect_inputs(self, entry_state: Any) -> None: ...
```
msimberg: I still have to understand this a bit better, but `collect_inputs` feels like it's designed quite specifically for the physics state use case. Do you think this is something that every component state would need?

model/common/src/icon4py/model/common/components/component_state.py
```diff
@@ -0,0 +1,30 @@
...
+    def as_component_input(self) -> dict[str, Any]: ...
```
msimberg: Nothing against this the way we have things set up currently (untyped dicts as inputs/outputs), but if we make this more typed, then I'm guessing `as_component_input` as a member function on a component state won't work so well anymore. What inputs a particular component needs is known by the component, not by the state from which the fields are taken.

Not necessarily something we need to worry about in this PR, but definitely a concern for the bigger components/state redesign.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,433 @@
...
+            buffer.ndarray[...] = 0.0  # type: ignore[index] # NDArrayObject Protocol doesn't support this
+
+    def accumulate(self, outputs: dict, outputs_properties: dict) -> None:
+        """Add a process's tendency outputs to the per-variable sums.
+
+        Element-wise sum with no neighbor access — a plain array operation on the
+        field buffers, valid for numpy and cupy backing storage alike.
+        """
+        for name, props in outputs_properties.items():
+            if props.get("kind") != "tendency":
```
msimberg: Ah, one more thing: we should definitely move away from the stringly kinds here and use proper enums. Not required for this PR, and can probably easily be changed in parallel in a separate PR.

model/common/src/icon4py/model/common/interpolation/stencils/compute_vn_from_uv.py
yiluchen1066: I have checked and there is no pre-existing generic equivalent. ```compute_vn_from_uv``` is the only uv to edge normal projection in ```common```. The closest is ```compute_tangential_wind```, which #1452 just moved from dycore into the very same directory; these two are the normal and tangential halves of the same projection. So the rename is worth doing, but the two names should probably be decided together rather than one in isolation. We can do this either in the follow-up pr or in this pr if you prefer

model/common/src/icon4py/model/common/components/component_state.py
yiluchen1066: Agreed. I have rewritten it with the generic contract and also mentions the physics driver only as the current user. 
and also I agreed sometimes "friends" wrote too many verbose comments. I will also go over all of them and make a clean in this pr

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
yiluchen1066: you are right that ```buffer``` cannot remain ```None``` after this check. 

Each tendency variable has its own accumulator in ```self.acc.get(name)```. When a variable is encountered for the first time, its accumulator is created. Therefore, I think we need to put the buffer set inside the loop, moving it outside would incorrectly reuse one buffer for different tendency variables.

model/common/src/icon4py/model/common/components/component_state.py
yiluchen1066: here, ```collect_inputs``` is to (implied by the name) generally collect or prepare the inputs for components. But it is used completely different by two components: for ```muphys```, it only saves a pointer to the entry state (```self._entry=entry_state```) and ```as_component_input``` then reads through that pointer and assembles the ten-key input dict, translating the layer's field names into the granule's argument names (for example ```diagnostics.temeprature -> te```).  For ```tmx```, it actually computes things and prepares what tmx needs:  air mass, cv_air and the surface-flux provider call. This is a bit difficult for people to understand. 

So I am thinking that we can collapse these two to a single method: 

```suggestion
     def as_component_input(self, state: Any) -> dict[str, Any]: ...
```

Then the differences among the components then can live inside this single methods which only runs on firing steps. This can also points the same way as you said in your next comment: 
"once inputs are typed, what a component needs is the component knowledge rather than the state's" 

Once we decide on the bigger components/state redesign, one method should be smaller and easier to redesign or match on. 

if you think this is a good idea, I would like to do it in the pr and then tmx in #1360. Please feel free to provide your opinion

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
yiluchen1066: yes, you are right. But then I realized that there was a specific reason for me to keep the ```kind``` for output since I would need this to select which outputs are put into the the ```TendencyAccumulator```. So I am thinking of two potential ways to do that: 
1. we can only have tendencies in outputs, then we will need to think about a smarter way to store diagnostic output
2. otherwise, we keep this kind, but make it better. 

Please also fell free to provide your ideas

model/common/src/icon4py/model/common/interpolation/stencils/compute_vn_from_uv.py
msimberg: Yes, follow up is fine. The important thing is that it's in common where it's hopefully a bit more likely to get reused. The rename can come later.

model/common/src/icon4py/model/common/components/component_state.py
msimberg: Thanks for the explanation! That makes it clearer, and I think it aligns also with what I've been trying to look at in terms of the generic component design.

I think you can indeed combine them right now into e.g. `as_component_input`, that looks aligned with where we're hopefully going. Good idea.

Outside of that I think there are then:
- The naming variations we can hopefully get rid of over time, i.e. we always refer to them with canonical (cf conventions if possible) names and if a component wants to refer to a field internally with a different name, noone is stopping it.
- The tmx case is the more interesting one. I think indeed you can for now combine it such that the fields are internally computed. The interesting case is if we ever have two components that need the same fields computed. But hopefully the explicitly typed input to a component will help us declare that a bit more nicely. And hopefully we can lift the computation of those fields out from the individual components so that it's not recomputed.

model/common/src/icon4py/model/common/components/component_state.py
msimberg: Just making another note in relation to https://github.com/C2SM/icon4py/pull/1436#discussion_r3933397054: basically what I'm saying here is that when things are more strictly typed, `as_component_input` effectively becomes a method on the `Component` type, i.e. either a member or free function dispatched on the type, because one can then determine which fields to collect as inputs from the state.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
msimberg: Ah, I might have made this sound more serious than it is. I just meant to go toward something like:
```
props.kind != kind.TENDENCY
```
i.e. slightly more strongly typed. The feature itself is good!

The only other alternative formulation I've been thinking about for this is to declare tendency fields separately and then give them a property like `tendency_of = "some_other_field"` which allows discovering (at least in one direction) what the target field is.

But to repeat, no change needed in this PR.

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/component.py
```diff
@@ -114,12 +114,25 @@ def __init__(
         self._step = step
 
         cell_k_domain = gtx.domain({dims.CellDim: self._ncells, dims.KDim: self._nlev})
-        self._pflx = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
-        self._pr = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
-        self._ps = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
-        self._pi = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
-        self._pg = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
-        self._pre = gtx.zeros(cell_k_domain, dtype=ta.wpfloat, allocator=allocator)
+        # standalone defaults — replaced by the layer's buffers via bind_output_buffers
```
msimberg: Can probably also be removed?

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/state.py
```diff
@@ -8,253 +8,38 @@
...
-      - reference  : a pointer into the dycore state, no copy -- dz, rho, q.
-      - owned      : a buffer allocated once here and overwritten in place each
-                     step -- te, p, tv, pressure_on_cells_half_levels, and the scratch buffers.
+    Maps the frozen ``EntryState`` of the PhysicsState layer to the muphys component contract.
+    The layer already diagnoses everything muphys consumes
+    (T, p) and points at the rest (rho, tracers), so this is pure name
+    translation and muphys derives nothing of its own; the only process-owned
+    input is ``dz``, fetched once from the metrics source. Holds no state beyond
+    it — muphys's outputs (tendencies, precip diagnostics) are routed by the
+    driver into the PhysicsState layer's sinks.
```
OngChia: Is it possible to make the docstring more readable?
And I think we can remove the term "layer" because users do not probably have the full picture of what we have been discussing. So, it does not mean anything to them. Moreover, I think we do not need to explicitly say dz is fetched from metric and outputs are tendencies and precipitation as they are quite clear from the code below. Maybe just mention non-static fields are taken from state?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
```diff
@@ -12,44 +12,90 @@
...
+    ``bind_output_buffers`` under type checking rather than by duck typing.
+    """
+
+    def bind_output_buffers(self, buffers: dict[str, Any]) -> None:
+        """Adopt the layer-owned buffers for this component's diagnostic outputs.
+
+        Called once per process at driver construction. Implementations keep their
+        own allocations as the standalone default (e.g. the granule datatests) and
+        replace them with these, so the granule writes into the layer's memory in
+        place and the driver never copies a diagnostic.
```
OngChia: I think we can rephrase this paragraph and remove any information about how the driver works.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
```diff
@@ -12,44 +12,90 @@
...
+    """Runs the physics processes under parallel coupling.
+
+    One timestep (``run``): the ``EntryState`` binds the model state and
+    diagnoses the physics fields once (dyn2phy); every enabled process reads that
+    same frozen state and computes; outputs tagged ``kind == "tendency"`` are
+    accumulated, while the diagnostic outputs are written by the granules directly
+    into the layer-owned ``diagnostics`` store buffers bound at construction; the
+    accumulated tendencies are applied to the model state exactly once at the end.
+    Processes never see the raw PrognosticState/TracerState and never
+    write it — the PhysicsState layer owns both conversion boundaries.
```
OngChia: I think we can remove the word "layer" in this paragraph?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
```diff
@@ -62,23 +108,22 @@ def run(
...
-            state.gather_from_prognostic(prognostic, tracers)
-            if not tc.enable_process:
-                continue
-            if not tc.is_in_window(step_start_datetime):
-                # outside the process window: no forcing
+            if not tc.enable_process or not tc.is_in_window(step_start_datetime):
                 continue
             # Compute on a firing (active) step, and also on the first in-window step -- when
             # there is nothing cached to recycle yet. Otherwise reuse the last computed forcing.
             if tc.is_active(step_start_datetime) or process.name not in self._recycle_cache:
```
OngChia, line 111: I am wondering whether we can move these checking about whether we run a particular physics process at the current time step inside the `PhysicsProcess`?
Something like:

```
for process in self._processes:
    process(
        process.state.as_component_input(self._entry), step_start_datetime
    )
```

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/component.py
```diff
@@ -114,12 +127,24 @@ def __init__(
...
+        )
+        self._pi: fa.CellKField[ta.wpfloat] = gtx.zeros(
+            cell_k_domain, dtype=ta.wpfloat, allocator=allocator
+        )
+        self._pg: fa.CellKField[ta.wpfloat] = gtx.zeros(
+            cell_k_domain, dtype=ta.wpfloat, allocator=allocator
+        )
+        self._pre: fa.CellKField[ta.wpfloat] = gtx.zeros(
+            cell_k_domain, dtype=ta.wpfloat, allocator=allocator
+        )
```
OngChia, line 147: Since you eventually bind these output to fields in an external state, is there a way to avoid these allocation and just bind the the external fields?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
```diff
@@ -62,23 +108,22 @@ def run(
...
-                # compute
-                outputs = process.component(state.as_component_input(), step_start_datetime)
+                outputs = process.component(
+                    process.state.as_component_input(self._entry), step_start_datetime
+                )
                 self._recycle_cache[process.name] = outputs
             else:
-                # recycle
                 outputs = self._recycle_cache[process.name]
```
OngChia, line 117: Okay, now I understand how you do with the tendencies when a physics process is skipped. A follow-up question of my comment above, instead of manually doing this recycle, is it more readable if we can somehow just do 
```
self._accumulators.accumulate(process.output)
```
when we accumulate?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
```diff
@@ -12,44 +12,90 @@
...
-    The component is the per-process adapter (e.g. ``MuphysComponent``); it
-    implements the generic ``Component`` protocol, which is how the driver types it.
-    The state adapter is process-specific (it translates the prognostic state to/from
-    *this* component's contract), so it is bundled per process rather than shared.
+    The component is the per-process adapter (e.g. ``MuphysComponent``), typed as
+    a ``PhysicsComponent``: the generic ``Component`` contract plus the
+    ``bind_output_buffers`` hook the driver needs to hand it the layer-owned
+    diagnostic buffers. The state adapter is process-specific (it maps the frozen
+    entry state to *this* component's contract), so it is bundled per process
+    rather than shared.
```
OngChia: The docstring is quite difficult to understand to me. Could you make it easier to understand?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+            field = outputs[name]
+            if (buffer := self.acc.get(name)) is None:
+                buffer = self.acc[name] = gtx.zeros(
+                    field.domain, dtype=field.dtype, allocator=self._backend
+                )
+            buffer.ndarray[...] += field.ndarray  # type: ignore[index] # NDArrayObject Protocol doesn't support this
+
+
+class ApplyToPrognostic:
+    """The single application of the accumulated tendencies to the model state.
```
OngChia, line 226: TendencyAccumulators and this class are closely related. Does it make sense if we can combine them together?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+        # 3. Cell-centre (u, v) from edge-normal vn via RBF
+        self._rbf_interpolation(
+            p_e_in=prognostic.vn,
+            p_u_out=self.diagnostics.u,
+            p_v_out=self.diagnostics.v,
+        )
+
+
+class TendencyAccumulators:
+    """Per-variable tendency sums over the processes of one timestep (ICON ``tend%*_phy``).
```
OngChia: we can remove the statement about icon in the parentheses.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+
+class ApplyToPrognostic:
+    """The single application of the accumulated tendencies to the model state.
+
+    The phy2dyn conversion of ``mo_interface_iconam_aes`` (``:513`` and following),
+    executed once per timestep — previously each process's ``scatter_to_prognostic``
+    repeated the exner/theta_v EOS update. Order: tracers first (the EOS update uses
+    the final moisture), then temperature -> exner/theta_v, then the winds.
+    Tendencies absent from the accumulators are skipped (e.g. no ``tend_u/v/w`` in a
+    muphys-only configuration).
```
OngChia: You can delete this paragraph where you explain about the corresponding icon code and your previous implementation. And I think you can simply explain more about `Tendencies absent from the accumulators are skipped` without mentioning a specific application such as muphys.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+            },
+            horizontal_sizes=edge_horizontal,
+            vertical_sizes=full_vertical,
+            offset_provider=grid.connectivities,
+        )
+
+        # Scratch buffers — allocated once
+        self._new_te = data_alloc.zero_field(grid, dims.CellDim, dims.KDim, allocator=backend)
+        self._tv_tendency = data_alloc.zero_field(grid, dims.CellDim, dims.KDim, allocator=backend)
+        self._ddt_vn = data_alloc.zero_field(grid, dims.EdgeDim, dims.KDim, allocator=backend)
```
OngChia, line 316: This is fine for the time being. But in NWP, ddt_vn accumulated from physics is passed back to dycore.

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+
+class DiagnosticsStore:
+    """Layer-owned storage of the process output diagnostics (ICON ``field%`` spirit).
+
+    Allocates every non-tendency output from its declared metadata (``dims``,
+    optionally ``is_on_half_levels``) at driver construction and hands the
+    buffers to the component (direct write): one buffer per field, owned here,
+    written by the granule. Keyed per process for order-independence and
+    collision-safety. Values are the last computed step's; zeros before a
+    process first fires.
```
OngChia: Rephrase this docstring to make it more concise and remove icon-related stuff?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,437 @@
...
+        buffers: dict[str, gtx.Field] = {}
+        for name, props in outputs_properties.items():
+            if props.kind == model.FieldKind.TENDENCY:
+                continue
+            if props.dims is None:
+                raise ValueError(
+                    f"diagnostic output '{name}' of process '{process_name}' declares no dims; "
+                    "the store allocates its buffer from that metadata"
+                )
+            extend = {dims.KDim: 1} if props.is_on_half_levels else None
```
OngChia, line 428: we now have `KHalfDim`, so I think generally you can remove `extend`, unless you have some variables that are defined only in a subset of the `KDim` or `KHalfDim`.

model/common/src/icon4py/model/common/states/model.py
```diff
@@ -24,32 +25,61 @@
...
+    #: we might not have this one for all fields. But it is useful to have it for tractability with ICON
+    icon_var_name: str | None = None
+    #: list index for variables stored in fortran lists (e.g. tracers)
+    icon_var_list_index: int | None = None
+    # TODO(halungge): dims should probably be required?
+    dims: Sequence[gtx.Dimension] | None = None
+    dtype: ta.wpfloat | ta.vpfloat | gtx.int32 | gtx.int64 | gtx.float32 | gtx.float64 | None = None
+    #: whether the vertical dimension of the field lives on interface (half) levels
+    #: rather than full levels
+    is_on_half_levels: bool | None = None
```
OngChia, line 67: Not sure we still need this or not after `KHalfDim` is introduced.

model/driver/src/icon4py/model/driver/driver_utils.py
```diff
@@ -465,15 +468,31 @@ def initialize_granules(
...
+                interpolation=interpolation_field_source,
+                metrics=metrics_field_source,
+                backend=backend,
+            ),
+            accumulators=physics_state.TendencyAccumulators(backend=backend),
+            apply_to_prognostic=physics_state.ApplyToPrognostic(
+                grid=grid,
+                geometry=geometry_field_source,
+                interpolation=interpolation_field_source,
+                backend=backend,
```
OngChia, line 492: Can we just pass in factories, grid, and backend, and instantiate `entry_state`, `accumulators`, `apply_to_prognostic`, and `diagnostics` inside the initialization of physics driver?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
yiluchen1066: I think it is a good idea! thanks!

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_driver.py
yiluchen1066: agreed!

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,444 @@
...
+        """
+        for name, props in outputs_properties.items():
+            if props.kind != model.FieldKind.TENDENCY:
+                continue
+            field = outputs[name]
+            if (buffer := self.acc.get(name)) is None:
+                buffer = self.acc[name] = gtx.zeros(
+                    field.domain, dtype=field.dtype, allocator=self._backend
+                )
+            buffer.ndarray[...] += field.ndarray  # type: ignore[index] # NDArrayObject Protocol doesn't support this
```
OngChia, line 222: Use field operator to add tendency?

model/atmosphere/subgrid_scale_physics/physics_driver/src/icon4py/model/atmosphere/subgrid_scale_physics/physics_driver/physics_state.py
```diff
@@ -0,0 +1,444 @@
...
+        if "tend_w" in acc:
+            self._apply_tendency_w(
+                field_a=entry_state.w,
+                coeff=dt_seconds,
+                field_b=acc["tend_w"],
+                output_field=entry_state.w,
+            )
+
+
+class DiagnosticsStore:
```
OngChia, line 401: Rename?

OngChia (approved):

PR #1438, opened 2026-08-21
Fixes to bump_version-related scripts

nfarabullini:
A few things needed to be edited after errors popped up in the 0.4.0 release:
 - `bump_version.py` didn't upgrade if version of smth was not the same as other (e.g. lower) nor it threw an error
 - CI didn't catch that some versions weren't upgraded 
 - `bump_version.py` didn't catch that physics_driver's .toml file didn't have a bump-version section


.github/workflows/icon4py-qa.yml
```diff
@@ -36,5 +36,8 @@ jobs:
       - name: Install icon4py namespace packages
         run: uv sync --extra all
 
+      - name: Check package version consistency
+        run: ./scripts/run check-versions
+
```
msimberg: Can this be made a pre-commit hook?

docs/release_procedure.md
```diff
@@ -52,6 +52,16 @@ This updates `version` in every `pyproject.toml` and `__init__.py`, and also
...
 `icon4py-common~=0.2.0`).
 
+Before making changes, the script verifies that every package `pyproject.toml`
+has a `[tool.bumpversion]` section, all `icon4py.*` package versions agree, and all pinned
+cross-package dependency constraints use that version.
+
+The same checks run in the quality-assurance CI workflow:
+
+```bash
+./scripts/run check-versions
```
msimberg: See other comment about pre-commit. If CI checks this and the check runs with pre-commit, it might not need to be mentioned here?

docs/release_procedure.md
```diff
@@ -83,6 +93,12 @@ Publishing the GitHub Release automatically triggers the
    pip install --index-url https://test.pypi.org/simple/ icon4py==<new_version>
    ```
 
+   If this fails, install `icon4py` with PyPI as an extra index:
+
+   ```bash
+   python3 -m pip install --extra-index-url https://pypi.org/simple -i https://test.pypi.org/simple icon4py==<new_version>
```
msimberg: Maybe consistently use `pip` vs `python3 -m pip`? But do note that the `**Note**` just below says this as well, maybe this can be integrated a bit nicer with the note below?

The `--index-strategy unsafe-best-match` is actually only required for `uv pip install` I think, and that should probably be mentioned explicitly (the reason is that `uv pip install` will decide to not versions/packages/something from different indices unless you ask it to do that with the `--index-strategy` flag; plain `pip install` is less strict about it).

msimberg (changes requested):
Looks good as a whole, some minor requests for changes. Thanks for updating this!

Potentially a follow-up PR or not needed at all, but would you be able to investigate if there's any way we can run in CI a test install that would've also caught the inconsistent versions? A plain `uv sync` clearly does not catch it and I'm wondering if there's any other `uv`/`pip` setup that would've verified that the packages actually are not installable together without having to push to pypi.org?

docs/release_procedure.md
```diff
@@ -83,19 +87,22 @@ Publishing the GitHub Release automatically triggers the
...
-   You may also need `--index-strategy unsafe-best-match` to allow picking the
-   best match from pypi.org and test.pypi.org. Finally, `--refresh` may be
-   needed to update cached package information.
+   If installation fails, try again with PyPI as an extra index:
+
+   ```bash
+   pip install --extra-index-url https://pypi.org/simple -i https://test.pypi.org/simple icon4py==<new_version>
+   ```
+
+   **Note:** TestPyPI may not have all transitive dependencies.
```
msimberg: Maybe still simpler:
````suggestion
   If installation fails, try again with PyPI as an extra index because TestPyPI may not have all transitive dependencies:

   ```bash
   pip install --extra-index-url https://pypi.org/simple -i https://test.pypi.org/simple icon4py==<new_version>
   ```
````
?

msimberg (approved):

PR #1446, opened 2026-08-26
Updated thread block configuration for gh200

iomaganaris:
- Set thread block size for GH200 GPU to (128, 2) for 2D kernels and (256, 1) for 1D kernels
- Updates to the structure of the `get_dace_options` were necessary to avoid `ruff` complains
- There were other options and heuristics considered for the thread block size but this option was the best for global and regional grids on 1 GPU

image


edopao (commented):
Nice refactoring, LGTM.

PR #1443, opened 2026-08-24
Restrict OMP_NUM_THREADS in bencher pipelines

msimberg:

PR #1386, opened 2026-07-21
Speed up CSCS CI jobs

msimberg:
This slightly tweaks the CSCS CI configurations:
- Sets OMP_NUM_THREADS=8 for all jobs to avoid oversubscription with jobs that use a lot of test workers or MPI ranks. This speeds up CPU backend tests a bit.
- Reduces the time limit for driver/common MPI test jobs to 1.5 hours.
- Uses only two GPUs per node for driver/common MPI test jobs. They currently use a full node but only use one GPU. One GPU is a bit tight with concurrent tests, so two GPUs is a compromise to slightly speed up the total wall time for those test jobs without taking the full node. The MPS wrapper is also changed to actually bind ranks to both (all) GPUs that are available instead of only using the first GPU.

PR #1448, opened 2026-08-26
CI: switch to normal partition

edopao:
Switching to slurm partition `normal`, since the `shared` partition was deleted on Säntis vCluster. The GPU count is kept the same as before: 1 GPU for regular jobs, 4 GPUs (full node) for MPI jobs and for benchmark jobs.

### DIFF
--- a/.cscs-ci/base.yml
+++ b/.cscs-ci/base.yml
@@ -286,7 +286,6 @@ variables:
     OMPI_MCA_mpi_cuda_support: "0"
     SLURM_NTASKS: "1"
     SLURM_GPUS_PER_TASK: "1"
-    SLURM_PARTITION: "normal"
     SLURM_TIMELIMIT: "00:10:00"
   script:
     - cd /icon4py
--- a/.cscs-ci/base.yml
+++ b/.cscs-ci/base.yml
@@ -75,7 +75,7 @@ variables:
   variables:
     SLURM_JOB_NUM_NODES: 1
     SLURM_GPUS_PER_NODE: 1
-    SLURM_PARTITION: "shared"
+    SLURM_PARTITION: "normal"
     SLURM_CPU_BIND: "verbose"
     SLURM_TIMELIMIT: '00:30:00'
     SLURM_MPI_TYPE: pmix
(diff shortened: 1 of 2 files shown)


edopao:
cscs-ci run default

edopao:
cscs-ci run default

edopao:
cscs-ci run merge

msimberg (approved):
Yes, thank you! I may tweak num gpus slightly in #1386, but that's a follow up. Getting this up and running again is highest priority.

msimberg:
Let's see. Build farm doesn't seem to be accepting jobs yet, and we should wait for one clean run to make sure the config is ok. It looks ok, but I'd like to see it in practice before merging.

edopao:
cscs-ci run default

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -17,7 +17,9 @@
   extends: [.benchmark_model_base_variables]
   variables:
     SLURM_JOB_NUM_NODES: 1
-    SLURM_PARTITION: "normal"
+    SLURM_GPUS_PER_NODE: 4
+    SLURM_GRES_FLAGS: allow-task-sharing
+    SLURM_PARTITION: normal
```
iomaganaris, line 22: Those are actually for the moment single rank jobs so they could also run with the default shared allocation

msimberg: Let's change that in #1443? I'm already setting tasks per node and gpus per task there.

iomaganaris: Right, I forgot about this change there 🙈

havogt:
Waiting for a fix in the CI infrastructure for properly forwarding task sharing.

.cscs-ci/base.yml
```diff
@@ -169,6 +169,8 @@ variables:
     SLURM_NTASKS: 12
     ICON4PY_TEST_MPI_SUBCOMM_SIZE: 4
     SLURM_LABELIO: 1
+    SLURM_GPUS_PER_NODE: 4
+    SLURM_GRES_FLAGS: allow-task-sharing
```
msimberg: We don't need 4 gpus per node here (and didn't rely on it in the past).
```suggestion
```

msimberg:
cscs-ci run default

PR #1450, opened 2026-08-31
Add driver bencher benchmarks

msimberg:
This PR adds end-to-end standalone driver benchmarks to the Bencher pipeline, timing the full JW steady-state aquaplanet global experiment over initialization, the time integration loop, and the total run.

Single-rank (`integration_tests/test_benchmark_driver.py`) and MPI (`mpi_tests/test_benchmark_driver.py`) variants share fixtures and constants in `model/driver/tests/driver/fixtures.py`. Both expose `test_benchmark_driver_init`, `test_benchmark_driver_timeloop`, and `test_benchmark_driver_total`, parameterized by `driver_benchmark_experiment` (JW, 100 steps, 50 s timestep). Benchmark ids are `False` for single-rank and `distributed` for MPI. The timeloop benchmark rebuilds driver states per round via a shared `initialize_driver_states` helper.

CI adds serial and distributed driver Bencher jobs to the feature and baseline pipelines across four backends (dace_cpu, dace_gpu, gtfn_cpu, gtfn_gpu) and R02B06_GLOBAL. Serial testbeds use `_serial_testbed()` shape `RUNNER:SYSTEM_TAG:BACKEND:GRID`; MPI testbeds append `:<nodes>N<ranks>R:<GHEX_TRANSPORT_BACKEND>`. MPS is disabled (one GPU per task), `OMP_NUM_THREADS=72`, and `.retry_on_transient_failure` is dropped from driver jobs to match the stencil/granule convention.

`_name_from_fullname` in `pytest_hooks.py` now shortens function-based benchmark fullnames so they appear in Bencher alongside StencilTest names.


.cscs-ci/benchmark_bencher.yml
```diff
@@ -18,6 +18,31 @@ include:
...
+    NOX_SESSION: "__bencher_driver_feature_branch_CI-${PYVERSION_SHORT}"
+    TEST_SELECTION: "test_benchmark_driver"
+    DRIVER_BENCHMARK_EXPERIMENT: "jw"
+    DRIVER_BENCHMARK_STEPS: 10
+    DRIVER_BENCHMARK_ROUNDS: 5
+    DRIVER_BENCHMARK_WARMUP_ROUNDS: 2
+    BACKEND: "dace_gpu"
+    GRID: "R02B04_GLOBAL"
+    SLURM_TIMELIMIT: '02:00:00'
+
```
msimberg, line 30: Two points: the new job should probably be "distributed" rather than "driver", in case we get more distributed bencher benchmarks. And it'd be good ot avoid duplicating all the `script` setup (into a helper script or a template).

.cscs-ci/benchmark_bencher.yml
```diff
@@ -28,3 +53,25 @@ benchmark_bencher_granules_feature_aarch64:
...
+benchmark_bencher_driver_feature_inter_aarch64:
+  extends: [.bencher_driver_feature_tests, .retry_on_transient_failure]
+  variables:
+    SLURM_NTASKS: 8
+    SLURM_JOB_NUM_NODES: 2
+    ICON4PY_TEST_MPI_SUBCOMM_SIZE: $SLURM_NTASKS
+  rules:
+    - if: $ICON4PY_DISABLE_DRIVER_BENCHMARK == "1" || $ICON4PY_DISABLE_DRIVER_BENCHMARK == "true"
+      when: never
+    - when: on_success
```
msimberg: Need to consider whether we want both of these or keep only e.g. the two nodes/8 ranks case as sufficiently representative.

Seeing this, I also think probably the distributed bencher tests should simply be a separate pipeline that can be triggered separately.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,361 @@
...
+            process_props=process_props,
+            backend=backend,
+        )
+
+    benchmark.pedantic(
+        _with_barriers(process_props, _timed),
+        rounds=_rounds(request),
+        iterations=1,
+        warmup_rounds=_warmup_rounds(request),
+    )
```
msimberg, line 281: It'd be nice if we could get these three numbers all from a single benchmark run, but I'm not sure if that's possible.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+@pytest.mark.mpi
+@pytest.mark.benchmark
+@pytest.mark.continuous_benchmarking
+@pytest.mark.benchmark_only
+@pytest.mark.parametrize("process_props", [True], indirect=True)
```
msimberg, line 209: Most likely it'd be nicer to parameterize directly here on grid/experiment rather than put the parameterization in the pipeline file.

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -121,6 +121,41 @@ def pytest_addoption(parser: pytest.Parser):
...
+
+    with contextlib.suppress(ValueError):
+        parser.addoption(
+            "--driver-benchmark-warmup-rounds",
+            action="store",
+            type=int,
+            default=2,
+            help="Number of pytest-benchmark warmup rounds for the driver benchmark.",
+        )
+
```
msimberg: Commented elsewhere, but probably these should be parameterizations on the benchmarks themselves.

noxfile.py
```diff
@@ -39,11 +41,7 @@ class _VenvBackendKwargs(TypedDict, total=False):
 )
 NO_TESTS_COLLECTED_EXIT_CODE: Final = 5
 
-_rank = (
-    os.environ.get("PMI_RANK")
-    or os.environ.get("OMPI_COMM_WORLD_RANK")
-    or os.environ.get("SLURM_PROCID")
-)
+_rank = resolve_rank()
```
msimberg: Questionable value to tie this to icon4py.model.testing.benchmark.

noxfile.py
```diff
@@ -173,6 +171,103 @@ def __bencher_feature_branch_CI(session: nox.Session) -> None:
     )
 
 
+@nox.session(python=SUPPORTED_PYTHON_VERSIONS)
+def benchmark_driver_mpi(session: nox.Session) -> None:
```
msimberg, line 219: Commented elsewhere, but I think this is probably better off as benchmark_mpi/distributed, without the "driver" specificity (despite the driver being the only distributed benchmark at the moment).

noxfile.py
```diff
@@ -173,6 +171,103 @@ def __bencher_feature_branch_CI(session: nox.Session) -> None:
...
+            "-m",
+            "continuous_benchmarking",
+            "--benchmark-json",
+            f"pytest_benchmark_results_{session.python}_{rank}.json",
+            "tests/driver/mpi_tests/test_benchmark_driver.py",
+            *session.posargs,
+        )
+
+
+def _driver_bencher_testbed() -> str:
```
msimberg: I wonder why so much of this has to be duplicated between distributed and non-distributed bencher sessions.

msimberg (commented):
High level comments, no comments on implementation itself yet.

msimberg:
cscs-ci run benchmark_bencher

msimberg:
cscs-ci run benchmark-bencher

(correction: previous comment used underscore; the registered pipeline name is hyphen)

msimberg:
cscs-ci run benchmark-bencher

(re-trigger after merging main; shared partition removed on main)

msimberg:
cscs-ci run benchmark-bencher

(re-trigger after 4 ranks/1 node change; mpitask4 archive exists)

msimberg:
cscs-ci run benchmark-bencher

fresh retrigger 2026-09-02T16:51:32Z after 4-rank/1-node commit 1b4b726274; previous trigger did not produce a pipeline

msimberg:
cscs-ci run benchmark-bencher

stability rerun 1/2 after OMP pinning + full-node GPU layout (4 ranks / 4 GPUs / allow-task-sharing) + testbed rename 1N4R, commit 259418a8a8

msimberg:
cscs-ci run benchmark-bencher

polished-config run on 6921efda57: naming conventions, MPS disabled for dedicated-GPU-per-task layout, BACKEND leftover removed

.cscs-ci/benchmark_bencher.yml
```diff
@@ -5,26 +5,58 @@ include:
...
+  variables:
+    NOX_SESSION: "__bencher_driver_feature_branch_CI-${PYVERSION_SHORT}"
+
 benchmark_bencher_stencils_feature_aarch64:
   extends: [.test_runner_serial, .test_template_aarch64, .bencher_feature_tests]
   variables:
     TEST_SELECTION: "${STENCIL_TEST_SELECTION}"
     GT4PY_COLLECT_METRICS_LEVEL: 10
+  rules:
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
```
msimberg, line 37: TODO: here and further down. This should be removed.

.cscs-ci/benchmark_bencher.yml
```diff
@@ -5,26 +5,58 @@ include:
...
+
+benchmark_bencher_driver_mpi_feature_aarch64:
+  extends: [.bencher_driver_mpi_feature_tests, .retry_on_transient_failure]
+  variables:
+    SLURM_NTASKS: 4
+    SLURM_JOB_NUM_NODES: 1
+    ICON4PY_TEST_MPI_SUBCOMM_SIZE: $SLURM_NTASKS
+  rules:
+    - if: $ICON4PY_DISABLE_DRIVER_BENCHMARK == "1" || $ICON4PY_DISABLE_DRIVER_BENCHMARK == "true"
```
msimberg: TODO: probably doesn't need to be disableable? Or overkill to run this all the time?

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -9,6 +9,15 @@
     OMP_PLACES: cores
     OMP_NUM_THREADS: 72
 
+.benchmark_driver_base_variables:
+  variables:
+    TEST_SELECTION: "test_benchmark_driver"
+    GRID: "R02B06_GLOBAL"
```
msimberg: TODO: Is this a good grid for this? It seems to be decent based on the benchmarks in the sense that there is a bit of strong scaling, but not perfect scaling (from 1 to 4 GPUs), which means that hopefully it's a good candidate for showing if we make strong scaling improvements.

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +36,61 @@
...
     SLURM_TIMELIMIT: '01:30:00'
 
+.benchmark_driver_mpi_nox_job:
+  stage: benchmark
+  script:
+    - .cscs-ci/scripts/ci-mpi-wrapper.sh nox -s "${NOX_SESSION}" -- --backend=$BACKEND --grid=$GRID -k "${TEST_SELECTION}"
+  # These jobs use a dedicated GPU per task (ci-mpi-wrapper.sh pins one GPU
+  # per rank), so override .test_template_aarch64's before_script without
+  # start-cuda-mps.sh, which is intended for shared GPUs (multiple ranks per
+  # GPU) and would also unset CUDA_VISIBLE_DEVICES after the wrapper pinned it.
```
msimberg: Can probably be shortened...

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +36,61 @@
...
+  before_script:
+    - cd /icon4py
+    - source .venv/bin/activate # For nox etc. not for the testing venv itself
```
msimberg: Useless comment. I also think this is already in path by default in the container image?

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +36,61 @@
...
+    - source .cscs-ci/scripts/gt4py-cache.sh
+  extends: [.benchmark_driver_base_variables]
+  variables:
+    # Full-node layout: 4 ranks on 4 GPUs (ci-mpi-wrapper.sh pins one GPU per
+    # rank), 72 CPUs per rank. allow-task-sharing lets NCCL open peer GPUs for
+    # GPU-GPU communication.
```
msimberg: Shorter comment. Here and further down.

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -9,6 +9,15 @@
     OMP_PLACES: cores
     OMP_NUM_THREADS: 72
 
+.benchmark_driver_base_variables:
+  variables:
+    TEST_SELECTION: "test_benchmark_driver"
```
msimberg, line 14: TODO: matrix variable?

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -9,6 +9,15 @@
...
+    GRID: "R02B06_GLOBAL"
+    SLURM_TIMELIMIT: '02:00:00'
+    OMP_PROC_BIND: close
+    OMP_PLACES: cores
+    OMP_NUM_THREADS: 72
```
msimberg, line 18: TODO: Also set all the recommended NCCL/libfabric variables?

msimberg:
cscs-ci run benchmark-bencher

OMP sweep run 1: OMP_NUM_THREADS=64

model/driver/src/icon4py/model/driver/driver.py
```diff
@@ -797,20 +797,17 @@ def initialize_driver(
...
+def initialize_driver_states(
+    icon4py_driver: Icon4pyDriver,
+    allocator: gtx.typing.Allocator,
+) -> driver_states.DriverStates:
+    """Build and validate the initial driver states from a freshly initialized driver.
+
+    This wraps the creation of the prognostic/tracer/nonhydro/diagnostic states, the
+    application of the initial condition, the assembly of ``DriverStates``, and the
+    consistency check that precedes the time loop. It is shared between ``run_driver``
+    and the driver benchmarks.
```
msimberg: This comment shouldn't say who uses it.

model/driver/src/icon4py/model/driver/driver.py
```diff
@@ -797,20 +797,17 @@ def initialize_driver(
...
-        config=config,
-        grid_manager=grid_manager,
-        process_props=process_props,
-        backend=backend,
-    )
-    allocator = model_backends.get_allocator(backend)
+def initialize_driver_states(
+    icon4py_driver: Icon4pyDriver,
+    allocator: gtx.typing.Allocator,
+) -> driver_states.DriverStates:
```
msimberg, line 803: Even though it's only two parameters these might still be worth passing by keyword.

model/driver/tests/driver/integration_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,250 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
```
msimberg, line 1: This file should drop the single_rank suffix.

model/driver/tests/driver/integration_tests/test_benchmark_driver_single_rank.py
```diff
@@ -0,0 +1,250 @@
...
+@pytest.mark.benchmark
+@pytest.mark.continuous_benchmarking
+@pytest.mark.benchmark_only
+@pytest.mark.parametrize(
+    "driver_benchmark_experiment",
+    BENCHMARK_EXPERIMENTS,
+    indirect=True,
+    ids=[e.name for e in BENCHMARK_EXPERIMENTS],
+)
+def test_benchmark_driver_single_rank_init(
```
msimberg: This also needs to drop `single_rank`

model/driver/tests/driver/integration_tests/test_benchmark_driver_single_rank.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+@pytest.mark.benchmark
+@pytest.mark.continuous_benchmarking
+@pytest.mark.benchmark_only
+@pytest.mark.parametrize(
+    "driver_benchmark_experiment",
+    BENCHMARK_EXPERIMENTS,
+    indirect=True,
+    ids=[e.name for e in BENCHMARK_EXPERIMENTS],
```
msimberg: This is a bit silly. It should be `lambda e: e.name`.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,288 @@
...
+from icon4py.model.driver import config as driver_config, driver, driver_states, driver_utils
+from icon4py.model.testing import datatest_utils as dt_utils, definitions as test_defs, grid_utils
+from icon4py.model.testing.fixtures.datatest import backend, process_props
+
+
+if mpi_decomposition.mpi4py is None:
+    pytest.skip(
+        "Skipping parallel driver benchmark tests on single-node installation",
+        allow_module_level=True,
+    )
```
msimberg: This needs to go.

model/testing/src/icon4py/model/testing/pytest_hooks.py
```diff
@@ -187,23 +187,40 @@ def pytest_runtest_setup(item: pytest.Item) -> None:
...
+    (?:\[(?P<params>.+)\])? )       # optional parameterization
+    """,
+    re.VERBOSE,
+)
+_function_name_pattern = re.compile(
+    r"""
+    ::(?P<func>[A-Za-z_]\w*)         # trailing function name
+    (?:\[(?P<params>.+)\])?          # optional parameterization
+    $
+    """,
```
msimberg, line 206: Double-check that this makes sense.

msimberg:
@edopao @iomaganaris @yiluchen1066 @OngChia this is not ready for merging, but I'd appreciate an early quick review from any of you. Two things I'd especially appreciate feedback on is 1. the bencher setup with testbed, test naming, etc. if it makes sense to you with more bencher experience and 2. whether you think the current test seems sane.

I've used JW with R02B06 as a decent compromise between between size and time. I'm running it on one GPU and on 4 GPUs, all intra-node. For 8 ranks I bumped into the problem in #1456 which I don't want to resolve for this work yet, so staying intra-node for now. There's a small speedup, but not perfect scaling, so I'm hoping this would be a decent size to see improvements to both scaling (halo exchange overheads) and general performance improvements. The benchmarks run now init, one timestep, and the full simulation with init+timestepping as three separate benchmarks, with the full simulation taking ~1 minute (with 100 timesteps). I've yet to enable other backends than dace_gpu.

msimberg:
cscs-ci run benchmark-bencher

OMP sweep run 2: OMP_NUM_THREADS=32

msimberg:
cscs-ci run benchmark-bencher

OMP sweep run 3: OMP_NUM_THREADS=8 (MPS disabled)

msimberg:
cscs-ci run benchmark-bencher

re-run: serial job died at container startup (bwrap uid map permission denied, runner-side flake); MPI passed. Re-measuring OMP_NUM_THREADS=8.

.cscs-ci/benchmark_bencher.yml
```diff
@@ -5,26 +5,58 @@ include:
...
 
 benchmark_bencher_granules_feature_aarch64:
   extends: [.test_runner_serial, .test_template_aarch64, .bencher_feature_tests]
   variables:
     TEST_SELECTION: "${GRANULE_TEST_SELECTION}"
+  rules:
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
+
+benchmark_bencher_driver_mpi_feature_aarch64:
+  extends: [.bencher_driver_mpi_feature_tests, .retry_on_transient_failure]
```
edopao, line 47: Not directly related to this PR. I was not aware of `.retry_on_transient_failure`, interesting. I see it is used in multiple places (in other fiels as well) and I wonder if a refactoring PR could move it down to some base config such as `.test_runner_mpi` and `.test_runner_serial`.

.cscs-ci/benchmark_bencher_baseline.yml
```diff
@@ -2,15 +2,49 @@ include:
...
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
+
+benchmark_bencher_driver_mpi_baseline_aarch64:
+  extends: [.test_runner_mpi, .test_template_aarch64, .bencher_driver_mpi_baseline_tests, .retry_on_transient_failure]
+  variables:
+    SLURM_NTASKS: 4
+    SLURM_JOB_NUM_NODES: 1
+    ICON4PY_TEST_MPI_SUBCOMM_SIZE: $SLURM_NTASKS
+  rules:
+    - if: $ICON4PY_DISABLE_DRIVER_BENCHMARK == "1" || $ICON4PY_DISABLE_DRIVER_BENCHMARK == "true"
```
edopao: Is the variable `ICON4PY_DISABLE_DRIVER_BENCHMARK` only temporary? If we want a CI job that is triggered manually I think we can use `when: manual`, and there will be a "button" in the CI-Ext page to trigger it. I did this a long time ago, I don't know if it still works like that.

model/driver/tests/driver/integration_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,250 @@
...
+
+def _make_config(
+    experiment: test_defs.ExperimentDescription,
+    grid: test_defs.GridDescription,
+    process_props: decomp_defs.ProcessProperties,
+) -> driver_config.ExperimentConfig:
+    dt_utils.download_experiment(experiment, process_props)
+    experiment_path = dt_utils.get_path_for_experiment(experiment, process_props)
+    config = driver_config.read_experiment_config_from_fortran(experiment_path)
+    return config.with_overrides(
```
edopao, line 68: I guess this is where I should plugin the External workspace memory for dace benchmarks on AMD GPUs.

.cscs-ci/benchmark_bencher_baseline.yml
msimberg: Yes, temporary! For testing I only wanted the driver benchmark to be run and I haven't made up my mind if it makes sense to always run the driver benchmark on benchmark-bencher pipelines. With a warm cache it's not so expensive, but it's also not for free. Opinions? If we actually want to make it optional, we probably need a similar setup as we have now for the regular tests, i.e. something where it's possible to choose what runs in the cscs-ci run comment.

.cscs-ci/benchmark_bencher.yml
msimberg: Yeah, that might be a good idea!

FYI, I introduced it to restart jobs that take long, e.g. because of JIT compilation taking longer than expected if filesystems are unhappy. It does come with false positives, i.e. if there's an actual hang caused by our bugs (or e.g. GHEX) the job would get restarted as well, but I figured it's a decent tradeoff since infrastructure-related hangs seem to be more common.

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +35,52 @@
     SLURM_PARTITION: normal
     SLURM_TIMELIMIT: '01:30:00'
 
+.benchmark_driver_mpi_nox_job:
+  stage: benchmark
+  script:
+    - .cscs-ci/scripts/ci-mpi-wrapper.sh nox -s "${NOX_SESSION}" -- --backend=$BACKEND --grid=$GRID -k "${TEST_SELECTION}"
+  # start-cuda-mps.sh is for shared GPUs; these jobs have one GPU per task.
```
msimberg, line 42:
```suggestion
  # MPS is explicitly not enabled here. The job uses one GPU per task.
```

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +35,52 @@
...
+  before_script:
+    - cd /icon4py
+    - source .cscs-ci/scripts/gt4py-cache.sh
+  extends: [.benchmark_driver_base_variables]
+  variables:
+    # allow-task-sharing lets NCCL open peer GPUs for GPU-GPU communication.
```
msimberg, line 48:
```suggestion
```

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +35,52 @@
...
+    SLURM_GRES_FLAGS: allow-task-sharing
+  parallel:
+    matrix:
+      - BACKEND: [dace_cpu, dace_gpu, gtfn_cpu, gtfn_gpu]
+        GRID: [R02B06_GLOBAL]
+.benchmark_driver_nox_job:
+  stage: benchmark
+  script:
+    - nox -s "${NOX_SESSION}" -- --backend=$BACKEND --grid=$GRID -k "${TEST_SELECTION}"
+  # start-cuda-mps.sh is for shared GPUs; this job has one GPU per task.
```
msimberg, line 60:
```suggestion
  # MPS is explicitly not enabled here. The job uses one GPU per task.
```

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,169 @@
...
+        iterations=1,
+        warmup_rounds=BENCHMARK_WARMUP_ROUNDS,
+    )
+
+
+@pytest.mark.mpi
+@pytest.mark.benchmark
+@pytest.mark.continuous_benchmarking
+@pytest.mark.benchmark_only
+@pytest.mark.parametrize("process_props", [True], indirect=True)
```
iomaganaris: It would be nice if we could add a name to this parameter when it is passed to the test so we don't end up having `True`/`False` in the tests' names in the bencher UI without knowing what it means

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,169 @@
...
+@pytest.mark.parametrize(
+    "driver_benchmark_experiment",
+    BENCHMARK_EXPERIMENTS,
+    indirect=True,
+    ids=lambda e: e.name,
```
iomaganaris: Is the following a meaningful name `exclaim_nh35_tri_jws`? Maybe this could be only `JW`? Would be nice to get some input from the domain experts
See: https://bencher.dev/perf/icon4py?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=icon4py&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1&branches=119ef9d7-20ad-4dcc-8568-02c4f2346ead%2C01a062df-6eb7-7ae3-a028-40758214842e&heads=38bce969-9f9b-4fb2-8fcf-890a1db19ebc%2C01a068ce-909a-7fc3-b614-094c55fa425c&testbeds=8adb8393-dfbe-489f-942e-e0b0817873e5%2C01a06722-acc2-7c23-87db-a8ca254cbc81%2C01a062df-6ec5-73f2-9cc9-f8b124c4610c%2C01a0638e-cd7c-7133-bddd-05f2a7d34532%2Ce63830cd-4b31-4e08-aee8-b6dac64c6fd6%2C384b3371-df87-435f-84d4-33a3479bf21a%2Cf7b9c3c1-7f1d-4d05-abc4-326fdcc08e1f%2C565a4607-8502-4561-b990-cb89bf1ae04e&measures=fd668f25-dee7-4693-8b49-0f15f3b0d1d2&start_time=1787084975000&end_time=1789504175000&lower_boundary=false&upper_boundary=true&clear=true&tab=benchmarks&specs=%2C%2C%2C%2C%2C%2C%2C&benchmarks_search=driver&benchmarks=01a0638e-cd82-7913-8df3-9177ee34d75d%2C01a0638e-cd80-7e93-ae0d-80bb8ce3fbff%2C01a0638e-cd81-76e1-b19e-cf2b447073c9%2C01a062df-6ec9-7262-b70a-3eaa8c389d47%2C01a062df-6ec7-7701-b386-9eeeb23d6901%2C01a062df-6ec8-7de1-9df5-d95f3f229819&testbeds_search=dace_gpu

model/testing/src/icon4py/model/testing/pytest_hooks.py
iomaganaris: Looks okay [here](https://bencher.dev/perf/icon4py?utm_medium=referral&utm_source=github&utm_content=comment&utm_campaign=pr%2Bcomments&utm_term=icon4py&key=true&reports_per_page=4&branches_per_page=8&testbeds_per_page=8&benchmarks_per_page=8&plots_per_page=8&reports_page=1&branches_page=1&testbeds_page=1&benchmarks_page=1&plots_page=1&branches=119ef9d7-20ad-4dcc-8568-02c4f2346ead%2C01a062df-6eb7-7ae3-a028-40758214842e&heads=38bce969-9f9b-4fb2-8fcf-890a1db19ebc%2C01a068ce-909a-7fc3-b614-094c55fa425c&testbeds=8adb8393-dfbe-489f-942e-e0b0817873e5%2C01a06722-acc2-7c23-87db-a8ca254cbc81%2C01a062df-6ec5-73f2-9cc9-f8b124c4610c%2C01a0638e-cd7c-7133-bddd-05f2a7d34532%2Ce63830cd-4b31-4e08-aee8-b6dac64c6fd6%2C384b3371-df87-435f-84d4-33a3479bf21a%2Cf7b9c3c1-7f1d-4d05-abc4-326fdcc08e1f%2C565a4607-8502-4561-b990-cb89bf1ae04e&measures=fd668f25-dee7-4693-8b49-0f15f3b0d1d2&start_time=1787084975000&end_time=1789504175000&lower_boundary=false&upper_boundary=true&clear=true&tab=benchmarks&specs=%2C%2C%2C%2C%2C%2C%2C&benchmarks_search=driver&benchmarks=01a0638e-cd82-7913-8df3-9177ee34d75d%2C01a0638e-cd80-7e93-ae0d-80bb8ce3fbff%2C01a0638e-cd81-76e1-b19e-cf2b447073c9%2C01a062df-6ec9-7262-b70a-3eaa8c389d47%2C01a062df-6ec7-7701-b386-9eeeb23d6901%2C01a062df-6ec8-7de1-9df5-d95f3f229819&testbeds_search=dace_gpu). I suppose though the run shown there is from a earlier version of the code

noxfile.py
```diff
@@ -147,27 +183,164 @@ def __bencher_feature_branch_CI(session: nox.Session) -> None:
...
+    In MPI runs only rank 0 uploads; in single-rank runs the resolved rank is
+    treated as rank 0.
+    """
+    return rank is None or rank == 0
+
+
+@nox.session(python=SUPPORTED_PYTHON_VERSIONS)
+def benchmark_driver_mpi(session: nox.Session) -> None:
+    """Run the distributed driver benchmark under MPI."""
+    _install_session_venv(session, extras=["all"], groups=["test"])
```
iomaganaris, line 229: Should we make sure that this is done by a single rank? Same for the other calls of this function in `mpi` tests/benchmarks

noxfile.py
```diff
@@ -147,27 +183,164 @@ def __bencher_feature_branch_CI(session: nox.Session) -> None:
...
+        return
+
+    session.run(
+        *_bencher_baseline_command(
+            f"model/driver/pytest_benchmark_results_{session.python}_{rank}.json"
+        ),
+        env=_bencher_baseline_env(_driver_mpi_bencher_testbed()),
+        external=True,
+        silent=True,
+    )
```
iomaganaris, line 275: Should we do some preprocessing here and update the maximum runtime across all ranks for example? Same for the feature branches below

noxfile.py
```diff
@@ -147,27 +183,164 @@ def __bencher_feature_branch_CI(session: nox.Session) -> None:
     """
     bencher_testbed = f"{os.environ['RUNNER']}:{os.environ['SYSTEM_TAG']}:{os.environ['BACKEND']}:{os.environ['GRID']}"
```
iomaganaris: Should this be replaced by `_driver_bencher_testbed` or a more generic function? It can also be reused in `_driver_mpi_bencher_testbed`

iomaganaris (commented):
Overall LGTM. Some suggestions to improve the bencher readability, avoid duplicate definitions and possible bugs

noxfile.py
msimberg: Yep! Already changed this locally, have not pushed that change.

noxfile.py
msimberg: There are barriers at the start and end of each benchmark, which should already give a relatively accurate timing with all ranks effectively reporting the max (though only rank 0 actually reports the timing). Do you think that's sufficient?

noxfile.py
msimberg: Not right now at least (if you're referring to the venv installation?). It's already like this for the existing MPI tests because getting nox to synchronize different ranks with only one rank actually doing the installation is more hassle than it's worth (or at least was last time I tried). All ranks get a unique venv to avoid any need for synchronization (at the cost of more symlinks).

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
msimberg: It's a fair question, but I think if we change it we should rather change the proper name in https://github.com/C2SM/icon4py/blob/11cddd359161bc56d53cefda43371c7a8ac65f2b/model/testing/src/icon4py/model/testing/definitions.py#L229. I think @yiluchen1066, @OngChia, @jcanton can answer better if it's a meaningful name.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
msimberg: Hmm, true. This is a more general pattern that we have for other tests and basically just means "initialize process props" with MPI. I'd like to avoid a bigger refactoring of that here, but we can easily add a label for the `True` (it would more or less become "distributed" or "multirank" or something in that direction. Technically it just means if MPI is initialized (it could be using only one rank), but I think the "distributed" meaning is the more interesting one. The number of nodes and ranks is anyway in the testbed name, so we don't need to replicate that in this parameter.

iomaganaris: I agree it's a bigger refactoring so it can take place in a separate PR. Not sure if the naming of the tests that have `distributed` in them should also change then

noxfile.py
iomaganaris: I think you're right. That's sufficient. When I run manually the driver from CLI I get in general multiple outputs with different timings. The differences are all in between ~0.5s but there's still some. I guess this is within the variation we can expect across multiple ranks

noxfile.py
iomaganaris, line 247: I think that this can also call `_serial_testbed` + the extra suffix?

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
OngChia: The name comes from the run script's name in ICON repo. Originally, the run scripts' name were simply the experiment name. I remember we later added exclaim in front of them to indicate these are the run scripts we used for generating the serialized data. nh35 means nonhydrostatic 35 levels..... 

I agree with renaming it. Maybe it is still good to keep exclaim in the run script? But we definitely can remove "exclaim_" in our icon4py
I think it can be renamed either JWS or jablonowski_williamson_steady (there is another experiment with baroclinic perturbation, so it is useful to keep the "S")

.cscs-ci/benchmark_bencher_common.yml
msimberg: Set in base.yml.

OngChia (commented):
Looks nice to me. Just wondering we also have a Jablonowski-Williamson unsteady test with initial perturbation. Does the performance benchmark depend on the complexity of calculations? (in steady state simulation, the tendencies from dycore are close to zero over the first few days).

model/driver/tests/driver/integration_tests/test_driver.py
```diff
@@ -68,7 +68,7 @@
 # Metadata selecting the MCH mid-time-step dynamics savepoints (see the MCH branch in
 # the test body): solve-nonhydro exit at the corrector (istep=2) of the last substep
 # (2 for MCH), and the non-initial diffusion savepoint. Only instantiated for MCH.
-@pytest.fixture  # type: ignore[no-redef]  # deliberately shadows the fixtures.py import
+@pytest.fixture
```
msimberg: This change seems unnecessary or a leftover.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,177 @@
...
+
+from ..fixtures import (
+    BENCHMARK_EXPERIMENTS,
+    BENCHMARK_ROUNDS,
+    BENCHMARK_WARMUP_ROUNDS,
+    driver_benchmark_config,
+    driver_benchmark_experiment,
+    driver_benchmark_grid,
+    driver_benchmark_grid_manager,
+)
```
msimberg, line 30: Could these not be in the single-rank test_benchmark_driver.py test?

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
msimberg: Ok to leave it unchanged now? I don't have anything against renaming it, but I'm not sure what's a better name (and it might need changes in the serialized data archives etc. etc.)

model/driver/tests/driver/fixtures.py
```diff
@@ -5,33 +5,95 @@
...
-    savepoint_velocity_init,
-    step_date_exit,
-    step_date_init,
-    topography_savepoint,
-)
+from icon4py.model.common import model_backends, time
+from icon4py.model.common.decomposition import definitions as decomp_defs
+from icon4py.model.common.grid import grid_manager as gm
+from icon4py.model.driver import config as driver_config, driver_utils
+from icon4py.model.testing import datatest_utils as dt_utils, definitions as test_defs, grid_utils
```
msimberg, line 17: This looks like an unnecessary change.

model/driver/tests/driver/fixtures.py
```diff
@@ -5,33 +5,95 @@
...
+
+
+BENCHMARK_EXPERIMENTS: list[test_defs.ExperimentDescription] = [test_defs.Experiments.JW]
+BENCHMARK_STEPS: int = 100
+BENCHMARK_ROUNDS: int = 5
+BENCHMARK_WARMUP_ROUNDS: int = 2
```
msimberg, line 51: These and all the driver benchmark-related fixtures seem better off closer to the benchmark files.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
OngChia: Agree. Let's leave it unchanged now and open another PR for renaming these experiments when necessary.

model/testing/src/icon4py/model/testing/grid_utils.py
```diff
@@ -28,6 +30,41 @@
...
+                grid = maybe_grid
+                break
+    if grid is None:
+        raise pytest.UsageError(
+            f"Unknown grid '{name}' in '--grid' option. "
+            f"Use a preset, a 'Grids' attribute name, or a grid description name."
+        )
+    return grid
+
+
```
msimberg: This seems a bit suspicious. Isn't `icon_regional` also used in some benchmarks. And `icon_global`/`icon_regional` are also "benchmark" grids...

model/testing/tests/testing/unit_tests/test_grid_utils.py
```diff
@@ -0,0 +1,43 @@
...
+    def test_unknown_name_raises_usage_error(self) -> None:
+        with pytest.raises(pytest.UsageError, match="Unknown grid 'no_such_grid'"):
+            grid_utils.resolve_grid_description("no_such_grid")
+
+    def test_custom_presets_override_default(self) -> None:
+        custom = {"custom": test_defs.Grids.R02B04_GLOBAL}
+        assert (
+            grid_utils.resolve_grid_description("custom", presets=custom)
+            is test_defs.Grids.R02B04_GLOBAL
+        )
```
msimberg: Is this worth it if the `resolve_grid_description` function is simplified?

msimberg:
> Looks nice to me. Just wondering we also have a Jablonowski-Williamson unsteady test with initial perturbation. Does the performance benchmark depend on the complexity of calculations? (in steady state simulation, the tendencies from dycore are close to zero over the first few days).

I'm not sure. Unless there's conditional computation based on the values then performance shouldn't be affected by them being close to zero? We can also expand this to other experiments over time.

.cscs-ci/benchmark_bencher.yml
msimberg: FYI, I removed it from here for now (only the unit tests use the retry). I'm not sure what the effects of restarting the benchmarks would be...

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
msimberg: FYI, I added the id "distributed" to the distributed benchmarks in this PR, but did not touch others.

noxfile.py
msimberg: Good idea, done.

.cscs-ci/benchmark_bencher.yml
msimberg:
```suggestion
```

.cscs-ci/benchmark_bencher.yml
```diff
@@ -28,3 +38,15 @@ benchmark_bencher_granules_feature_aarch64:
   extends: [.test_runner_serial, .test_template_aarch64, .bencher_feature_tests]
   variables:
     TEST_SELECTION: "${GRANULE_TEST_SELECTION}"
+  rules:
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
```
msimberg:
```suggestion
```

.cscs-ci/benchmark_bencher_baseline.yml
```diff
@@ -2,15 +2,41 @@ include:
...
+    NOX_SESSION: "__bencher_driver_baseline_CI-${PYVERSION_SHORT}"
+
 benchmark_bencher_stencils_baseline_aarch64:
   extends: [.test_runner_serial, .test_template_aarch64, .benchmark_nox_job]
   variables:
     NOX_SESSION: "__bencher_baseline_CI-${PYVERSION_SHORT}"
     TEST_SELECTION: "${STENCIL_TEST_SELECTION}"
     GT4PY_COLLECT_METRICS_LEVEL: 10
+  rules:
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
```
msimberg:
```suggestion
```

.cscs-ci/benchmark_bencher_baseline.yml
```diff
@@ -14,3 +26,15 @@ benchmark_bencher_granules_baseline_aarch64:
   variables:
     NOX_SESSION: "__bencher_baseline_CI-${PYVERSION_SHORT}"
     TEST_SELECTION: "${GRANULE_TEST_SELECTION}"
+  rules:
+    - when: never  # TEMPORARY: disabled for driver bencher trial; restore before merge
```
msimberg:
```suggestion
  rules:
```

OngChia:
> > Looks nice to me. Just wondering we also have a Jablonowski-Williamson unsteady test with initial perturbation. Does the performance benchmark depend on the complexity of calculations? (in steady state simulation, the tendencies from dycore are close to zero over the first few days).
> 
> I'm not sure. Unless there's conditional computation based on the values then performance shouldn't be affected by them being close to zero? We can also expand this to other experiments over time.

If that is the case. I think it should be fine with JWS experiment. Both unsteady and steady use the same set of algorithms.

msimberg:
cscs-ci run benchmark-bencher

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +34,52 @@
...
+  variables:
+    # allow-task-sharing lets NCCL open peer GPUs for GPU-GPU communication.
+    SLURM_GPUS_PER_NODE: 4
+    SLURM_CPUS_PER_TASK: 72
+    SLURM_GRES_FLAGS: allow-task-sharing
+  parallel:
+    matrix:
+      - BACKEND: [dace_cpu, dace_gpu, gtfn_cpu, gtfn_gpu]
+        GRID: [R02B06_GLOBAL]
+.benchmark_driver_nox_job:
```
edopao, line 55: Only a nit comment. Most other configurations are separate by a blank line.
```suggestion

.benchmark_driver_nox_job:
```

.cscs-ci/benchmark_bencher_common.yml
```diff
@@ -27,6 +34,52 @@
...
+    - source .cscs-ci/scripts/gt4py-cache.sh
+  extends: [.benchmark_driver_base_variables]
+  variables:
+    SLURM_GPUS_PER_TASK: 1
+    SLURM_CPUS_PER_TASK: 72
+  parallel:
+    matrix:
+      - BACKEND: [dace_cpu, dace_gpu, gtfn_cpu, gtfn_gpu]
+        GRID: [R02B06_GLOBAL]
+.bencher_github_event_setup:
```
edopao, line 71:
```suggestion

.bencher_github_event_setup:
```

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
```diff
@@ -0,0 +1,158 @@
...
+            grid_manager=grid_manager,
+            process_props=props,
+            backend=bench_backend,
+        )
+
+    benchmark.pedantic(
+        _with_barriers(process_props, _timed),
+        setup=_setup,
+        rounds=BENCHMARK_ROUNDS,
+        iterations=1,
```
edopao, line 89: You maybe already discussed it, but I do not understand the setting rounds/iterations. Each round will average over iterations. I would rather have one round with X=10? iterations.

noxfile.py
edopao, line 357: Should we include the marker `not benchmark_only` here too? I see you used it in MPI tests.
```suggestion
            *f"pytest -sv -n {os.environ.get('NUM_PROCESSES', 'auto')}".split(),
            "--benchmark-disable",
            "-k",
            "not benchmark_only",            
```

edopao, line 341: These two comments seem obsolete. Worth trying to switch to `extras=["all"]` as you did for MPI tests?

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
iomaganaris: I think the difference between rounds and iterations is related to the timers used. IIUC there is one timer/data point per `round`. If `iterations` is more than 1 then the time reported for each round will be the average of them. The idea behind using iterations is that if there is a very short function timed then we get a more stable measurement with multiple iterations because we lower the influence of timer overheads/resolution.
Currently `bencher` reports the `mean` of all rounds reported by pytest-benchmark but we could change that in the future and measure the `median` for example

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
iomaganaris: I don't think it's possible out of the box but it should be possible with some customization. Maybe it could be part of another PR.

model/driver/tests/driver/mpi_tests/test_benchmark_driver.py
edopao: OK, so Bencher is reporting the average over rounds. That should be good enough, then.

PR #1452, opened 2026-09-01
Move shared stencils and the diagnostic calculations into common

havogt:
Split 1 of C2SM/icon4py#1359 ("Port tmx turbulence"). This PR carries **only the changes that
are independent of the tmx port**: relocations of code that already exists on `main`.
No tmx code, no new operators.

The criterion, applied file by file: a change is in this PR if it moves or renames
something that already exists on `main`; it stays with the port if the port introduces
it.


model/common/src/icon4py/model/common/physics/stencils/compute_hydrostatic_pressure.py
```diff
@@ -54,7 +54,7 @@ def _diagnose_pressure(
 
 
 @gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
-def diagnose_pressure(
+def compute_hydrostatic_pressure(
```
havogt: do we actually need all the programs or are we actually only using the field_operators?

model/common/src/icon4py/model/common/math/stencils/init_cell_kdim_field_with_zero_wp.py
```diff
@@ -14,7 +14,14 @@
...
 def _init_cell_kdim_field_with_zero_wp() -> fa.CellKField[wpfloat]:
-    """Formerly known as _mo_solve_nonhydro_stencil_57 or _mo_solve_nonhydro_stencil_64."""
+    """
+    Set a cell K field to zero.
+
+    Formerly known as _mo_solve_nonhydro_stencil_57 or
+    _mo_solve_nonhydro_stencil_64; also ports the ``CALL init(...)``
+    (mo_fortran_tools) zero fills, e.g. of the tracer and energy tendencies in
+    the tmx scalar diffusion stages (mo_vdf.f90).
+    """
```
havogt:
```suggestion
```

model/common/src/icon4py/model/common/physics/thermodynamics.py
```diff
@@ -0,0 +1,209 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
+#
+# Copyright (c) 2022-2024, ETH Zurich and MeteoSwiss
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+"""
+AES thermodynamic helper functions shared across physics parameterizations.
```
havogt: Let's rewrite this docstring, avoiding "AES" or we should introduce a different package somewhere that is AES specific.

havogt: or as a workaround for now, call this module aes_thermodynamics?

havogt (commented):
@OngChia 
- Can you double check if any additions to docstrings/comments make sense?
- Should we combine some of the operators/programs into fewer modules?

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
```diff
@@ -534,9 +534,9 @@ def __init__(
         # stencils
         self._compute_edge_tangential = model_options.setup_program(
```
OngChia: Since this stencil is used for computing tangential wind from the averaged `vn` in tracer advection. Better to have the same name as the gt4py program here.

model/atmosphere/tracer_advection/src/icon4py/model/atmosphere/tracer_advection/tracer_advection_horizontal.py
```diff
@@ -604,12 +604,12 @@ def _compute_numerical_flux(
         ## tracer-independent part
 
         # compute tangential velocity
-        log.debug("running stencil compute_edge_tangential - start")
+        log.debug("running stencil compute_tangential_wind_wp - start")
         self._compute_edge_tangential(
```
OngChia: Similarly here.

model/common/src/icon4py/model/common/interpolation/stencils/interpolate_to_cell_center_vp.py
```diff
@@ -14,18 +14,22 @@
...
-def _interpolate_to_cell_center(
+def _interpolate_to_cell_center_vp(
     interpolant: fa.EdgeKField[vpfloat],
     e_bln_c_s: gtx.Field[gtx.Dims[dims.CellDim, dims.C2EDim], wpfloat],
 ) -> fa.CellKField[vpfloat]:
-    """Formerly known as mo_velocity_advection_stencil_08 or mo_velocity_advection_stencil_09."""
+    """Interpolate an edge field to the cell centers with the bilinear C2E weights.
+
+    Formerly known as mo_velocity_advection_stencil_08 or
+    mo_velocity_advection_stencil_09.
```
OngChia: I think we can remove this comment.
```suggestion
```

model/common/src/icon4py/model/common/interpolation/stencils/compute_tangential_wind.py
```diff
@@ -0,0 +1,92 @@
...
+
+@gtx.field_operator
+def _compute_tangential_wind(
+    vn: fa.EdgeKField[wpfloat],
+    rbf_vec_coeff_e: gtx.Field[gtx.Dims[dims.EdgeDim, dims.E2C2EDim], wpfloat],
+) -> fa.EdgeKField[vpfloat]:
+    """
+    Variable-precision variant of ``_compute_tangential_wind_wp``.
+
+    Formerly known as _mo_velocity_advection_stencil_01.
```
OngChia: Remove this comment?
```suggestion
```

model/common/src/icon4py/model/common/interpolation/stencils/compute_tangential_wind.py
```diff
@@ -0,0 +1,92 @@
...
+
+    The stencil is agnostic of the vertical staggering: it can be applied to
+    half-level input (e.g. vn_ie -> vt_ie) as well as full-level input
+    (vn -> vt); the K extent is just a domain argument.
+    """
+    return neighbor_sum(rbf_vec_coeff_e * vn(E2C2E), axis=dims.E2C2EDim)
+
+
+@gtx.field_operator
+def _compute_tangential_wind(
```
OngChia: Do we add `_vp` to the name just to be consistent with the naming of some other similar stencils that also have `_vp`?
```suggestion
def _compute_tangential_wind_vp(
```

model/common/src/icon4py/model/common/physics/pressure.py
```diff
@@ -25,11 +25,11 @@
 from typing import TYPE_CHECKING
 
 from icon4py.model.common import dimension as dims, type_alias as ta
-from icon4py.model.common.diagnostic_calculations.stencils import (
-    diagnose_pressure as diagnose_pressure_stencil,
-    diagnose_surface_pressure as diagnose_surface_pressure_stencil,
-)
 from icon4py.model.common.grid import horizontal as h_grid
+from icon4py.model.common.physics.stencils import (
+    compute_hydrostatic_pressure as diagnose_pressure_stencil,
```
OngChia: Better to just use ether compute_*** because all other thermodynamics stencils are already changed from diagnose_*** to compute_***

model/common/src/icon4py/model/common/physics/pressure.py
OngChia:
```
def compute_pressure_surface_to_top(
```
Furthermore, is it possible to combined these two stencils? I ported these two long time ago, and I remember I could not put the scan and the field operator that computes the surface pressure together. Now with new features in gt4py, I wonder if we can combine them together, as I think they come in set (you can't compute hydrostatic pressure without first computing surface pressure by extrapolation).
surface_to_top is a strange name. If we can combine them, we can simply rename the combined stencil as `compute_hydrostatic_pressure`, or if longer name is allowed, for example, `compute_surface_and_hydrostatic_pressure`

OngChia: `compute_pressure_surface_to_top_ndarray(`
or `compute_surface_and_hydrostatic_pressure_ndarray(`

OngChia (commented):
I looked at the tridiagonal (forward and backward) stencils in dycore and tmx, the only difference is precision vpfloat and wpfloat. Does it make sense if we move them to `math/stencils` in `common` in this PR and , again, have _wp and _vp two versions?

model/common/src/icon4py/model/common/physics/thermodynamics.py
OngChia: Sorry for late reply. After looking more carefully into the fortran thermodynamics computations code and icon4py together with @yiluchen1066, we find that:

1. Although the all computations here are only used by aes (and I also checked the Fortran code that these functions are only used by aes), the formula of the internal energy is a general physical relationship that can be used by any physics in the future. I prefer to remove any "aes" from all docstring in `common/physics/`.
2. Secondly, we should group the stencils here into different categories like what is done in tmx_turbulence stencils. Suggestion: we can have four modules under `common/physics/thermodynamics`: `compute_temperature.py` (contain all stencils that output temperature-related variables), `compute_pressure.py`, `compute_moisture.py`, and `compute_energy.py` (leave this `compute_brunt_vaisala_frequency.py` alone in `common/physics` as we have no idea where this belongs to. In fortran code, it is only used by all turbulence schemes).
3. And we realized that there are also some thermodynamics functions (in numpy) sitting in `/common/thermodynamic_functions.py`. They probably should also be moved inside `common/physics/thermodynamics/compute_moisture.py` and `common/physics/thermodynamics/compute_pressure.py`). Remove unnecessary comments in the docstring for example, `Mirrors ``spec_humi`` in ``mo_thdyn_functions.f90``:` in specific_humidity.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro_stencils.py
```diff
@@ -10,14 +10,14 @@
...
     _update_density_exner_wind,
 )
 from icon4py.model.atmosphere.dycore.stencils.update_wind import _update_wind
 from icon4py.model.common import dimension as dims, field_type_aliases as fa
+from icon4py.model.common.math.stencils.init_cell_kdim_field_with_zero_wp import (
+    _init_cell_kdim_field_with_zero_wp,
+)
 
 
 @gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
```
yiluchen1066, line 21: Minor thing: shall we move these two stencils in this file also to ```/dycore/stencils```, with renaming the ```stencil_61_62```? and also group all stencils into different modules like we suggested for tmx stencils. But we should also do it another pr probably?

model/common/src/icon4py/model/common/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -51,7 +51,9 @@ class JablonowskiWilliamsonConfig:
     # reads zp_ape from the nh_testcase_nml
     # The default values are from mo_nh_jabw_exp.f90 and mo_nh_testcases_nml.f90
     p_sfc: float = 100000.0
-    # amplitude of the u-perturbation [m/s] (jw_up); jabw_s resets it to 0.0.
+    # amplitude of the u-perturbation [m/s] (jw_up). Default matches the ICON
```
yiluchen1066: the original comment should be enough or make it simpler 
```suggestion
    # amplitude of the u-perturbation [m/s] (jw_up); jabw_s resets it to 0.0 (matches the ICON namelist default).
```

model/common/src/icon4py/model/common/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -68,9 +70,11 @@ class JablonowskiWilliamsonConfig:
     # number of iterations to converge qv against the moisture-dependent
     # temperature (Fortran l_rediag=.TRUE. => 10 iterations).
     moisture_init_iterations: int = 10
-    # target column-integrated moisture for APE cases [kg/m**2] (ztmc_ape).
+    # target column-integrated moisture [kg/m**2] (ztmc_ape). Only used by the
```
yiluchen1066: same as above

model/common/src/icon4py/model/common/initial_condition/analytical/jablonowski_williamson.py
```diff
@@ -68,9 +70,11 @@ class JablonowskiWilliamsonConfig:
...
+    # APE test cases, which rescale qv to match this global value.
     global_moisture_content: float = 25.006
-    # rescale qv to global_moisture_content (APE only; Fortran opt_global_moist).
+    # whether to rescale qv to global_moisture_content; enabled for the APE
```
yiluchen1066: same as above

yiluchen1066 (commented):
LGTM! left a few minor things. talked and agreed offline with @OngChia about the stencils part

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+    rho: fa.CellKField[ta.wpfloat],
+    dz: fa.CellKField[ta.wpfloat],
+) -> fa.CellKField[ta.wpfloat]:
+    """
+    Compute the temperature from the internal energy
+
+    Args:
+        u:                  Internal energy (extensive)
+        qv:                 Water vapor specific humidity
+        qliq:               Specific mass of liquid phases
```
OngChia:
```suggestion
        qliq:               Specific liquid-phase content
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+        qice:               Specific mass of solid phases
```
OngChia:
```suggestion
        qice:               Specific ice-phase content
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+
+import gt4py.next as gtx
+
+from icon4py.model.common import field_type_aliases as fa, type_alias as ta
+from icon4py.model.common.constants import PhysicsConstants
+from icon4py.model.common.type_alias import wpfloat
+
+
+@gtx.field_operator(grid_type=gtx.GridType.UNSTRUCTURED)
+def compute_temperature_from_internal_energy(  # noqa: PLR0917 [too-many-positional-arguments]
```
OngChia: Because it computes and returns temperature, could you move this to `compute_temperature.py`?

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Compute the temperature from the internal energy
+
+    Args:
+        u:                  Internal energy (extensive)
+        qv:                 Water vapor specific humidity
+        qliq:               Specific mass of liquid phases
+        qice:               Specific mass of solid phases
+        rho:                Ambient density
+        dz:                 Extent of grid cell
```
OngChia: I feel that this is better
```suggestion
        dz:                 Vertical grid spacing
```
or
```suggestion
        dz:                 Vertical extent of grid cell
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+        )
+        * rho
+        * dz
+    )  # Moist isometric specific heat
+
+    return (u + rho * dz * (qliq * PhysicsConstants.lvc + qice * PhysicsConstants.lsc)) / cv
+
+
+@gtx.field_operator
+def compute_temperature_from_internal_energy_scalar(  # noqa: PLR0917 [too-many-positional-arguments]
```
OngChia: Similarly, move this to `compute_temperature.py`.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Compute the temperature from the internal energy (scalar version callable from scan_operator)
+
+    Args:
+        u:                  Internal energy (extensive)
+        qv:                 Water vapor specific humidity
+        qliq:               Specific mass of liquid phases
+        qice:               Specific mass of solid phases
+        rho:                Ambient density
+        dz:                 Extent of grid cell
```
OngChia: Consider the suggestion in my comment above about `dz`

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+    """
+    Compute the internal energy from the temperature
+
+    Args:
+        t:                 Temperature
+        qv:                Specific mass of vapor
+        qliq:              Specific mass of liquid phases
+        qice:              Specific mass of solid phases
+        rho:               Ambient density
+        dz:                Extent of grid cell
```
OngChia: Consider the suggestion in my comment above about `dz`

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
```
OngChia: Consider the suggestion in my comment above about `dz`

model/common/src/icon4py/model/common/physics/thermodynamics/compute_temperature.py
```diff
@@ -6,6 +6,10 @@
 # Please, refer to the LICENSE file in the root directory.
 # SPDX-License-Identifier: BSD-3-Clause
 
+"""Virtual temperature and temperature diagnostics, and the equation-of-state-consistent
+update of exner and theta_v from a virtual-temperature tendency.
+"""
+
```
OngChia:
```suggestion
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,261 @@
...
+
+
+@gtx.field_operator
+def _compute_hydrostatic_pressure(
+    ddqz_z_full: fa.CellKField[ta.wpfloat],
+    virtual_temperature: fa.CellKField[ta.wpfloat],
+    surface_pressure: gtx.Field[gtx.Dims[dims.CellDim], ta.wpfloat],
+) -> tuple[fa.CellKField[ta.wpfloat], fa.CellKField[ta.wpfloat]]:
+    """
+    Update pressure by assuming hydrostatic balance (dp/dz = -rho g = p g / Rd / Tv).
```
OngChia:
```suggestion
    Compute the hydrostatic pressure using the hydrostatic balance equation dp/dz = -rho g = p g / Rd / Tv, which is different from the total pressure derived from the Exner function ``P0REF * exner ** CPD_O_RD`` because the latter also contains the non-hydrostatic component.
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,261 @@
...
+The surface pressure is extrapolated from the lowest three levels, then the
+pressure is obtained by vertical integration of the virtual temperature;
+``compute_surface_and_hydrostatic_pressure`` sequences the two stencils the way
+ICON's ``diagnose_pres_temp`` does.
+
+This is the *hydrostatic* pressure, deliberately not the Exner-function shortcut
+``P0REF * exner ** CPD_O_RD``. The moist initial conditions rely on the
+hydrostatic value because ICON re-diagnoses pressure via ``diagnose_pres_temp``
+when initializing the water vapour (``init_nh_inwp_tracers`` with
+``l_rediag=.TRUE.``), so the converged/serialized state matches this integration.
```
OngChia:
```suggestion
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,261 @@
...
+#
+# Copyright (c) 2022-2024, ETH Zurich and MeteoSwiss
+# All rights reserved.
+#
+# Please, refer to the LICENSE file in the root directory.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Hydrostatic pressure, and the saturation vapour pressures.
+
+The surface pressure is extrapolated from the lowest three levels, then the
```
OngChia: We can move this comment to the docstring of compute_surface_pressure stencil.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,261 @@
...
+pressure is obtained by vertical integration of the virtual temperature;
+``compute_surface_and_hydrostatic_pressure`` sequences the two stencils the way
+ICON's ``diagnose_pres_temp`` does.
```
OngChia: Remove this?
```suggestion
```

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
```diff
@@ -0,0 +1,164 @@
...
+Moist internal energy and its inverse, the temperature.
+
+The internal energy of a grid cell follows from the specific heats of dry air,
+vapour, liquid and ice weighted by their mass fractions, minus the latent heats
+of the condensed phases; the inverse relation recovers the temperature from it.
+
+The constants come from :class:`icon4py.model.common.constants.PhysicsConstants`
+(a ``wpfloat``-based ``enum.Enum``, which is what GT4Py/gtfn needs for symbols
+referenced inside field operators).
+"""
```
OngChia:
```suggestion
```
Until now, I did not notice that the internal energy defined in these stencils is actually **`internal energy per unit area`**.
I do not know whether we should change the variable name `internal_energy` to `internal_energy_per_area`, although we already have variables with a much longer name. How do you think?
At a minimum, change `internal energy` to `internal energy per unit area` in all docstring in this module.

OngChia (commented):
Another round. Now, mainly some comments about docstring

havogt:
cscs-ci run default

havogt:
cscs-ci run default

model/common/src/icon4py/model/common/physics/thermodynamics.py
havogt: Done as proposed, with one addition: a fifth module, `compute_tendencies.py`, because the three tendency stencils fit none of the four quantities and muphys imports them as a unit. The two temperature-from-internal-energy operators are now in `compute_temperature.py` as asked in the follow-up. `compute_brunt_vaisala_frequency` is not in this PR — it arrives with the tmx port.

model/common/src/icon4py/model/common/physics/pressure.py
havogt: Combined in ce36f6dad: the surface pressure is written into the bottom row of `pressure_ifc` first, and the scan reads that row through `pressure_ifc(dims.KDim + 0.5)` in its first iteration, so the `CellField` and the host-side copies are gone. `compute_surface_pressure` is deleted; muphys and the driver call the one program. What gt4py still refuses is one call with a K and a KHalf output (the TODO in the program), so it is three calls inside one program.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_energy.py
havogt: Docstrings now say "internal energy per unit area" throughout (the formula is rho*dz*(...), i.e. J/m^2), and the name follows: `compute_internal_energy_per_area[_scalar]`, `compute_temperature_from_internal_energy_per_area[_scalar]`, with the inverse's input parameter `internal_energy_per_area`. The module docstring is deleted as suggested; its one real constraint (constants must come from a `wpfloat`-based Enum for gtfn to accept them inside field operators) now sits as a comment on the `PhysicsConstants` class.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro_stencils.py
havogt: Agreed, and as a separate PR as you suggest: it touches `solve_nonhydro.py` / `solve_nonhydro_stencils.py`, which this PR already modifies, and the regroup spans all 72 dycore stencil modules. Queued for after this merges.

havogt:
On moving the tridiagonal sweeps into `common/math/stencils` (review comment above): I compared the two forward sweeps and they differ in more than precision. dycore stores the negated off-diagonal (`c_new = (0 - c) / (b + a*c_prev)`), the tmx version stores standard Thomas (`c' = c / (b - a*c'_prev)`), so a merge has to pick one convention and change the matching back-substitution. That is new work rather than a relocation, so I would keep it out of this PR and do it on its own after the tmx port lands.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: Took your sentence, with one correction and a wrap: with rho = p/(Rd Tv) the hydrostatic balance reads dp/dz = -rho g = -p g/(Rd Tv) — the minus was missing — and the line is wrapped to the guideline width.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,282 @@
...
+            dims.KHalfDim: (vertical_start, vertical_end + 1),
+        },
+    )
+
+
+if TYPE_CHECKING:
+    import gt4py.next.typing as gtx_typing
+
+    from icon4py.model.common.grid import base as grid_base
+
```
havogt: top of the file

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,282 @@
...
+    _pressure_on_half_levels(
+        pressure_ifc_on_model_levels,
+        surface_pressure,
+        vertical_end,
+        out=pressure_ifc,
+        domain={
+            dims.CellDim: (horizontal_start, horizontal_end),
+            dims.KHalfDim: (vertical_start, vertical_end + 1),
+        },
+    )
```
havogt, line 147: Is it correct that this merge is blocked because of the comment in `_pressure_on_half_levels`?

havogt:
cscs-ci run default

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,282 @@
...
+    Note that virtual temperature is used in the equation.
+
+    Args:
+        ddqz_z_full: vertical grid spacing at full levels [m]
+        virtual_temperature: air virtual temperature [K]
+        surface_pressure: surface air pressure [Pa]
+    Returns:
+        pressure at full levels, pressure at the bounding upper interface of each level
+    """
+    pressure, pressure_ifc, _ = _scan_pressure(ddqz_z_full, virtual_temperature, surface_pressure)
```
havogt: It feels like we can actually call the surface_pressure computation from here. Or is there a problem with k range in the scan?

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,282 @@
...
+            ddqz_z_full(dims.KHalfDim - 0.5) / virtual_temperature(dims.KHalfDim - 0.5)
+            + ddqz_z_full(dims.KHalfDim - 1.5) / virtual_temperature(dims.KHalfDim - 1.5)
+            + 0.5 * ddqz_z_full(dims.KHalfDim - 2.5) / virtual_temperature(dims.KHalfDim - 2.5)
+        )
+    )
+    return surface_pressure
+
+
+@gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
+def compute_surface_pressure(  # noqa: PLR0917 [too-many-positional-arguments]
```
havogt: is this program used on its own? If not delete.

havogt:
cscs-ci run default

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: Moved to the top.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: Only the one-call form: an operator with a K and a KHalf output is rejected by the frontend on every backend (`Dimensions in out field and field domain are not equivalent`), and embedded asserts on two vertical dims in the scan range. Computing the surface pressure inside the program is not blocked; done in ce36f6dad as three calls in one program.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: The scan itself can't (an iteration sees one level, the extrapolation spans three), but it doesn't need a `CellField` either: the surface row is written into `pressure_ifc[nlev]` first and the scan reads `pressure_ifc(dims.KDim + 0.5)`, which only its first iteration (k = nlev-1) uses. Checked bit-identical to the two-program result on embedded, gtfn_cpu and dace_cpu before switching; in ce36f6dad.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: It was, by muphys and two tests, only because the hydrostatic program wanted the `CellField`. Deleted in ce36f6dad; they call `compute_hydrostatic_pressure`.

yiluchen1066 (approved):

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
```diff
@@ -0,0 +1,236 @@
...
+        pressure at full levels, pressure at the bounding upper interface of each level
+    """
+    pressure, pressure_ifc_on_model_levels, _ = _scan_pressure(
+        ddqz_z_full, virtual_temperature, pressure_ifc(dims.KDim + 0.5)
+    )
+    return pressure, pressure_ifc_on_model_levels
+
+
+@gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
+def compute_hydrostatic_pressure(  # noqa: PLR0917 [too-many-positional-arguments]
```
OngChia: I would rename this to `compute_surface_and_hydrostatic_pressure` to explicitly tell users that surface pressure is computed in the stencil, because I think modelers tend to think that this stencil only changes pressure above the ground using hydrostatic balance if the name is only compute_hydrostatic_pressure, even though they can look inside and see surface pressure is modified.

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/state.py
```diff
@@ -85,24 +83,14 @@ def __init__(
...
-            horizontal_sizes=full_horizontal,
-            vertical_sizes={
-                "vertical_start": gtx.int32(self._num_levels),
-                "vertical_end": gtx.int32(self._num_levels + 1),
-            },
-            offset_provider={},
-        )
-        self._diagnose_pressure = model_options.setup_program(
-            program=diagnose_pressure.diagnose_pressure,
+        self._compute_hydrostatic_pressure = model_options.setup_program(
```
OngChia: According to my comment above, we can change to `compute_surface_and_hydrostatic_pressure`.

OngChia (approved):
LGTM. I left two 2 comments asking to change the gt4py program `compute_hydrostatic_pressure` to `compute_surface_and_hydrostatic_pressure`. The rest is good.

model/common/src/icon4py/model/common/physics/thermodynamics/compute_pressure.py
havogt: Renamed in b7a039ee3.

One thing worth flagging, because it is not just a rename: the name you asked for was already held by the python host wrapper in this module, which is `diagnose_pressure_surface_to_top` on main (this PR had renamed it along with the move). There is no precedent in `model/common` for naming such a wrapper, so instead of inventing one to resolve the collision I removed it. **That does delete a function that exists on main**, so it is not purely a move. Both of its callers are updated here: the driver now calls the program directly, the way its neighbouring stencil calls already do, and the scratch buffer it used to allocate on every step moved to the driver's constructor, so it is allocated once per run. `diagnose_pressure_surface_to_top_ndarray` survives as `compute_surface_and_hydrostatic_pressure_ndarray` and drives the program directly.

If you would rather this PR not drop it, say so and I will restore it under a name of your choosing.

model/atmosphere/subgrid_scale_physics/muphys/src/icon4py/model/atmosphere/subgrid_scale_physics/muphys/state.py
havogt: Done in b7a039ee3, and the attribute here is `_compute_surface_and_hydrostatic_pressure` to match.

havogt:
I re-run the job that failed in the merge queue and it passed. I'll merge with bypassing the queue, no other things are in the queue.

PR #1318, opened 2026-06-15
Remove unnecessary mpi4py import in test_parallel_advection.py

msimberg:
This became unnecessary in #1223.

PR #1223, opened 2026-04-28
Update MPI test initialization

msimberg:
Fixes #1217.

Slightly refactors how MPI tests are skipped. If mpi4py isn't installed and --with-mpi is passed, there's now a hard error instead of a silent skip. MPI tests are filtered out at the pytest collection time, so I've removed the `if mpi not available: skip the full test module` from every mpi_tests file. The pytest hook for MPI tests will display the import error if mpi4py (or ghex) couldn't me imported and --with-mpi was passed. One of the MPI tests (test_compute_domain_bounds) was missing an mpi test marker, which means that it eagerly tried to import mpi4py which would always fail if mpi4py isn't available.

Finally, MPI is also initialized eagerly now in the pytest hook so that tests that just access e.g. communicator size or rank don't fail with e.g. "Attempting to use an MPI routine before initializing MPICH". MPI is initialized only if --with-mpi is set.

`test_create_single_node_runtime_without_mpi` was also incorrectly duplicated in an mpi_tests directory (there's another unit test with the same name). `test_create_single_node_runtime_with_mpi` also got a datatest marker now (since it needs DecompositionInfo which depends on grid data).

PR #1453, opened 2026-09-02
Remove redundant mpi4py import checks from MPI test files

msimberg:
These module-level `if mpi_decomposition.mpi4py is None: pytest.skip(...)` guards are redundant since #1223: MPI tests are filtered out at collection time, and `--with-mpi` raises a hard error if mpi4py/ghex couldn't be imported. #1318 removed the same pattern from `test_parallel_advection.py`; this removes it from the remaining files where it crept back in.


msimberg:
cscs-ci run default

edopao (approved):
LGTM

PR #1290, opened 2026-05-28
Enable more tests with embedded backend

havogt:

PR #1457, opened 2026-09-03
Add the tmx package, its configuration and static states

havogt:
This is the first of a series carving the tmx port #1359 into independently reviewable pieces: the package plus everything that needs no operators. Each later PR adds one subcomponent with its own datatest (diagnostics, scalar and momentum diffusion, energy update).

### What is here

- **The package**: `pyproject.toml`, README, workspace member, `tach` module (depends on `common` only), nox session, CI matrix entry.
- **`TmxConfig`** and its enums, in `tmx/config.py`, so that reading the configuration does not import every stencil package the granule does.
- **Static states**: `TmxMetricState` and `TmxInterpolationState`.
- **Metrics factory additions** the states need: `wgtfacq1_c/e`, the inverse layer thicknesses (cells on half levels, edges on full levels), geopotential and height above ground — each with a savepoint test in `test_metrics_factory.py`.
- **Quadratic extrapolation weights in gt4py**: `wgtfacq_c/e` and the new `wgtfacq1_c/e` replace the numpy implementations.
- **One savepoint** (`tmx_init`) and the test-data version bump.
- `unnamed_index` in the common config options (positional derived-type namelists), with its tests in common.
- ISO-8601 duration parsing moved to `common/utils/time_utils.py`; the driver had its own copy.
- `get_grid_geometry` keys its cache on the number of levels as well, since two experiments share the R02B04 grid file.

Co-authored-by: Jacopo Canton <jacopo.canton@gmail.com>
Co-authored-by: Ong Chia Rui <93439630+OngChia@users.noreply.github.com>
Co-authored-by: ychen <67939569+yiluchen1066@users.noreply.github.com>


model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
```diff
@@ -134,3 +134,53 @@ def compute_wgtfacq_e_dsl(
...
+    Args:
+        e2c: Edge to Cell offset
+        wgtfacq1_c: (ncells, 3) cell weights in Fortran coefficient order
+        c_lin_e: cell-to-edge linear interpolation weights
+    Returns:
+        (nedges, 3): row k multiplies full level k (``wgtfacq1_e(je, k+1, jb)``).
+    """
+    array_ns = data_alloc.array_namespace(e2c)
+    wgtfacq1_e = array_ns.sum(c_lin_e[:, :, array_ns.newaxis] * wgtfacq1_c[e2c], axis=1)
+    exchange.exchange(dims.EdgeDim, wgtfacq1_e, stream=decomposition.BLOCK)
```
havogt: Isn't the exchange a property of the factory entry?

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
```diff
@@ -134,3 +134,53 @@ def compute_wgtfacq_e_dsl(
...
+    """
+    array_ns = data_alloc.array_namespace(z_ifc)
+    z1, z2, z3 = _compute_z1_z2_z3(z_ifc, 0, 1, 2, 3)
+    w3 = z1 * z2 / (z2 - z3) / (z1 - z3)
+    w2 = (z1 - w3 * (z1 - z3)) / (z1 - z2)
+    w1 = 1.0 - (w2 + w3)
+    return array_ns.stack([w1, w2, w3], axis=1)
+
+
+def compute_wgtfacq1_e(
```
havogt: Could use gt4py?

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
```diff
@@ -134,3 +134,53 @@ def compute_wgtfacq_e_dsl(
     wgtfacq_e_dsl[:, nlev - 2] = z_aux_e[:, 2]
 
     return wgtfacq_e_dsl[:, -3:]
+
+
+def compute_wgtfacq1_c(
```
havogt: gt4py?

model/common/src/icon4py/model/common/math/operators.py
```diff
@@ -62,3 +62,19 @@ def _compute_field_a_plus_coeff_times_field_b_on_cell_k(
...
+    """Compute the elementwise reciprocal ``1 / input_field``."""
+    return wpfloat("1.0") / input_field
+
+
+@gtx.field_operator
+def _compute_reciprocal_on_edge_k(
+    input_field: fa.EdgeKField[ta.wpfloat],
+) -> fa.EdgeKField[ta.wpfloat]:
+    """Compute the elementwise reciprocal ``1 / input_field``."""
+    return wpfloat("1.0") / input_field
```
havogt: they are used in factories that's why we create these trivial operators?

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
```diff
@@ -0,0 +1,180 @@
...
+    "rbf_coeff_v1": 3.0e-10,
+    "rbf_coeff_v2": 3.0e-10,
+    "rbf_coeff_e": 5.0e-13,
+    "rbf_coeff_c1": 5.0e-9,
+    "rbf_coeff_c2": 5.0e-9,
+}
+
+
+@pytest.mark.datatest
+# through 'compute_ddqz_z_half', as in common's 'test_factory_ddqz_z_half'
```
havogt: what does this comment mean?

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+# The vertex-RBF solve requires the missing V2E neighbours of pentagon vertices to
+# be negative; the shared fixture keeps ICON's repeated-index padding, which makes
+# those twelve interpolation matrices singular. Mirror the driver's grid-file path.
+@pytest.fixture  # type: ignore[no-redef]
+def icon_grid(grid_savepoint, backend):
+    return grid_savepoint.construct_icon_grid(
+        backend=backend, keep_skip_values=True, with_repeated_index=False
+    )
```
havogt: let's think about this. maybe we should not invert the matrix outside its defined range?

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/surface_fluxes.py
```diff
@@ -0,0 +1,98 @@
...
+    surface_temperature: fa.CellField[ta.wpfloat]
+
+    def __post_init__(self) -> None:
+        if self.config.surface_type is not tmx_config.SurfaceType.FIXED_HEAT_FLUXES:
+            raise NotImplementedError(
+                "PrescribedFluxProvider implements "
+                f"{tmx_config.SurfaceType.FIXED_HEAT_FLUXES!r}, got {self.config.surface_type!r}."
+            )
+
+    def compute(self, *, out: tmx_states.TmxSurfaceFluxState) -> None:
```
havogt: Is this called only once or in each timestep?

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
```diff
@@ -0,0 +1,345 @@
...
+        surface_fluxes = {
+            field: convert(testcase[name])
+            for name, (field, convert) in surface_options.items()
+            if name in testcase
+        }
+        return common_conf_opt.construct_config_from_icon(
+            cls, atm_dict, **(surface_fluxes | overrides)
+        )
+
+    def _validate(self) -> None:
```
havogt: I think I would inline _validate into post_init. It's the typical place were validation lives.

model/common/src/icon4py/model/common/config/options.py
```diff
@@ -38,6 +38,13 @@ class IconOption:
     list_to_value: bool = False
     read_from_icon: bool = True
     converter: typing.Callable[[typing.Any], typing.Any] | None = None
+    # position of the option in an unnamed (positional) namelist record.
+    # Derived-type namelists (e.g. the AES physics ``aes_*_nml``) are echoed
+    # by ICON as an anonymous array of the member values in declaration
+    # order; for these, ``path`` leads to that array and ``unnamed_index``
+    # is the 0-based member position within one record (i.e. one domain),
+    # while ``name`` only serves as documentation.
+    unnamed_index: int | None = None
```
havogt, line 47: Let's ping @DropD again. Your comment on the other PR was that it's not used, but it actually is in the tmx/config.py

model/common/src/icon4py/model/common/math/operators.py
havogt: Yes, exactly that: a factory entry can only invoke a program, and a program needs an operator behind it. `copy_field_on_cell_khalf` in the same module exists for the same reason.

model/common/src/icon4py/model/common/config/options.py
havogt: For the ping: it is used. `TmxConfig` maps every option to its position in the Fortran `t_vdiff_config` derived type, which the echoed `aes_vdf_nml` reports as an unnamed array, so `unnamed_index` is what makes `from_fortran_dict` work. Two tests pin it: `test_config_from_fortran_dict_rejects_changed_member_count` and `..._rejects_missing_use_tmx`.

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
havogt: Moved. `compute_wgtfacq1_e` is gone entirely — the edge weights are now a `cell_2_edge_interpolation` of the cell ones — and the exchange is `do_exchange=True` on the factory entry. Same for `compute_wgtfacq_e_dsl`, which had the inline exchange too.

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
havogt: Done, and it needed no new stencil: the edge weights are exactly `c_lin_e` weighted over E2C of the cell weights, so both edge variants now reuse `cell_2_edge_interpolation`.

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
havogt: Done, both cell variants. One operator computes the three coefficients and each level picks its own; since all three read the same four interface heights, the shift differs per level (from full level k, interface j is at `KDim + (j - k) - 0.5`). Checked bit-for-bit against the old numpy on embedded and gtfn_cpu before wiring it in.

This needed one thing from the factory: a vertical sub-range in a provider's domain was ambiguous. For `rayleigh_w` and `coeff_dwdz` it means "the field spans the column, compute part of it"; here it means "the field is only these three levels". `ProgramFieldProvider` now takes `vertically_bounded_by_domain`, defaulting to the old behaviour.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
havogt: It was explaining why the marker is there — the test reaches `concat_where` through `compute_ddqz_z_half`. It now sits on the marker line instead of floating above it.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/surface_fluxes.py
havogt: Every step: the surface fluxes are recomputed each time the caller asks. That is why the docstring insists `pressure_ifc` be the live buffer. It is a gt4py stencil now — everything is on the surface, so a plain CellDim one — and the surface row is taken as a view of that buffer rather than copied per step.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
havogt: Agreed, inlined.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
```diff
@@ -0,0 +1,179 @@
...
+    "rbf_coeff_e": 5.0e-13,
+    "rbf_coeff_c1": 5.0e-9,
+    "rbf_coeff_c2": 5.0e-9,
+}
+
+
+@pytest.mark.datatest
+@pytest.mark.uses_concat_where  # reached through 'compute_ddqz_z_half'
+@pytest.mark.parametrize("experiment_description", [definitions.Experiments.EXCLAIM_APE_AES])
+def test_factory_static_states_match_savepoints(
```
OngChia: What is the reason of having a data test of static fields located inside tmx?
All static fields should already be tested in `common/test/`. If new static fields are added, we should put the data test of those new fields in `common/test/`?

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/unit_tests/test_tmx_config.py
```diff
@@ -0,0 +1,182 @@
...
+)
+from icon4py.model.common import model_backends
+from icon4py.model.common.config import config_io
+
+
+if TYPE_CHECKING:
+    from icon4py.model.common.grid import base as base_grid
+
+
+def test_default_config_matches_fortran_defaults() -> None:
```
OngChia: I am not very sure what is this test for?
I suggest to remove it.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
havogt: Agreed, and checking this turned up a gap. All nine interpolation fields are already covered field-by-field in `test_interpolation_factory.py`, but of the metric fields this PR newly registers — `inv_ddqz_z_half`, `inv_ddqz_z_full_e`, `inv_ddqz_z_half_e`, `inv_ddqz_z_half_v`, `geopot_agl_ifc`, `height_above_ground`, `wgtfacq1_c`, `wgtfacq1_e` — none had a test in `test_metrics_factory.py`. Those eight are now there, one per field, in the style of that file, and the rest is dropped. Six of them have their reference values only in the `tmx-init` savepoint, so those tests pin `EXCLAIM_APE_AES`. What remains in tmx is only that `from_sources` wires each state member to the right attribute, which needs no serialized data.

Two things fell out of the move: `height_above_ground` was being compared against the very function the factory registers, so it is now checked against `z_mc - z_ifc[:, -1:]` from the savepoints; and `get_grid_geometry` caches by grid file without the level count, so the two aquaplanet experiments — which share the R02B04 grid — were handing each other their `num_levels`. That is fixed in the same commit.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/unit_tests/test_tmx_config.py
havogt: Agreed, removed. It asserted the dataclass defaults against literals taken from the same dataclass, so it could not fail on a real change. `test_tmx_namelist_config.py::test_tmx_config_cross_checks_input_namelist_and_defaults` already checks every default the experiment does not set explicitly against what ICON's `vdiff_config_init` echoes, and the provenance is on the config fields themselves.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/integration_tests/test_static_fields.py
havogt: Heads-up on the anchor: this file is gone as of the latest commit (the static-field checks moved into `test_metrics_factory.py`, see the thread above), and with it the `icon_grid` fixture that passed `with_repeated_index=False`. The workaround existed only for the vertex-RBF comparison, which is now covered in common, where the grid comes from the grid file and the padding never appears.

The question itself stands and is unaffected: ICON's repeated-index padding makes the twelve pentagon vertex matrices singular in the RBF solve. Leaving this open for the separate discussion.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
```diff
@@ -0,0 +1,149 @@
...
+    wgtfac_c: fa.CellKHalfField[ta.wpfloat]
+    """Weighting factor for interpolation from full to half levels at cell centers (half levels)."""
+    wgtfac_e: fa.EdgeKHalfField[ta.wpfloat]
+    """Weighting factor for interpolation from full to half levels at edge midpoints (half levels)."""
+    wgtfacq_c: fa.CellKField[ta.wpfloat]
+    """Extrapolation coefficients to the bottom surface half level at cell centers.
+
+    Three K rows aligned to the levels they multiply: the row at K index j is the
+    weight of full level j, so the field is defined on KDim in [nlev - 3, nlev).
+    This is what the metrics factory emits and what the dycore stencils consume."""
```
OngChia:
```suggestion
    """Quadratic extrapolation coefficients to the surface level at cell centers."""
```

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
```diff
@@ -0,0 +1,149 @@
...
+    wgtfacq1_c: fa.CellKField[ta.wpfloat]
+    """Extrapolation coefficients to the top half level at cell centers.
+
+    Three K rows aligned to the levels they multiply, i.e. KDim in [0, 3)."""
```
OngChia:
```suggestion
    """Quadratic extrapolation coefficients to the model top level at cell centers."""
```

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
```diff
@@ -0,0 +1,149 @@
...
+    wgtfacq_e: fa.EdgeKField[ta.wpfloat]
+    """Extrapolation coefficients to the bottom surface half level at edges.
+
+    Aligned to the levels they multiply, see :attr:`wgtfacq_c`."""
```
OngChia:
```suggestion
    """Quadratic extrapolation coefficients to the surface level at edges."""
```

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
```diff
@@ -0,0 +1,149 @@
...
+    wgtfacq1_e: fa.EdgeKField[ta.wpfloat]
+    """Extrapolation coefficients to the top half level at edges.
+
+    Aligned to the levels they multiply, see :attr:`wgtfacq1_c`."""
```
OngChia:
```suggestion
    """Quadratic extrapolation coefficients to the model top level at edges."""
```

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
```diff
@@ -0,0 +1,149 @@
...
+    z_mc: fa.CellKField[ta.wpfloat]
+    """Geometric height at cell centers on full levels [m]."""
+    z_ifc: fa.CellKHalfField[ta.wpfloat]
+    """Geometric height at cell centers on half levels [m]."""
+    edge_cell_length: gtx.Field[gtx.Dims[dims.EdgeDim, dims.E2CDim], ta.wpfloat]
+    """Distance between the edge midpoint and the circumcenters of the two
+    adjacent cells [m] (``t_grid_edges%edge_cell_length`` in mo_model_domain.f90).
+
+    A grid-geometry field carried here because it is not part of the common
+    ``grid_states.EdgeParams`` (used by the horizontal w diffusion)."""
```
OngChia: As the comment says, I think we should add `edge_cell_length` (it is called `edge_cell_distance` in grids) in `EdgeParams` and remove from here because it is purely mesh or grid property. It should not be a metric member because metric state depends on the vertical levels and user config parameters set when running a simulation.

At the same time, I am questioning to myself what is `EdgeParams`. Why do we need to pass in the entire `EdgeParams` to all granules or components even when some components only use part of it? Can't we just pass in geometry source? Maybe we will have to wait until the state and component task is completed before I can answer that.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/unit_tests/test_tmx_config.py
```diff
@@ -0,0 +1,151 @@
...
+def test_config_from_fortran_dict_rejects_missing_use_tmx() -> None:
+    record = _echoed_vdf_record()
+    record[22] = False
+    with pytest.raises(ValueError, match="use_tmx"):
+        TmxConfig.from_fortran_dict(
+            atm_dict={"aes_vdf_nml": {"aes_vdf_config": record}}, input_dict={}
+        )
+
+
+def test_config_round_trips_through_config_io() -> None:
```
OngChia, line 126: @DropD 
Do we need this roundtrip test in every component for io config?

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/fixtures.py
```diff
@@ -0,0 +1,108 @@
...
+    input_dict = load_fortran_dict(
+        experiment_description=experiment_description,
+        process_props=process_props,
+        fname=fortran_config.INPUT_DICT_FNAME,
+    )
+    return TmxConfig.from_fortran_dict(atm_dict=atm_dict, input_dict=input_dict)
+
+
+# ISO 8601 duration, fixed-length components only (duplicated from
+# driver.config, which the tmx tests may not depend on)
```
OngChia: Maybe move this `_ISO8601_DURATION` to some helper function in testing?

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -955,3 +955,39 @@ def compute_exner_w_implicit_weight_parameter(
...
+    Geometric height of the full levels above the surface.
+
+    Port of ``compute_geopotential_height_above_ground`` (mo_vdf_atmo.f90):
+
+        ghf(jk) = z_mc(jk) - z_ifc(nlevp1)
+
+    Despite the Fortran name the result is a geometric height in meters;
+    gravity is only applied by the users of the field (e.g.
+    ``compute_dry_static_energy``). Computed with numpy because GT4Py offsets
+    are relative and cannot address the fixed absolute surface row of ``z_ifc``.
```
OngChia:
```suggestion
        height_above_ground = z_mc - z_ifc[nlevp1]

    Computed with numpy because GT4Py offsets
    are relative and cannot address the fixed absolute surface row of ``z_ifc``.
```

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -955,3 +955,39 @@ def compute_exner_w_implicit_weight_parameter(
...
+
+    Args:
+        z_mc: geometric height of the full levels [m]
+        z_ifc: geometric height of the half levels [m], surface at the last row
+
+    Returns:
+        height of the full levels above the surface [m]
```
OngChia:
```suggestion
        height_above_ground: height of the full levels above the surface [m]
```

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -522,14 +521,31 @@ def __init__(
         }
         self._do_exchange = do_exchange
 
+    def _field_extent(self, dim: gtx.Dimension, grid: GridProvider) -> tuple[int, int]:
+        """The extent to allocate for ``dim``.
```
OngChia: I did not know this.
So, just to confirm, we will create `CellKField` fields such as the quadratic extrapolation coefficients `wgtfacq_c` or `wgtfacq1_c` with only three levels in the vertical coordinate.

model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -1962,6 +1962,105 @@ def topo_smt_c(self):
...
+
+
+class TmxEntrySavepoint(IconSavepoint):
+    """Savepoint at entry of vdf Compute in mo_vdf.f90 (inputs of the TMX scheme)."""
+
+    def ta(self):
+        return self._get_field("ta", dims.CellDim, dims.KDim)
+
+    def ta_phy(self):
+        # Sanity twin of `ta`: prm_field%ta, must be identical to the tmx input temp_c.
```
OngChia: I thought this is tendency of temperature due to physics processes. It seems that this is not used anywhere when running the datatest. Could you check again?

OngChia (commented):
I am still thinking about the bunch of tests of tmx config. No clue yet. Could @DropD have a look at those config things?

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
```diff
@@ -60,77 +60,121 @@ def _compute_z1_z2_z3(
...
+        z_ifc(dims.KDim + 1.5),
+        z_ifc(dims.KDim + 2.5),
+    )
+    _, w2, _ = _quadratic_extrapolation_weights(
+        z_ifc(dims.KDim - 1.5),
+        z_ifc(dims.KDim - 0.5),
+        z_ifc(dims.KDim + 0.5),
+        z_ifc(dims.KDim + 1.5),
+    )
+    _, _, w3 = _quadratic_extrapolation_weights(
```
OngChia: This workaround of cyclic permutation of vertical levels to compute simple quadratic extrapolation weights seems not very ideal and hard to swallow. I guess this is because of gt4py limitations.

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
```diff
@@ -60,77 +60,121 @@ def _compute_z1_z2_z3(
     return z1, z2, z3
```
OngChia: This helper function can be removed since the quadratic extrapolation weights are computed in gt4py stencils in this PR.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
havogt: Applied as suggested.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
havogt: Applied as suggested.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
havogt: Applied as suggested.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
havogt: Applied as suggested.

model/common/src/icon4py/model/common/metrics/metric_fields.py
havogt: Applied as suggested.

model/common/src/icon4py/model/common/metrics/metric_fields.py
havogt: Applied as suggested.

model/common/src/icon4py/model/common/metrics/compute_weight_factors.py
havogt: Right — it was the numpy path the gt4py conversion replaced, and it had no caller left anywhere in the repo. Removed.

model/testing/src/icon4py/model/testing/serialbox.py
havogt: Checked, and you are right about the name but the data is something else again.

In ICON `ta_phy` belongs to the *tendency* list — `mo_aes_phy_memory.f90` l.544 declares `tend%ta_phy(:,:,:)` with unit `[K/s]`, registered as `prefix//'ta_phy'` at l.4443 — so reading it as a physics temperature tendency is the natural reading. But that is not what the archive holds: in the `tmx-entry` savepoint of `exclaim_ape_aesPhys` v08, `ta` and `ta_phy` are bit-identical (`np.array_equal` true, max 305.6 K), i.e. a temperature, not a tendency. Whatever the serialization wrote under that name, it is not `tend%ta_phy`; I could not check the `!$ser` statement itself, the branch that produced these savepoints is not one I have here.

And your second point settles it: nothing reads the accessor, here or in the port. Removed rather than left as a mislabelled field waiting to be picked up by name.

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/fixtures.py
havogt: Moved, though one level further than you suggested: it went to `common/utils/time_utils.py` rather than into the testing package, because it was already a duplicate — `driver/config.py` had the same regex and the same `relativetime_from_iso8601`, and the copy here carried a comment saying so. Both now call the shared one, and its two unit tests moved from the driver test suite to `common/tests/common/utils/unit_tests/test_time_utils.py`.

The tmx fixture is now `relativetime_from_iso8601(dt_vdf).total_seconds()`.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/tmx_states.py
havogt: Agreed on the classification, and we tried it — but it does not land as a straight move, and the reason is the second half of your comment.

`EdgeParams` is also the Fortran-facing bundle. `bindings/grid_wrapper.py:208` builds one out of arrays passed in from Fortran, and there is no `edge_cell_length` among them (the `TODO(havogt)` right above that call anticipates exactly this case). Two bindings datatests then compare that object attribute-by-attribute against `grid_savepoint.construct_edge_geometry()` — `test_diffusion_wrapper.py:218` and `test_dycore_wrapper.py:660`, via a helper that loops `vars()` — so a member only the savepoint side can fill turns into `None` vs a field and blows up inside gt4py's `__ne__`.

So adding to `EdgeParams` obliges the Fortran side, which is a cross-repo change and not one to slip into this PR. The alternatives are to add the member but leave it unfilled everywhere (useless for the consumer this is for), or to leave the field on `TmxMetricState` for now with the ICON name spelled out. Holding it here until that is decided — the patch is written and green apart from the bindings suite.

On the wider question: this is the same edge as "why does a component get the whole `EdgeParams`". A component-shaped geometry state, or a geometry source passed directly, would make the answer here obvious. I would rather it come out of the state/component work than be pre-empted by this PR.

havogt: Actually I opened now https://github.com/C2SM/icon4py/pull/1462 which adds this properly to EdgeParams and cleans up code around it.

model/common/src/icon4py/model/common/states/factory.py
havogt: Was extracted into #1463

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
```diff
@@ -0,0 +1,258 @@
...
+
+from icon4py.model.common.config import config_io, options as common_conf_opt
+
+
+@config_io.register_enum
+class TurbulenceSolverType(int, enum.Enum):
+    """
+    Type of the vertical diffusion solver.
+
+    Note: Called ``solver_type`` in ``mo_turb_vdiff_config.f90``.
```
yiluchen1066:
```suggestion
```
No need to keep this, also for the rest of the codes.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
```diff
@@ -0,0 +1,258 @@
...
+import dataclasses
+import enum
+import typing
+from typing import Any
+
+from icon4py.model.common.config import config_io, options as common_conf_opt
+
+
+@config_io.register_enum
+class TurbulenceSolverType(int, enum.Enum):
```
yiluchen1066: shall we rename this to ```TmxSolverType```?

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
```diff
@@ -0,0 +1,258 @@
...
+        """
+        Construct the configuration from the echoed ICON namelist.
+
+        ``aes_vdf_nml`` is a derived-type namelist (``t_vdiff_config``), which
+        ICON echoes as an anonymous positional array of the member values in
+        declaration order, so the options are located by ``unnamed_index``
+        (pinned to mo_turb_vdiff_config.f90) instead of by name. Only the
+        first domain is read. The guards below make a change of the Fortran
+        type fail loudly instead of silently mis-assigning values.
+        """
```
yiluchen1066, line 224:
```suggestion
        """
        Build the configuration from the echoed ICON namelist. 
        
        ICON writes ```aes_vdf_config``` values in Fortran member order, without names. 
        We read the first domain using each options's ```unnamed_index```. The checks below 
        help detect changes to the expected Fortran layout
        """
```
let's simplify the doc string a bit.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
```diff
@@ -0,0 +1,258 @@
...
+        # number of members of the Fortran t_vdiff_config derived type
+        # (mo_turb_vdiff_config.f90); the echoed aes_vdf_nml namelist holds this
+        # many values per domain, in declaration order. Must be kept in sync with
+        # the 'unnamed_index' positions of the options above.
+        num_members = 42
+        # position of 'use_tmx' in t_vdiff_config, used as an order canary
+        use_tmx_index = 22
```
yiluchen1066:
```suggestion
        # Layout of t_vdiff_config in mo_turb_vdiff_config.f90
        # Keep these values and the options' unnamed_index positions in sync
        num_members = 42
        use_tmx_index = 22
```

model/common/src/icon4py/model/common/config/options.py
yiluchen1066: I was testing GPT-6 Astra, and it caught this here: 

```suggestion
values = dict(iter_pairs_from_icon(config_cls, icon_config))
values.update(overrides)
return config_cls(**values)
```
Both dictionaries can contain ```km_const``` which caused the error

model/common/tests/common/metrics/unit_tests/test_compute_weight_factors.py
```diff
@@ -70,40 +70,27 @@ def test_compute_wgtfac_c(
...
-        nlev=icon_grid.num_levels,
-        exchange=decomposition.SingleNodeExchange(),
-    )
-
-    assert test_utils.dallclose(data_alloc.as_numpy(wgtfacq_e_dsl), wgtfacq_e_ref.asnumpy())
-
-
-@pytest.mark.datatest
-def test_compute_wgtfacq_c_dsl(
+def test_compute_wgtfacq_c(
```
yiluchen1066, line 74:
```suggestion
@pytest.mark.datatest
@pytest.mark.uses_concat_where
def test_compute_wgtfacq_c(
```

model/testing/src/icon4py/model/testing/serialbox.py
```diff
@@ -1975,6 +1975,37 @@ def topo_smt_c(self):
...
+        # surface-first, i.e. reversed w.r.t. increasing k, hence flipped in its accessor),
+        # `wgtfacq1_c(jc,k,jb)` with k=1..3 multiplies the full level k counted from the model
+        # top (mo_vertical_grid.f90 ll. 955-968), which already matches icon4py's top-down KDim
+        # orientation. No flip needed.
+        return self._get_field("wgtfacq1_c", dims.CellDim, dims.KDim)
+
+    def wgtfacq1_e(self):
+        # No flip, see `wgtfacq1_c`. Usage in mo_vdf_atmo.f90 (interpolate_normal_velocity_
+        # edge_interface, ll. 1247-1250): vn_ie(je,1,jb) = sum_k wgtfacq1_e(je,k,jb)*vn(je,k,jb).
+        return self._get_field("wgtfacq1_e", dims.EdgeDim, dims.KDim)
```
yiluchen1066, line 1996:
```suggestion
    def wgtfacq1_c(self):
        return self._get_field("wgtfacq1_c", dims.CellDim, dims.KDim)

    def wgtfacq1_e(self):
        return self._get_field("wgtfacq1_e", dims.EdgeDim, dims.KDim)
```
let's remove the docs here

model/atmosphere/subgrid_scale_physics/tmx/tests/tmx/fixtures.py
```diff
@@ -0,0 +1,49 @@
...
+from icon4py.model.common.utils import fortran_config
+from icon4py.model.testing import datatest_utils as dt_utils, definitions
+from icon4py.model.testing.fixtures.datatest import (
+    download_ser_data,
+    experiment_description,
+    process_props,
+)
+
+
+def load_fortran_dict(
```
yiluchen1066: shall we move this to ```model/testing/datatest_utils``` since it uses the helper function ```get_path_for_experiment```  and potentially other components would also need this helper

yiluchen1066 (commented):
Thanks for the PR! I did not find something need to be changed urgently. Mostly minor things. We can merge this one pretty soon!

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
havogt: Actually, I propose to drop the prefix. In the code we should probably do `import icon4py.<stuff>.tmx.config as tmx_config`, then the type reads nicely as `tmx_config.SolverType`. Ok?

model/common/tests/common/metrics/unit_tests/test_compute_weight_factors.py
havogt: This test is actually ok with embedded, so we don't need to put the marker. The marker should probably be renamed to `uses_concat_where_with_unbounded_domain` or something. I have #1290 for dropping more of these markers, but it's blocked by another PR.

model/atmosphere/subgrid_scale_physics/tmx/src/icon4py/model/atmosphere/subgrid_scale_physics/tmx/config.py
yiluchen1066: agree! this makes more sense

yiluchen1066 (approved):
LGTM!

PR #1458, opened 2026-09-04
Prescribed surface fluxes for tmx

havogt:
Split out of #1457, where the review pointed out that the flux providers have
no caller in that PR. They are the surface-flux input to the scalar and
momentum diffusion components of tmx and arrive ahead of them.

Contents: the two providers (zero and prescribed), `TmxSurfaceFluxState`, the
`tmx_surface_fluxes` savepoint, and the datatest that checks the prescribed
fluxes against the serialized reference.

The computation is a gt4py stencil rather than numpy. Everything is on the
surface, one horizontal layer, so it is a plain `CellDim` one; the surface
pressure is the bottom interface row of `pressure_ifc`, taken as a view of
the live buffer rather than copied on every step. `sat_pres_water` and
`specific_humidity` gain field operators beside their numpy versions, and
`PhysicsConstants` the four values they reference, because gtfn needs
constants used inside an operator to come from the wpfloat enum.

Tests: 19 pass on gtfn_cpu, including the prescribed-flux datatest.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01QVSWPQauF9gNjADtuSTjxF


havogt:
This is extracted, because even in the full port it didn't have callers, so maybe it comes together with @yiluchen1066 's driver?

yiluchen1066:
> This is extracted, because even in the full port it didn't have callers, so maybe it comes together with @yiluchen1066 's driver?

you are right! the code wasn’t used at all in the tmx port pr, so how it’s used would only show up in the integration into the driver. And we checked together with @OngChia  and @jcanton in the fortran code, ```tmx``` has two parts: ```sfc``` (which means surface flux) and ```atmo``` (which means the tmx computation which is also the tmx ported in #1359). And the ```PrescribedFluxProvider``` is an simplified surface flux component. Later on we will port the whole proper surface flux as a component, and then ```PrescribedFluxProvider``` should be part of it, and also the ```ZeroFluxProvider```. 

so agreed that this should go into the physics driver for tmx pr #1360, where we can already have the skeleton for the proper surface flux to fit in later on.

PR #1460, opened 2026-09-04
Merge VelocityAdvection into two programs with per-output domains

havogt:
### Programs: 5 calls → 2

| Step | Before | After |
|---|---|---|
| Predictor | `compute_diagnostics_from_normal_wind`, `compute_advection_in_predictor_vertical_momentum`, `compute_advection_in_horizontal_momentum` | `compute_velocity_advection_in_predictor_step` (9 outputs) |
| Corrector | `compute_advection_in_corrector_vertical_momentum`, `compute_advection_in_horizontal_momentum` | `compute_velocity_advection_in_corrector_step` (3 outputs) |

- `horizontal_advection_of_w_at_edges_on_half_levels` (ICON `z_v_grad_w`) and `contravariant_corrected_w_at_cells_on_model_levels` (`z_w_con_c_full`) are no longer program fields. GT4Py derives their extents from their consumers' neighbour reads.
- Both host `max_vertical_cfl` reductions run after the merged call. This is safe because `apply_extra_diffusion_on_vn` is fixed to `True`, and nothing between the calls reads the reduction.

### `VelocityAdvection` is dissolved into `SolveNonhydro`

`SolveNonhydro` sets up and calls the two programs itself, and `velocity_advection.py` is gone. The `dtime` scale factors and the max-CFL reduction are module-level functions in `solve_nonhydro.py`, shared with the datatests.

### Operators

- **Modules.** Step-agnostic operators live in `stencils/velocity_advection_terms.py`, and step-specific wiring in `velocity_advection_predictor.py` and `velocity_advection_corrector.py`. `compute_advection_in_{vertical,horizontal}_momentum_equation.py` are deleted. `compute_diagnostics_from_normal_wind.py` keeps the two operators that other programs import.
- **One vertical-momentum operator** is shared by predictor and corrector. The predictor's `skip_compute_predictor_vertical_advection` is applied at the call site: the skipped branch returns a placeholder that is never read, and the predictor keeps the previous tendency.
- **The vertical wind tendency is computed as three named contributions** (`_compute_*`, not `_add_*`). The additions stay sequenced so that each contribution is rounded to `vpfloat` as in ICON; a TODO asks whether those wp/vp round trips were deliberate on the Fortran side.
- **Computing the CFL is separated from clipping `w`**: `_compute_cfl` and `_clip_contravariant_corrected_w`.
- **One tangential-wind operator** (`_compute_tangential_wind`), with the precision conversion at the call sites. The `compute_tangential_wind_vp` program is removed.
- `add_extra_diffusion_for_normal_wind_tendency_approaching_cfl.py` (the levelmask variant) is kept with a note: it is unused, and kept for an ICON-faithful version.


model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+    cfl_w_limit: ta.vpfloat,
+    dtime: ta.wpfloat,
+) -> fa.CellKHalfField[ta.vpfloat]:
+    vertical_wind_advective_tendency = (
+        _add_vertical_advection_of_w_to_advective_vertical_wind_tendency(
+            contravariant_corrected_w_at_cells_on_half_levels, w, coeff1_dwdz, coeff2_dwdz
+        )
+    )
+
+    vertical_wind_advective_tendency = _add_interpolated_horizontal_advection_of_w(
```
havogt: I think it reads nicer if we make this function compute the tendency, i.e. make it `compute_interpolated_horizontal_advection_of_w`, then do the sum here.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+        e_bln_c_s,
+        horizontal_advection_of_w_at_edges_on_half_levels,
+        vertical_wind_advective_tendency,
+    )
+
+    vertical_wind_advective_tendency = _add_extra_diffusion_for_w_con_approaching_cfl(
```
havogt: maybe we move the clipping mask to here and call this one `compute_extra_diffusion_for_w`?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+    geofac_n2s: gtx.Field[gtx.Dims[dims.CellDim, dims.C2E2CODim], ta.wpfloat],
+    owner_mask: fa.CellField[bool],
+    scalfac_exdiff: ta.wpfloat,
+    cfl_w_limit: ta.vpfloat,
+    dtime: ta.wpfloat,
+    nlev: gtx.int32,
+    end_index_of_damping_layer: gtx.int32,
+) -> tuple[fa.CellKHalfField[ta.vpfloat], fa.CellKField[ta.vpfloat], fa.CellKHalfField[ta.vpfloat]]:
+    #: intermediate variable horizontal_advection_of_w_at_edges_on_half_levels is originally declared as z_v_grad_w in ICON
+    horizontal_advection_of_w_at_edges_on_half_levels = _compute_horizontal_advection_of_w(
```
havogt: what about moving this call into `_compute_advective_vertical_wind_tendency`?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+        inv_primal_edge_length=inv_primal_edge_length,
+        tangent_orientation=tangent_orientation,
+    )
+
+    (
+        contravariant_corrected_w_at_cells_on_half_levels,
+        cfl_clipping,
+        vertical_cfl,
+    ) = _compute_contravariant_corrected_w_and_cfl(
+        w=w,
```
havogt: I think we should split this operator into `compute_contravariant_corrected_w` and `compute_cfl`

havogt: Not, straight-forward, but this reads just weird.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+    skip_compute_predictor_vertical_advection: bool,
+    nlev: gtx.int32,
+) -> tuple[
+    fa.EdgeKField[ta.vpfloat],
+    fa.EdgeKHalfField[ta.vpfloat],
+    fa.EdgeKHalfField[ta.vpfloat],
+    fa.EdgeKField[ta.vpfloat],
+    fa.EdgeKField[ta.vpfloat],
+]:
+    tangential_wind = _compute_tangential_wind_vp(vn, rbf_vec_coeff_e)
```
havogt:
```suggestion
    tangential_wind = astype(_compute_tangential_wind(vn, rbf_vec_coeff_e), vpfloat)
```
and delete the _vp variant, the _wp is dropped.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
```diff
@@ -0,0 +1,1300 @@
...
+        dims.KHalfDim < nlev,
+        _interpolate_to_half_levels(wgtfac_e, vn),
+        _extrapolate_at_top(wgtfacq_e, vn),
+    )
+
+    tangential_wind_on_half_levels = (
+        _interpolate_to_half_levels(wgtfac_e, tangential_wind)
+        if not skip_compute_predictor_vertical_advection
+        else tangential_wind_on_half_levels
+    )
```
havogt: maybe we can find a better pattern for this that allows us to not pass in the `tangential_wind_on_half_levels`, just to do nothing `if skip_compute_predictor_vertical_advection`

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/compute_velocity_advection.py
havogt: maybe inlining here would resolve it: first compute cor_w, then compute clipping, then clip

havogt:
cscs-ci run default;MODEL_SUBPACKAGES=dycore:common:tracer_advection;SESSIONS=model

havogt:
Overview of the current state https://claude.ai/code/artifact/1441c859-e2ed-49b0-a517-6613716af1e3

model/atmosphere/dycore/tests/dycore/stencil_tests/test_compute_horizontal_velocity_quantities_and_fluxes.py
```diff
@@ -22,13 +22,34 @@
...
+def interpolate_vn_to_half_levels_and_compute_kinetic_energy_numpy(
+    wgtfac_e: np.ndarray, vn: np.ndarray, vt: np.ndarray
+) -> tuple[np.ndarray, np.ndarray]:
+    return (
+        interpolate_vn_to_half_levels_numpy(wgtfac_e, vn),
+        compute_horizontal_kinetic_energy_at_edges_numpy(vn, vt),
+    )
+
+
+def copy_to_half_levels_and_compute_kinetic_energy_at_top_numpy(
```
nfarabullini, line 43: these are incredibly long names, can you shorten?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
```diff
@@ -0,0 +1,609 @@
...
+    ddqz_z_half: fa.CellKHalfField[ta.vpfloat],
+    area: fa.CellField[ta.wpfloat],
+    geofac_n2s: gtx.Field[gtx.Dims[dims.CellDim, dims.C2E2CODim], ta.wpfloat],
+    owner_mask: fa.CellField[bool],
+    scalfac_exdiff: ta.wpfloat,
+    cfl_w_limit: ta.vpfloat,
+    dtime: ta.wpfloat,
+) -> fa.CellKHalfField[ta.vpfloat]:
+    # TODO(havogt): the wp-vp roundtrips are here to be faithful to ICON's mixed precision,
+    # but was that a deliberate decision on the Fortran side?
```
nfarabullini: was it?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
```diff
@@ -0,0 +1,609 @@
...
+            owner_mask=owner_mask,
+            scalfac_exdiff=scalfac_exdiff,
+            cfl_w_limit=cfl_w_limit,
+            dtime=dtime,
+        )
+        if not skip_vertical_wind_advective_tendency
+        else broadcast(
+            vpfloat("0.0"),
+            (dims.CellDim, dims.KHalfDim),
+        )  # "0.0" is a dummy value and should never be used
```
nfarabullini, line 384: The comment doesn't match the action

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_corrector.py
```diff
@@ -0,0 +1,307 @@
...
+    compute_advective_vertical_wind_tendency_and_apply_diffusion_numpy,
+    compute_horizontal_advection_of_w_numpy,
+    compute_maximum_cfl_and_clip_contravariant_vertical_velocity_numpy,
+    interpolate_contravariant_vertical_velocity_to_full_levels_numpy,
+)
+
+
+@pytest.mark.embedded_remap_error
+@pytest.mark.uses_concat_where
+@pytest.mark.continuous_benchmarking
```
nfarabullini, line 34: are all of these still relevant or have any been solved by now?

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_corrector.py
```diff
@@ -0,0 +1,307 @@
...
+            cfl_w_limit=cfl_w_limit,
+            dtime=dtime,
+            apply_extra_diffusion_on_vn=request.param["apply_extra_diffusion_on_vn"],
+            end_index_of_damping_layer=end_index_of_damping_layer,
+            start_cell_lateral_boundary_level_4=grid.start_index(
+                cell_domain(h_grid.Zone.LATERAL_BOUNDARY_LEVEL_4)
+            ),
+            end_cell_halo=grid.end_index(cell_domain(h_grid.Zone.HALO)),
+            start_edge_nudging_level_2=grid.start_index(edge_domain(h_grid.Zone.NUDGING_LEVEL_2)),
+            end_edge_local=grid.end_index(edge_domain(h_grid.Zone.LOCAL)),
```
nfarabullini, line 304: can we make this cleaner and define these above?

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_predictor.py
```diff
@@ -0,0 +1,430 @@
...
+        wgtfac_e = data_alloc.random_field(dims.EdgeDim, dims.KHalfDim)
+        wgtfacq_e = data_alloc.random_field(dims.EdgeDim, dims.KDim)
+        ddxn_z_full = data_alloc.random_field(dims.EdgeDim, dims.KDim)
+        ddxt_z_full = data_alloc.random_field(dims.EdgeDim, dims.KDim)
+        coeff1_dwdz = data_alloc.random_field(dims.CellDim, dims.KDim)
+        coeff2_dwdz = data_alloc.random_field(dims.CellDim, dims.KDim)
+        c_intp = data_alloc.random_field(dims.VertexDim, dims.V2CDim)
+        inv_dual_edge_length = data_alloc.random_field(dims.EdgeDim, low=1.0e-5)
+        inv_primal_edge_length = data_alloc.random_field(dims.EdgeDim, low=1.0e-5)
+        tangent_orientation = data_alloc.random_field(dims.EdgeDim, low=1.0e-5)
```
nfarabullini, line 346: why do these have a low value here if in other parts of the granules they do not?

nfarabullini (commented):
just some comments

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: We concluded it wasn't and @OngChia and me came to the conclusion that we want to open a PR after this big one which removes the wp-vp conversions in a way that matches what we think ICON wanted to do.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_corrector.py
havogt: I have https://github.com/C2SM/icon4py/pull/1290 open to address embedded related markers. That PR is not in a good shape, because I found a problem in the solver related to KHalfDim/KDim handling in scans.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_predictor.py
havogt: I guess 0 lengths would result in a blow up somewhere which would give huge errors (or nans).

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: I think it does, but maybe the comment is not clear enough. See the corresponding comment in the predictor field_operator.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
nfarabullini: ok then can you remove this comment?

havogt:
```suggestion
    # TODO(havogt): the wp-vp roundtrips are here to be faithful to ICON's mixed precision.
    # We assume it was not a deliberate decision in ICON and plan to drop the extra conversions
    # in a separate PR.
```

havogt:
cscs-ci run default;MODEL_SUBPACKAGES=dycore:common:tracer_advection;SESSIONS=model

havogt:
cscs-ci run default;MODEL_SUBPACKAGES=dycore:common:tracer_advection;SESSIONS=model

model/atmosphere/dycore/tests/dycore/stencil_tests/test_compute_horizontal_velocity_quantities_and_fluxes.py
havogt: Inlined both in 8789c9286. Each had a single caller, and the level selection they fed was redundant: the interpolation helpers already copy model level 0 to the top half level, and the kinetic energy uses the same formula on every level. The reference is now three assignments.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: Reworded and moved onto the `else` branch in 72faed521. The zero field is a placeholder that is never read: when the tendency is skipped, the caller keeps its previous value.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_corrector.py
havogt: Done in a1a42dec8, in the predictor test as well.

havogt:
cscs-ci run default;MODEL_SUBPACKAGES=dycore:common:driver:tracer_advection;MODEL_MPI_SUBPACKAGES=dycore:driver;SESSIONS=model:model_mpi:tools

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
```diff
@@ -0,0 +1,608 @@
...
+    vn_on_half_levels: fa.EdgeKHalfField[ta.vpfloat],
+    c_intp: gtx.Field[gtx.Dims[dims.VertexDim, dims.V2CDim], ta.wpfloat],
+    inv_dual_edge_length: fa.EdgeField[ta.wpfloat],
+    inv_primal_edge_length: fa.EdgeField[ta.wpfloat],
+    tangent_orientation: fa.EdgeField[ta.wpfloat],
+) -> fa.EdgeKHalfField[ta.vpfloat]:
+    w_at_vertices = _mo_icon_interpolation_scalar_cells2verts_scalar_ri_dsl(w, c_intp)
+
+    horizontal_advection_of_w_at_edges_on_half_levels = (
+        _compute_horizontal_advection_term_for_vertical_velocity(
```
OngChia: I prefer to remove this field operator and directly inline it here. 
Basically all it does is `vn dw/dn + vt dw/dt`, which is exactly `horizontal_advection_of_w`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
```diff
@@ -0,0 +1,608 @@
...
+    neighbor_sum,
+    where,
+)
+from gt4py.next.experimental import concat_where
+
+from icon4py.model.atmosphere.dycore.stencils.compute_horizontal_advection_term_for_vertical_velocity import (
+    _compute_horizontal_advection_term_for_vertical_velocity,
+)
+from icon4py.model.atmosphere.dycore.stencils.mo_icon_interpolation_scalar_cells2verts_scalar_ri_dsl import (
+    _mo_icon_interpolation_scalar_cells2verts_scalar_ri_dsl,
```
OngChia: Can we rename it to something like `interpolate_cell_field_to_vertex` in this PR?

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
```diff
@@ -0,0 +1,608 @@
...
+)
+from icon4py.model.atmosphere.dycore.stencils.mo_math_divrot_rot_vertex_ri_dsl import (
+    _mo_math_divrot_rot_vertex_ri_dsl,
```
OngChia: and this can be renamed as `compute_curl`.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
```diff
@@ -0,0 +1,456 @@
...
+import gt4py.next as gtx
+from gt4py.next import astype, broadcast
+from gt4py.next.experimental import concat_where
+
+from icon4py.model.atmosphere.dycore.stencils.compute_contravariant_correction import (
+    _compute_contravariant_correction,
+)
+from icon4py.model.atmosphere.dycore.stencils.compute_diagnostics_from_normal_wind import (
+    _compute_horizontal_kinetic_energy,
+    _interpolate_to_half_levels,
```
OngChia, line 17: this looks like it should be in common/interpolation. But I think it can also be moved in tmx subcomponent port.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
```diff
@@ -0,0 +1,456 @@
...
+    _interpolate_cell_field_to_half_levels_vp,
+)
+from icon4py.model.common.interpolation.stencils.interpolate_to_cell_center_vp import (
+    _interpolate_to_cell_center_vp,
+)
+from icon4py.model.common.type_alias import vpfloat
+
+
+@gtx.field_operator
+def _compute_diagnostics_from_normal_wind(
```
OngChia, line 38: Just leaving a note here, as we discussed personally, this field operator can be moved somewhere else in the future because the computations here are also repeated in `_compute_horizontal_velocity_quantities_and_fluxes` which is called in `solve_nonhydro.py` after updating `vn`.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_terms.py
```diff
@@ -0,0 +1,1105 @@
...
+    c_intp: np.ndarray,
+    inv_dual_edge_length: np.ndarray,
+    inv_primal_edge_length: np.ndarray,
+    tangent_orientation: np.ndarray,
+) -> np.ndarray:
+    w_at_vertices = mo_icon_interpolation_scalar_cells2verts_scalar_ri_dsl_numpy(
+        connectivities, w, c_intp
+    )
+
+    return compute_horizontal_advection_term_for_vertical_velocity_numpy(
```
OngChia: Similar to the field operator, I think we can remove this numpy helper function and inline there in

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_terms.py
```diff
@@ -0,0 +1,1105 @@
...
+                (c2e2cO != -1)[:, :, np.newaxis],
+                w[c2e2cO] * geofac_n2s,
+                0,
+            ),
+            axis=1,
+        )
+    )
+
+
+class TestComputeInterpolatedHorizontalAdvectionOfW(stencil_tests.StencilTest):
```
OngChia, line 590: Do we need to have all these stencil tests, or just keep let's say compute_diagnostics_from_normal_wind?

OngChia (commented):
Looks good. Mainly whether to remove one field operator and renaming.

.cscs-ci/base.yml
```diff
@@ -224,6 +226,9 @@ variables:
 
 .test_tools_aarch64:
   extends: [.test_runner_serial, .test_template_aarch64, .retry_on_transient_failure]
+  variables:
+    # The bindings reference test compares against clang-formatted generated code.
```
nfarabullini: remove this comment

bindings/src/icon4py/bindings/dycore_wrapper.py
```diff
@@ -374,7 +374,8 @@ def solve_nh_run(  # noqa: PLR0917 [too-many-positional-arguments]
         dynamical_vertical_volumetric_flux_at_cells_on_half_levels=vol_flx_ic,
     )
 
-    # Make `max_vcfl` a 0-d array to avoid cupy synchronization, see `velocity_advection.py`.
+    # Make `max_vcfl` a 0-d array to avoid cupy synchronization, see `_update_max_vertical_cfl`
+    # in `solve_nonhydro.py`.
```
nfarabullini: remove in `solve_nonhydro.py`

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
```diff
@@ -456,6 +461,28 @@ def __init__(self, config: NonHydrostaticConfig):
         """
 
 
+def _velocity_advection_scale_factors(dtime: float) -> tuple[float, float]:
+    scaled_cfl_w_limit = 0.65 / dtime
+    scalfac_exdiff = 0.05 / (dtime * (0.85 - scaled_cfl_w_limit * dtime))
```
nfarabullini: how about:
```
scalfac_exdiff = 0.05/(d*(0.85d-0.65))
```
If you fold out scalfac_exdiff original equation, you get this

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: Inlined in 0bd1117c9: the expression now sits in `_compute_horizontal_advection_of_w`, and the operator, its program and its stencil test are gone. Since that test was the only focused test of this computation, `test_velocity_advection_terms.py` gains `TestComputeHorizontalAdvectionOfW` with the same numpy reference.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: Done in 0bd1117c9, but differently than asked, because the operator turned out to be a duplicate: `model/common` already had the same `neighbor_sum` over V2C as `_compute_cell_2_vertex_interpolation` (used by `metric_fields.py`). So the dycore copy is deleted and the **common** one carries your name, `interpolate_cell_field_to_vertex`. It returns wpfloat and velocity advection casts to vp at the call site. Its program had no user besides its own test, so it is gone too and the test drives the field operator.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_terms.py
havogt: Renamed to `compute_curl` in 0bd1117c9 and moved to `model/common/math/stencils/`, since a curl is a generic operation under CODING_GUIDELINES. It returns wpfloat with the cast at the call site, and its program is dropped as well (only its test used it). The test moved to `model/common/tests/common/math/stencil_tests/test_compute_curl.py`.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_terms.py
havogt: Inlined in 0bd1117c9, together with the field operator. `compute_horizontal_advection_of_w_numpy` now writes the expression out and uses the shared `interpolate_cell_field_to_vertex_numpy` from `icon4py.model.testing.reference_funcs`. Checked bit-identical against the previous helpers on random inputs, including missing neighbours.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
havogt: Agreed that they are generic. Leaving them here for now and moving them with the tmx port, as you suggested, to keep this PR's scope. Leaving this thread open as the reminder.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
havogt: Noted, no change here. Leaving the thread open so the duplication with `_compute_horizontal_velocity_quantities_and_fluxes` is not lost.

model/atmosphere/dycore/tests/dycore/stencil_tests/test_velocity_advection_terms.py
havogt: These are deliberate: each one covers a non-trivial leaf operator whose only focused test was a legacy stencil test this PR deletes. The merged-program tests do cover them, but they don't say which operator broke, and their references restate the whole composition. Trivial and composite operators got no test. Happy to drop any you think are not worth it.

bindings/src/icon4py/bindings/dycore_wrapper.py
havogt: Shortened in ef760dff7.

.cscs-ci/base.yml
havogt: I would rather keep this one. Every other test job in this file has clang-format disabled through `.test_runner_base`, and this line puts it back for the tools job only. Without a reason next to it, the override reads as an oversight and invites deletion: removing it breaks `bindings/tests/bindings/test_codegen_references.py`, which compares py2fgen's clang-formatted C header against a committed reference. That is exactly how it broke once in this PR. Happy to shorten it to `# the bindings reference test compares formatted code` if the length is the issue.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
havogt: The folding is exact for every realistic `dtime`, but not in general, so I kept the unfolded form for now.

Measured in double precision: for 24 values from 0.5 s to 720 s, `0.05/(dtime*(0.85 - (0.65/dtime)*dtime))` and `0.05/(dtime*(0.85 - 0.65))` agree bit for bit. Over 200000 random values in [0.001, 3600] s, 4 % differ, by one ulp at most (worst relative difference 7.4e-16), because `(0.65/dtime)*dtime` is not always exactly 0.65.

The current form also matches ICON, which computes `cfl_w_limit = 0.65/dtime` first and then `scalfac_exdiff = 0.05/(dtime*(0.85 - cfl_w_limit*dtime))`. Note `0.85 - 0.65` is `0.19999999999999996`, not 0.2, so the folded form is not simpler numerically either. Say the word and I'll fold it.

.cscs-ci/base.yml
havogt: Actually I'll remove the clang-format changes completely from here again.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
havogt: We should leave the form as in ICON.

.cscs-ci/base.yml
havogt: Moot here: the clang-format change is out of this PR as of fd788ed32 and lives in #1477, which sits on main so its CI-time effect can be measured on its own. The comment question moves there with it.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
havogt: Both are going into a separate cleanup PR rather than this one. On the tmx side the kinetic energy has no counterpart at all, and the half-level interpolation would land in `interpolate_edge_field_to_half_levels.py`, a module that exists only in the tmx port (#1475) — creating it in both PRs would just collide. Leaving this thread open as the marker.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/stencils/velocity_advection_predictor.py
havogt: Tried this one out: a shared `_compute_diagnostics_from_normal_wind` in `compute_diagnostics_from_normal_wind.py` computing the five quantities both callers derive from `vn`, with the predictor keeping the top extrapolation and the skip at its call site, and `_compute_horizontal_velocity_quantities_and_fluxes` keeping the `nflatlev` guard and the fluxes. It passes on gtfn_cpu (merged predictor 6 passed, horizontal velocity quantities 3 passed) and drops about 50 lines. It goes into the same separate cleanup PR, to keep this one's scope. Leaving the thread open until that lands.

model/atmosphere/dycore/src/icon4py/model/atmosphere/dycore/solve_nonhydro.py
havogt: Or actually I think we can change it. @OngChia Is the folded form or the current form more natural in papers or is that anyway random because these are arbitrary limits in some sense.

OngChia: As far as I know, this magic number does not appear anywhere in any papers. You definitely can fold it, I would suggest you simply do
```python
cfl_w_limit = 0.65 # I am fine with either exposing it in config or not. If yes, 0.85 is another magic number that may need to be made configurable too (maximum cfl number at which w is allowed)
scalfac_exdiff = 0.05 /  (0.85 - cfl_w_limit) # 0.05 is just a diffusion "strength"
```
The computaation in the cfl clipping and extra vertical diffusion for w may need to be modified accordingly.

havogt: Done in fe803ebdc, in your dimensionless form. `VerticalCflConstants` now holds `W_LIMIT = 0.65`, `W_MAX = 0.85` and `EXTRA_DIFFUSION_SCALING = 0.05 / (W_MAX - W_LIMIT)` next to the operators, as a wpfloat enum like `PhysicsConstants`, and the operators close over it. So `cfl_w_limit` and `scalfac_exdiff` are gone from every operator and program signature, and `SolveNonhydro` no longer computes them per step. The stencils compare the vertical CFL number with the limit and scale the diffusion coefficient by 1/dtime. Round-off in the extra diffusion changes in the last bits; the clipping decision does not. Datatests against ICON pass, and the serialized per-second values are still checked after dividing by dtime. I did not make them configurable — say if you want that here.

havogt:
Note the last change regarding the cfl limits, do you agree with the changes? https://github.com/C2SM/icon4py/pull/1460/commits/fe803ebdc10094d9691934e589dd50421f3dce91

model/common/src/icon4py/model/common/interpolation/stencils/interpolate_cell_field_to_vertex.py
```diff
@@ -0,0 +1,22 @@
...
+import gt4py.next as gtx
+from gt4py.next import neighbor_sum
+
+import icon4py.model.common.type_alias as types
+from icon4py.model.common import dimension as dims, field_type_aliases as fa
+from icon4py.model.common.dimension import V2C
+
+
+@gtx.field_operator
+def _interpolate_cell_field_to_vertex(
```
nfarabullini, line 17: should we place this in some utils file instead of being self-standing?

nfarabullini (approved):

model/common/src/icon4py/model/common/interpolation/stencils/interpolate_cell_field_to_vertex.py
havogt: I agree. I'd propose we refactor the `interpolation/stencils` directory separately. There is also a mix of `x2y` and `x_to_y` in that directory.

PR #1461, opened 2026-09-08
Add note about : as separator in CSCS CI pipeline variables to CI reminder

msimberg:
Also adds an empty line that was missing after a list.


starkphi (approved):
Looking good. Didn't know we could also list multiple pipelines in the same command, so I think it's a very helpful additional line.

PR #1463, opened 2026-09-10
ProgramFieldProvider: the declared domain is the field's vertical extent

havogt:
## Description

The field providers in `states/factory.py` allocated (or labelled) every output at full grid size and read `domain` only as a compute range. Now there is one rule, `_field_extent`, shared by all three providers: a declared vertical range is the field's extent, with absolute level indices; horizontal dimensions are always full local size, because the halo exchange fills entries outside the compute range and neighbour access indexes the field by absolute local index. So `domain` is the field's domain and the compute range at once.

Per provider:

- `ProgramFieldProvider`: allocates through the rule; `vertical_start`/`vertical_end` are that range. The two programs that only fill part of the column, `compute_rayleigh_w` and `compute_coeff_dwdz`, now write the zeros themselves with `concat_where` and run over the full column. `compute_rayleigh_w` takes `end_index_of_damping_layer` for that. The zero branches are a closed-domain field (`0.0 * rayleigh_w`) and a single-level scalar (`KDim == 0`) because embedded rejects a scalar branch on an unbounded region (`Cannot compute length of open 'UnitRange'`).
- `EmbeddedFieldOperatorProvider`: allocates through the rule and, as before, computes on the whole extent of its outputs. Its private size map is gone; missing output metadata falls back to the default dtype as in the program provider.
- `NumpyDataProvider`: a `domain` given with ranges is labelled through the rule, so `wgtfacq_c` and `wgtfacq_e` declare their three bottom levels with zone markers instead of `num_levels - 3` computed at construction. A bare dimension tuple still labels the returned array by its shape, since dimensions such as `LsqUnkDim` have no grid size.

### DIFF
--- a/model/common/src/icon4py/model/common/metrics/metric_fields.py
+++ b/model/common/src/icon4py/model/common/metrics/metric_fields.py
@@ -285,9 +285,13 @@ def _compute_coeff_dwdz(
         ddqz_z_full(dims.KDim - 1) / ddqz_z_full / (z_ifc(dims.KDim - 1.5) - z_ifc(dims.KDim + 0.5))
     )
 
+    # TODO(havogt): This is a workaround for 2 things:
+    # a) with a plain `0.0` embedded will not work because of the infinite range
+    # b) for `concat_where(dims.KDim == 0, 0.0, ...)` the domain inference is broken in GT4Py,
+    #    see https://github.com/gridTools/gt4py/issues/2205.
     return (
-        concat_where(dims.KDim == 0, 0.0, coeff1_dwdz),
-        concat_where(dims.KDim == 0, 0.0, coeff2_dwdz),
+        concat_where(dims.KDim >= 1, coeff1_dwdz, 0.0 * ddqz_z_full),
+        concat_where(dims.KDim >= 1, coeff2_dwdz, 0.0 * ddqz_z_full),
     )
 
 
--- a/model/common/src/icon4py/model/common/metrics/metric_fields.py
+++ b/model/common/src/icon4py/model/common/metrics/metric_fields.py
@@ -210,6 +210,7 @@ def _compute_rayleigh_w(  # noqa: PLR0917 [too-many-positional-arguments]
     rayleigh_coeff: wpfloat,
     vct_a_1: wpfloat,
     pi_const: wpfloat,
+    end_index_of_damping_layer: gtx.int32,
 ) -> fa.KHalfField[wpfloat]:
     rayleigh_w = broadcast(0.0, (dims.KHalfDim,))
     z_sin_diff = maximum(0.0, vct_a - damping_height)
@@ -224,7 +225,8 @@ def _compute_rayleigh_w(  # noqa: PLR0917 [too-many-positional-arguments]
         rayleigh_w = rayleigh_coeff * (
             1.0 - tanh(3.8 * z_tanh_diff / maximum(0.000001, vct_a_1 - damping_height))
         )
-    return rayleigh_w
+    # embedded rejects a scalar branch on an unbounded region, so the zeros are a field
+    return concat_where(dims.KHalfDim <= end_index_of_damping_layer, rayleigh_w, 0.0 * rayleigh_w)
 
 
 @gtx.program
@@ -236,6 +238,7 @@ def compute_rayleigh_w(  # noqa: PLR0917 [too-many-positional-arguments]
     rayleigh_coeff: wpfloat,
     vct_a_1: wpfloat,
     pi_const: wpfloat,
+    end_index_of_damping_layer: gtx.int32,
     vertical_start: gtx.int32,
     vertical_end: gtx.int32,
 ):
@@ -254,6 +257,7 @@ def compute_rayleigh_w(  # noqa: PLR0917 [too-many-positional-arguments]
         rayleigh_klemp: Klemp (2008) type Rayleigh damping
         rayleigh_coeff: Rayleigh damping coefficient in w-equation
         pi_const: pi constant
+        end_index_of_damping_layer: last level index with damping, rayleigh_w is zero below
         vertical_start: vertical start index
         vertical_end: vertical end index
     """
@@ -264,6 +268,7 @@ def compute_rayleigh_w(  # noqa: PLR0917 [too-many-positional-arguments]
         rayleigh_coeff,
         vct_a_1,
         pi_const,
+        end_index_of_damping_layer,
         out=rayleigh_w,
         domain={dims.KHalfDim: (vertical_start, vertical_end)},
     )
@@ -280,7 +285,10 @@ def _compute_coeff_dwdz(
         ddqz_z_full(dims.KDim - 1) / ddqz_z_full / (z_ifc(dims.KDim - 1.5) - z_ifc(dims.KDim + 0.5))
     )
 
-    return coeff1_dwdz, coeff2_dwdz
+    return (
+        concat_where(dims.KDim == 0, 0.0, coeff1_dwdz),
+        concat_where(dims.KDim == 0, 0.0, coeff2_dwdz),
+    )
 
 
 @gtx.program(grid_type=gtx.GridType.UNSTRUCTURED)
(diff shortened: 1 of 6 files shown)


havogt:
cscs-ci run default;LEVELS=unit:integration;MODEL_SUBPACKAGES=common:driver;MODEL_SUBSETS=datatest

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -400,9 +411,11 @@ def _compute(self, factory: FieldSource, grid_provider: GridProvider) -> None:
             f"{data_alloc.backend_name(compute_backend)}, target backend is: "
             f"{data_alloc.backend_name(factory.backend)}"
         )
-        xp = data_alloc.import_array_ns(factory.backend)
-        metadata = {k: factory.get(k, RetrievalType.METADATA) for k in self.fields}
-        self._fields = self._allocate_fields(compute_backend, grid_provider, xp, metadata)
+        try:
+            metadata = {k: factory.get(k, RetrievalType.METADATA) for k in self.fields}
+        except (ValueError, KeyError):
+            metadata = {}
```
havogt: what's this about?

havogt: ok, it's the pattern that is used in the other providers.

model/common/src/icon4py/model/common/states/factory.py
```diff
@@ -794,7 +787,5 @@ def _func_name(callable_: Callable[..., Any]) -> str:
         return callable_.__name__
 
 
-def dtype_or_default(
-    field_name: str, metadata: dict[str, model.FieldMetaData]
-) -> state_utils.ScalarType:
-    return metadata[field_name].get("dtype", ta.wpfloat)
+def output_dtype(field_src: FieldSource, field_name: str) -> state_utils.ScalarType:
+    return field_src.get(field_name, RetrievalType.METADATA).get("dtype", ta.wpfloat)
```
havogt, line 792: this replaces the weird try...except clause in all Providers.

model/common/tests/common/states/unit_tests/test_factory.py
```diff
@@ -208,6 +210,119 @@ def test_program_provider(height_coordinate_source: SimpleFieldSource) -> None:
...
 
+def _run_program_provider(source: SimpleFieldSource, domain: dict) -> gtx.Field:
+    provider = factory.ProgramFieldProvider(
+        func=vertical_ops.average_two_vertical_levels_downwards_on_cells,
+        domain=domain,
+        fields={"average": "output_f"},
+        deps={"input_field": "height_coordinate"},
+        do_exchange=False,
+    )
+    source.with_metadata({"output_f": {"standard_name": "output_f", "units": ""}})
```
havogt: Previously the factories would work with entries without metadata because of the try...except block, which could have hidden real errors. Now we have to provide fake metadata to test the providers.

model/common/tests/common/states/unit_tests/test_factory.py
```diff
@@ -208,6 +210,73 @@ def test_program_provider(height_coordinate_source: SimpleFieldSource) -> None:
...
+) -> None:
+    assert height_coordinate_source.vertical_grid is not None
+    num_levels = height_coordinate_source.vertical_grid.num_levels
+    provider = make_provider(
+        {
+            dims.CellDim: (cell_domain(h_grid.Zone.LOCAL), cell_domain(h_grid.Zone.END)),
+            dims.KDim: (v_grid.Domain(dims.KDim, v_grid.Zone.TOP, 1), k_domain(v_grid.Zone.BOTTOM)),
+        }
+    )
+    height_coordinate_source.with_metadata({"output_f": {"standard_name": "output_f", "units": ""}})
```
havogt, line 262: Previously the factories would work with entries without metadata because of the try...except block, which could have hidden real errors. Now we have to provide fake metadata to test the providers.

model/common/src/icon4py/model/common/states/factory.py
msimberg: Just FYI: https://github.com/C2SM/icon4py/pull/970#discussion_r3729456383 (TL;DR I like the cleanup). And FYI for @starkphi that this will change and possibly affect your single-precision PR?

msimberg, line 792: This will also overlap with #970, just FYI @starkphi.

model/common/tests/common/metrics/unit_tests/test_metric_fields.py
```diff
@@ -141,7 +141,7 @@ def test_compute_rayleigh_w(
 ) -> None:
     rayleigh_w_ref = metrics_savepoint.rayleigh_w()
     vct_a_1 = grid_savepoint.vct_a().asnumpy()[0]
-    rayleigh_w_full = data_alloc.zero_field(icon_grid, dims.KHalfDim, allocator=backend)
+    rayleigh_w_full = data_alloc.random_field(icon_grid, dims.KHalfDim, allocator=backend)
```
msimberg, line 144: This is to test that we're not accidentally getting values at default-zero, right? I'm surprised if these are the only tests that suffered from this?

msimberg (approved):
This looks like a nice cleanup! I'm not 100% familiar with all of the factory internals, but from what I do know all of your changes look like relatively safe and clean simplifications to my eye.

There's overlap with @starkphi's changes in #970 which will need resolving, but I'm assuming this goes in first.

Regarding

> horizontal dimensions are always full local size, because the halo exchange fills entries outside the compute range and neighbour access indexes the field by absolute local index

would it be fair to say that we could consider allocating a subset also in the horizontal dimension, but it's much more difficult because we don't know which programs might end up needing the halos?(We don't always fill in the halos, or we probably wouldn't always need to fill in the halos, but it's manually decided, not automatically. This is of course outside the scope of this PR.

havogt:
> would it be fair to say that we could consider allocating a subset also in the horizontal dimension, but it's much more difficult because we don't know which programs might end up needing the halos?(We don't always fill in the halos, or we probably wouldn't always need to fill in the halos, but it's manually decided, not automatically. This is of course outside the scope of this PR.

I wanted to say yes, but I am not 100% sure. For global it's probably easy: if we request "halo_exchange" then we should allocate with halo, if we don't request halo we should only allocate interior. Now with local area and boundary I am not sure which cases we have. Maybe there are actually arrays which only live on the boundary?

model/common/tests/common/metrics/unit_tests/test_metric_fields.py
havogt: Yes, probably there are more. Previously it would have actually failed with `random_field` because we would have computed only in the subset...

model/common/src/icon4py/model/common/states/factory.py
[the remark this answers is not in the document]
havogt: this is intentional...

model/common/src/icon4py/model/common/metrics/metric_fields.py
```diff
@@ -302,8 +310,8 @@ def compute_coeff_dwdz(  # noqa: PLR0917 [too-many-positional-arguments]
     Args:
         ddqz_z_full: functional determinant of the metrics (is positive), full levels
         z_ifc: geometric height of half levels
-        coeff1_dwdz: coefficient for second-order acurate dw/dz term
-        coeff2_dwdz: coefficient for second-order acurate dw/dz term
+        coeff1_dwdz: coefficient for second-order acurate dw/dz term, zero on the top level
+        coeff2_dwdz: coefficient for second-order acurate dw/dz term, zero on the top level
```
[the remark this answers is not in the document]
havogt: not wasting a ci run for this...

havogt:
cscs-ci run default;BACKENDS=dace_cpu:gtfn_gpu;LEVELS=unit:integration;MODEL_SUBPACKAGES=common:driver;MODEL_MPI_SUBPACKAGES=common:driver;MODEL_SUBSETS=datatest

havogt:
This is the failed pipeline. If restarted jobs pass, I'll force merge https://gitlab.com/cscs-ci/ci-testing/webhook-ci/mirrors/5125340235196978/2255149825504678/-/pipelines/2839892355.

havogt:
Looks like a real problem, but not sure if it's on icon4py or gt4py side.

PR #1465, opened 2026-09-10
Remove do prep adv from config

OngChia:
Remove `do_prep_adv` from the DriverConfig because it is a derived property. 
The dycore always needs to compute the backward trajectory, mass and volumetric fluxes whenever tracer advection is turned on because tracer advection needs these fields to transport the tracers. Therefore, `do_prep_adv` is `True` if dycore and tracer advection are both turned on.

For reference, in the original ICON driver [mo_nh_stepping,f90], `lprep_adv` is also not a namelist parameter.(https://gitlab.dkrz.de/icon/icon/-/blob/main/src/atm_dyn_iconam/mo_nh_stepping.f90#L3167), `lprep_adv` is True according to the if statement below
```
IF ( ltransport .OR. p_patch%n_childdom > 0 .AND. grf_intmethod_e == 6 ) THEN
      lprep_adv = .TRUE. ! do computations for preparing tracer advection in solve_nh
ELSE
      lprep_adv = .FALSE.
ENDIF
```
p_patch%n_childdom and grf_intmethod_e are unrelated in icon4py since we still do not support nesting and limited area for the time being.

### DIFF
--- a/model/driver/src/icon4py/model/driver/config.py
+++ b/model/driver/src/icon4py/model/driver/config.py
@@ -210,19 +210,6 @@ class DriverConfig:
             ),
         ),
     ] = False
-    do_prep_adv: typing.Annotated[
-        bool,
-        common_conf_opt.ConfigOption(
-            description="No description available yet.",
-            icon_equivalent=common_conf_opt.IconOption(
-                name="ltransport",
-                path=(
-                    "model_cfg",
-                    "run_nml",
-                ),
-            ),
-        ),
-    ] = False  # lprep_adv in fortran
     diffuse_before_time_loop: typing.Annotated[
         bool,
         common_conf_opt.ConfigOption(
--- a/model/driver/src/icon4py/model/driver/driver.py
+++ b/model/driver/src/icon4py/model/driver/driver.py
@@ -123,7 +123,7 @@ class Icon4pyDriver:
         return f"{self.__class__.__name__}:{func.__name__}"
 
     @functools.cached_property
-    def do_prep_adv_in_dycore(self) -> bool:
+    def _do_prep_adv_in_dycore(self) -> bool:
         return self.config.nonhydrostatic is not None and self.config.tracer_advection is not None
 
     @functools.cached_property
@@ -450,7 +450,7 @@ class Icon4pyDriver:
                     dtime=self.model_time_variables.substep_timestep,
                     ndyn_substeps_var=self.model_time_variables.ndyn_substeps_var,
                     at_initial_timestep=self.model_time_variables.is_first_step_in_simulation,
-                    lprep_adv=self.do_prep_adv_in_dycore,
+                    lprep_adv=self._do_prep_adv_in_dycore,
                     at_first_substep=self._is_first_substep(dyn_substep),
                     at_last_substep=self._is_last_substep(dyn_substep),
                 )
--- a/model/driver/src/icon4py/model/driver/driver.py
+++ b/model/driver/src/icon4py/model/driver/driver.py
@@ -123,6 +123,10 @@ class Icon4pyDriver:
         return f"{self.__class__.__name__}:{func.__name__}"
 
     @functools.cached_property
+    def do_prep_adv_in_dycore(self) -> bool:
+        return True if self.config.nonhydrostatic is not None and self.config.tracer_advection is not None else False
+
+    @functools.cached_property
     def _diagnostics_computer(self) -> driver_io.DiagnosticsComputer:
         """Reuses its scratch/output buffers across output steps (allocated once)."""
         return driver_io.DiagnosticsComputer(grid=self.grid, backend=self.backend)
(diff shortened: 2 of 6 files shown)


model/driver/src/icon4py/model/driver/driver.py
```diff
@@ -122,6 +122,10 @@ def _is_first_substep(step_nr: int) -> bool:
     def _full_name(self, func: Callable) -> str:
         return f"{self.__class__.__name__}:{func.__name__}"
 
+    @functools.cached_property
+    def _do_prep_adv_in_dycore(self) -> bool:
+        return self.config.nonhydrostatic is not None and self.config.tracer_advection is not None
```
yiluchen1066: More for question from my side, if I understand it correctly: 
```self.config.tracer_advection``` is None only in the MCH of APE_R02B04 experiment case even though ```ltransport=.TRUE.``` and ```GAUSS3D1``` get it even though ```ltransport=.FALSE.```

And also here the  ```self.config.nonhydrostatic is not None``` part can never be False here. ```_do_prep_adv_in_dycore(self)``` is only called inside if ```self.config.nonhydrostatic is not None``` (driver.py:306 → :310)

File: model/driver/src/icon4py/model/driver/driver.py [hunk 2]
```diff
@@ -122,6 +122,10 @@ def _is_first_substep(step_nr: int) -> bool:
     def _full_name(self, func: Callable) -> str:
         return f"{self.__class__.__name__}:{func.__name__}"
 
+    @functools.cached_property
+    def _do_prep_adv_in_dycore(self) -> bool:
```
yiluchen1066, line 126: shall we put this in ```ExperimentConfig.__post_init__``` instead of in the driver since there is already very similar derivation there. Then it could also be visible to the other users of ExperimentConfig

model/driver/src/icon4py/model/driver/config.py
```diff
@@ -210,19 +210,6 @@ class DriverConfig:
             ),
         ),
     ] = False
-    do_prep_adv: typing.Annotated[
-        bool,
-        common_conf_opt.ConfigOption(
-            description="No description available yet.",
-            icon_equivalent=common_conf_opt.IconOption(
-                name="ltransport",
```
yiluchen1066, line 218: No super important: does removing the field breaks existing YAML configs? if it does, then any YAML run config someone already has that still contains ```do_prep_adv: false``` will now raise error. Maybe worth writing a line in the pr description so people know to delete the key

model/driver/src/icon4py/model/driver/config.py
```diff
@@ -210,19 +210,6 @@ class DriverConfig:
             ),
         ),
     ] = False
-    do_prep_adv: typing.Annotated[
```
yiluchen1066, line 213: I feel like we should not delete the whole ```do_pre_adv``` here. this is the only ```IconOption``` in the whole driver config pointing at ```ltransport``` Maybe we can keep the annotation, rename the field to something like ```transport_enabled``` and keep ```IconOption(name="ltransport",...)

model/driver/src/icon4py/model/driver/driver.py
yiluchen1066, line 127: A small question about the first half of the condition: the property is only read at :453, inside ```_do_dyn_substepping```, and that is only called at :314, inside if ```self.config.nonhydrostatic is not None:``` at :310. So ```self.config.nonhydrostatic`` is not None is always True when this runs. Is it there on purpose, as protection in case the property is used somewhere else later? If yes, maybe it is worth a short comment, otherwise it is easy to read it as a condition that can actually be False.

yiluchen1066 (commented):
LGTW, I left some points mostly are from my questions :) Thanks

model/driver/src/icon4py/model/driver/config.py
OngChia: Actually in the original code, `prep_adv` is a local variable in `mo_nh_stepping.f90` https://gitlab.dkrz.de/icon/icon/-/blob/main/src%2Fatm_dyn_iconam%2Fmo_nh_stepping.f90?blame=1#L3156,  It is `True` if tracer advection is on or nesting with a type of interpolation. We do not support nesting, so it is equivalent to whether tracer advection is turned on or not (equivalent to `ltransport` in ICON). But I think it does not fit into our current code structure if we keep `ltransport` (tracer advection), or `ldynamics` (solve_nonhydro), or other l*** for granules or components in the config here. Currently, the driver just checks if config for a particular granule is missing or not. So, I will prefer to delete `do_prep_adv`

model/driver/src/icon4py/model/driver/config.py
OngChia: Yes. The only YAML configs available in icon4py are the configs I added in David's [PR](https://github.com/C2SM/icon4py/pull/580) which I merged. I will remove them. And then probably no one will even notice this `do_prep_adv` unless in the data test.

model/driver/src/icon4py/model/driver/driver.py
OngChia: I prefer to keep config clean and short, and add this here. But this can be discussed.

I have changed the name `lprep_adv` and `do_prep_adv_in_dycore` to `prepare_fluxes_for_advection`.

model/driver/src/icon4py/model/driver/driver.py
OngChia: You are right! I removed it. Thanks!

model/driver/src/icon4py/model/driver/config.py
OngChia: I will also remove it in test config.

yiluchen1066 (approved):

PR #1466, opened 2026-09-10
Update the driver CLI to read config from yaml & remove legacy jsons & move f90->yaml conversion to scripts

OngChia:


jcanton (approved):
nice

scripts/python/fortran_config_converter.py
```diff
@@ -0,0 +1,659 @@
...
+def make_driver_config(
+    *, atm_dict: dict[str, Any], master_dict: dict[str, Any], **overrides: Any
+) -> driver_config.DriverConfig:
+    icon_config = {"master_cfg": master_dict, "model_cfg": atm_dict}
+    kwargs = _kwargs_from_table(driver_config.DriverConfig, _DRIVER_OPTIONS, icon_config)
+    # `modeltimestep` (ISO 8601 duration) takes priority over the legacy `dtime`
+    # (seconds); ICON writes it as a fixed-width, blank-padded string.
+    run_nml = atm_dict["run_nml"]
+    kwargs["dtime"] = driver_config.relativetime_from_iconformat(
+        run_nml["dtime"], run_nml["modeltimestep"].strip()
```
DropD: Odd mix of using the table full of IconOptions above and manually reading from `atm_dict`

scripts/python/fortran_config_converter.py
```diff
@@ -0,0 +1,659 @@
...
+        thslp_zdiffu=nonhydrostatic_nml["thslp_zdiffu"],
+        thhgtd_zdiffu=nonhydrostatic_nml["thhgtd_zdiffu"],
+        rayleigh_type=constants.RayleighType(nonhydrostatic_nml["rayleigh_type"]),
+        rayleigh_coeff=list_to_value(nonhydrostatic_nml["rayleigh_coeff"]),
+        divdamp_trans_start=nonhydrostatic_nml["divdamp_trans_start"],
+        divdamp_trans_end=nonhydrostatic_nml["divdamp_trans_end"],
+        divdamp_type=nonhydrostatic_nml["divdamp_type"],
+        igradp_method=nonhydrostatic_nml["igradp_method"],
+        **overrides,
+    )
```
DropD: Odd mix of some configs using `IconOption`s and others not.

scripts/python/fortran_config_converter.py
```diff
@@ -0,0 +1,659 @@
...
+    atm_dict: dict[str, Any],
+) -> prescribed_tendencies.PrescribedTendenciesConfig:
+    if atm_dict["run_nml"]["ltestcase"]:
+        return prescribed_tendencies.PrescribedTendenciesConfig(data_path=None)
+    return prescribed_tendencies.PrescribedTendenciesConfig(data_path=SER_DATA_PATH)
+
+
+# ---------------------------------------------------------------------------
+# Orchestration
+# ---------------------------------------------------------------------------
```
DropD: If a module contains such disparate sections that they need to be annotated with comments, doesn't that mean it should be a package with multiple modules?

scripts/python/fortran_config_converter.py
```diff
@@ -0,0 +1,659 @@
...
+    IconOption("lsq_high_ord", ("interpol_nml", "lsq_high_ord")),
+]
+
+
+def make_interpolation_config(
+    atm_dict: dict[str, Any], **overrides: Any
+) -> interpolation_factory.InterpolationConfig:
+    return _build_from_table(
+        interpolation_factory.InterpolationConfig, _INTERPOLATION_OPTIONS, atm_dict, overrides
+    )
```
DropD: Why does this function exist? If it needs to, why is it not a `functools.partial`?

scripts/python/fortran_config_converter.py
```diff
@@ -0,0 +1,659 @@
...
+        exponent_for_density_factor_in_ice_sedimentation=nwp_tuning_nml["tune_icesedi_exp"],
+        power_law_coeff_for_snow_fall_speed=nwp_tuning_nml["tune_v0snow"],
+        rain_mu=nwp_phy_nml["mu_rain"],
+        rain_n0=nwp_phy_nml["rain_n0_factor"],
+        snow2graupel_riming_coeff=nwp_tuning_nml["tune_zcsg"],
+        **overrides,
+    )
+
+
+# Fortran namelist keys of the analytical test cases → ICON4Py dataclass fields.
```
DropD: Is the plan to simplify this from three different ways to translate to one? When the three different ways were all over the place it was obvious to me why converging had to be done incrementally. But this way it looks like it's begging to be cleaned up soon.

PR #1470, opened 2026-09-14
Port of weno scheme and idealized experiments

muellch:


jcanton:
@muellch should then https://github.com/C2SM/icon4py/pull/1396 be closed? please do it if so.

muellch:
> @muellch should then #1396 be closed? please do it if so.

Yes, and I am sorry, I did not see your PR, otherwise I would have continued there.

PR #1473, opened 2026-09-14
Fix nabla2_of_vn for edges with missing E2C2V neighbors

edopao:
## Summary

`calculate_nabla2_and_smag_coefficients_for_vn` computes `nabla2_of_vn` as a weighted sum over the E2C2V neighbors. The previous `neighbor_sum` form kept the `vn` contribution as a separate, unconditional term:

```python
nabla2_of_vn = neighbor_sum(v_n * weight, axis=E2C2VDim) - 2.0 * vn * (inv_primal_edge_length_2 + inv_vert_vert_length_2)
```

`neighbor_sum` drops **skipped** E2C2V neighbors (connectivity value `-1`) from the `v_n` sum, but the `- 2.0 * vn * (...)` term was subtracted regardless. On real grids these disagree: the **MCH R04B09** grid has **428 edges whose 4th E2C2V neighbor (index 3, a `vert_vert_length` neighbor) is a skip value**. For those edges the reduction omits that neighbor's `v_n` while the full `vn` term is still subtracted, yielding a wrong `nabla2_of_vn` on the boundary edges (and, downstream, a `vn` validation mismatch in the diffusion run).

## Fix

Fold the `vn` subtraction into the reduction. Each neighbor's `weight` already carries its `inv_length**2` factor, so the whole expression is:

```python
nabla2_of_vn = neighbor_sum((v_n - vn) * weight, axis=E2C2VDim)
```

A skipped neighbor now drops **both** its `v_n` and its `vn` contribution together, which is consistent. For edges whose E2C2V neighbors are all present the result is algebraically identical to before.

## Verification

- `test_run_diffusion_single_step` (MCH_CH_R04B09 and EXCLAIM_APE) **passes on `gtfn_cpu`** — the real grid exercises the 428 skip-neighbor edges.
- The stencil test `test_calculate_nabla2_and_smag_coefficients_for_vn` **passes on `dace_cpu` and embedded**.

## Note

While reproducing on `dace_cpu`, the full diffusion granule surfaced a **separate, pre-existing** `UnsupportedPythonFeatureError: 'ast.Assign'` at the `astype(...)` tuple-unpacking (line 42) on the dace lowering path — unrelated to this numerical fix and not addressed here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
+++ b/model/atmosphere/diffusion/src/icon4py/model/atmosphere/diffusion/diffusion.py
@@ -542,6 +542,14 @@ class Diffusion:
             offset_provider=self._grid.connectivities,
         )
 
+        # See calculate_nabla2_and_smag_coefficients_for_vn: E2C2V neighbors 0,1 carry
+        # the primal_edge_length**2 weight, neighbors 2,3 the vert_vert_length**2 one.
+        primal_edge_mask = data_alloc.field_from_values(
+            dims.E2C2VDim, [1.0, 1.0, 0.0, 0.0], allocator=self._allocator
+        )
+        vert_vert_mask = data_alloc.field_from_values(
+            dims.E2C2VDim, [0.0, 0.0, 1.0, 1.0], allocator=self._allocator
+        )
         self.calculate_nabla2_and_smag_coefficients_for_vn = setup_program(
             backend=backend,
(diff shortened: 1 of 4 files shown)


edopao:
cscs-ci run default

edopao:
This was not an issue. The compute domain ensures that the skip values are not accessed.

PR #1479, opened 2026-09-15
refactor[diffusion]: replace concat_where pass-throughs with per-output domains in w diffusion

edopao:
A6: dwdx and dwdy get their own per-output domain
{CellDim: (horizontal_start, horizontal_end), KHalfDim: (vertical_start+1, vertical_end)},
removing the concat_where(0 < dims.KHalfDim, ...) that skipped the surface
half level. The else-branch was a pure pass-through.

A7: w gets its own per-output domain
{CellDim: (interior_idx, halo_idx), KHalfDim: (vertical_start, vertical_end)},
removing the concat_where((interior_idx <= dims.CellDim) & (dims.CellDim < halo_idx), ..., w_old)
that wrote w_old (a value-identical copy of w) to boundary and halo rows.
The second concat_where (upper damping layer) stays because it is a
conditional modification on top of w, not the final write.

The stencil test reference and input_data are updated to match: the output
w is initialized to w_old (instead of zeros) since the boundary is no
longer written back.

---

<sub>Stack created with <a href="https://github.com/github/gh-stack">GitHub Stacks CLI</a> • <a href="https://gh.io/stacks-feedback">Give Feedback 💬</a></sub>


edopao:
cscs-ci run dace

edopao:
cscs-ci run benchmark-bencher

PR #1484, opened 2026-09-16
Dycore optimizations related to the global vs regional grid profiling

dganellari:
https://github.com/dganellari/icon4py/blob/dycore-optimizations/amd_scripts/docs/DYCORE_GRANULE_ANALYSIS.md


model/atmosphere/dycore/tests/dycore/integration_tests/test_benchmark_solve_nonhydro.py
```diff
@@ -44,6 +49,72 @@
 from icon4py.model.testing.fixtures.stencil_tests import grid_manager
 
 
+@pytest.fixture(scope="session", autouse=True)
```
iomaganaris: The changes in this file are probably useful for Claude to do the benchmarking but should be removed before merging

amd_scripts/docs/DYCORE_GRANULE_ANALYSIS.md
```diff
@@ -0,0 +1,316 @@
+# Dycore performance: regional and global grids
```
iomaganaris: All the files in the `amd_scripts` directory should be removed before merging but a summary of the changes and findings can go in https://hackmd.io/KIpq02uoQKmdgPTwVFeoug or a separate HackMD file in the GridTools group

model/common/src/icon4py/model/common/model_options.py
```diff
@@ -35,6 +37,68 @@ def _dace_remove_access_node_copies(sdfg: dace.SDFG) -> None:
...
+        return True
+    edge, level = "i_Edge_gtx_horizontal", "i_K_gtx_vertical"
+    first = dict(zip(first_map.params, first_map.range, strict=True))
+    second = dict(zip(second_map.params, second_map.range, strict=True))
+    transformation.allow_shared_data = (
+        set(first) == set(second) == {edge, level}
+        and first[edge] == second[edge]
+        and first[level] != second[level]
+    )
+    return True
```
iomaganaris: These checks seem very `gt4py`/`SDFG` related so that is not the best place to do them. I'm wondering though if there is an issue enabling for the `compute_rho_theta_pgrad_and_update_vn` program `allow_shared_data` by passing it as an option to https://github.com/GridTools/gt4py/blob/f463f013a4a305340300ae406d91daafce3a7e4e/src/gt4py/next/program_processors/runners/dace/transformations/auto_optimize.py#L146 or it should actually be an option at all and just the default behavior in this pass. I will run the `blueline` with this as default to check if there is any regression. If not I think we should enabled it by default everywhere without any check

model/common/tests/common/test_model_options.py
```diff
@@ -131,3 +135,137 @@ def test_setup_program_specify_inputs(
         )
     # TODO(havogt): test should be improved to work without string comparison
     assert repr(testee) == repr(expected_program)
+
+
+@pytest.mark.parametrize("setting", [None, "0", "1"])
```
iomaganaris: By removing the configuration checks and options from the `model_options.py` I think the tests in this file can also be removed I believe

model/testing/src/icon4py/model/testing/paired_benchmark.py
```diff
@@ -0,0 +1,475 @@
+# ICON4Py - ICON inspired code in Python and GT4Py
```
iomaganaris: The new tests in this folder look like Claude testing/timing framework so they can be removed before merging

iomaganaris (commented):
I think we can clean this up a bit to remove the Claude infrastructure and tests.
Also it should be merged after the `gt4py` related PR once we figure out if the passes are beneficial in general for other stencils or have any regression. See me comments below

